PR #309, opened 2021-02-17
Initial heap serialization design proposal

dmlloyd:


dgrove-oss:
For the record, we ended up implementing a scheme where we write a fully-formed heap as a (large) collection of data definitions into a separate Section.  Pointers into and within the initial heap are implemented by naming the targeted program object.  The linker then can fix up these references as part of initial program loading.

This is simple and avoids the runtime overheads of deserialization (good), but does make it harder to reclaim any of the initial heap that becomes garbage during subsequent program execution.  We'll assume this is ok until proven otherwise...

Issue #473, opened 2021-04-27
Enhance LLVM backend to invoke statepoint placement passes


dmlloyd:
The LLVM backend must be enhanced such that statepoints are placed in the bitcode; this is required to enable the backend to then parse the stack maps that are produced as a result.  There is some related prerequisite work that is a part of this task.

@aviansie-ben is working on this presently; the purpose of this task is to be a dependency for other tasks.

PR #483, opened 2021-04-29
Generate stack maps using Statepoint and RS4GC

aviansie-ben:
This PR implements sufficient support for Statepoint to begin using LLVM's built-in RS4GC pass to generate stack maps. These stack maps should be mostly correct and complete, with the exception of potentially missing some collected pointers in certain critical sections (e.g. immediately after allocating memory for an object but before a pointer to it is converted from raw to collected. This has involved the introduction of a couple pseudointrinsics to hide `addrspacecast` instructions that may be treated incorrectly by LLVM prior to running RS4GC. These pseudointrinsics are defined in the same module as the code itself and are inlined immediately after RS4GC is run. Theoretically, there may still be problems if certain LLVM optimizations are run prior to running RS4GC (namely, inlining and function attribute inference). These issues could be corrected in the future by placing the pseudointrinsics in question in a separate LLVM module and linking them together using the `llvm-link` after running potentially problematic optimizations.

Note that this PR slightly changes the structure of the temporary files in the output directory. The `.ll` files in the output directory represent the LLVM IR as it was immediately after QBICC emitted them. A separate `_opt.bc` file will then be used to store the optimized bitcode before running `llc`. These bitcode files can be disassembled using the `llvm-dis` tool if they need to be examined for debugging purposes.

Closes: #473

### DIFF
--- a/machine/llvm/src/main/java/org/qbicc/machine/llvm/FunctionAttributes.java
+++ b/machine/llvm/src/main/java/org/qbicc/machine/llvm/FunctionAttributes.java
@@ -9,7 +9,11 @@ public final class FunctionAttributes {
     private FunctionAttributes() {}
 
     public static final LLValue alwaysinline = LLVM.flagAttribute("alwaysinline");
-    public static final LLValue gcLeafFunction = LLVM.flagAttribute("gc-leaf-function");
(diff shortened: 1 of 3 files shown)


dmlloyd:
Adding a note that this LLVM bug is tracked here: https://bugs.llvm.org/show_bug.cgi?id=50553

PR #509, opened 2021-05-10
Changes to enable exception handling

ashu-mehra:
- Updated landingpad implementation to use cleanup instead of catch-all
to make it work with Statepoints as mentioned in the following link:
https://groups.google.com/g/llvm-dev/c/9N6OLp130C0/m/T1jDaL0hBgAJ

- Fix issues in personality function

- Remove UnwindSetupHook and avoid calls to getExactFunction() when
creating declaration for the personality function

- Add unit tests for try-catch control flow


runtime/unwind/src/main/java/org/qbicc/runtime/unwind/Unwind.java
```diff
@@ -90,6 +90,8 @@ private Unwind() {
          * The function which frees this object.
          */
         public function_ptr<_Unwind_Exception_Cleanup_Fn> exception_cleanup;
+        public uint64_t private_1;
```
dmlloyd, line 93: It looks like the actual type of these fields should be `unsigned_long`.  It could be incorrect on 32-bit systems otherwise (`long` is often 32 bits on such systems).

dmlloyd (commented):
Only one comment, otherwise looks OK to me.

dmlloyd (approved):

ashu-mehra:
@dmlloyd can this be merged now?

PR #511, opened 2021-05-11
Nullability overhaul

dmlloyd:
This is a draft PR to overhaul nullability checks.  At present, all method calls are assumed to never return `null`, which is obviously incorrect.  This PR fixes that problem.

* Make nullability a property of `Value`
* Introduce a `NotNull` node that asserts the non-nullability of a `Value`
* Introduce a `BooleanValue` interface which can be used to transform outbound values based on the truth of the boolean expression (for example, if `foo != null` then uses of `foo` will see `NotNull(foo)`)
* Make `CheckCast` and `InstanceOf` implement `OrderedNode` to prevent scheduling issues
* Many misc. fixes around nullability assumptions

### DIFF
--- a/compiler/src/main/java/org/qbicc/graph/InstanceOf.java
+++ b/compiler/src/main/java/org/qbicc/graph/InstanceOf.java
@@ -9,15 +9,17 @@ import org.qbicc.type.definition.element.ExecutableElement;
 /**
  * A node that represents a check of the upper bound of the value against the given type.
  */
-public final class InstanceOf extends AbstractValue implements InstanceOperation {
+public final class InstanceOf extends AbstractValue implements InstanceOperation, OrderedNode {
+    private final Node dependency;
     private final Value input;
     private final ObjectType checkType;
     private final int checkDimensions;
     private final BooleanType booleanType;
 
-    InstanceOf(final Node callSite, final ExecutableElement element, final int line, final int bci, final Value input,
+    InstanceOf(final Node callSite, final ExecutableElement element, final int line, final int bci, Node dependency, final Value input,
                final ObjectType checkType, final int checkDimensions, final BooleanType booleanType) {
         super(callSite, element, line, bci);
+        this.dependency = dependency;
         this.input = input;
         this.checkType = checkType;
         this.checkDimensions = checkDimensions;
@@ -33,7 +35,7 @@ public final class InstanceOf extends AbstractValue implements InstanceOperation
      }
 
     int calcHashCode() {
-        return Objects.hash(input, checkType, checkDimensions);
+        return Objects.hash(dependency, input, checkType, checkDimensions);
     }
 
     public boolean equals(final Object other) {
@@ -41,7 +43,7 @@ public final class InstanceOf extends AbstractValue implements InstanceOperation
     }
 
     public boolean equals(final InstanceOf other) {
-        return this == other || other != null && input.equals(other.input) && checkType.equals(other.checkType) && checkDimensions == other.checkDimensions;
+        return this == other || other != null && dependency.equals(other.dependency) && input.equals(other.input) && checkType.equals(other.checkType) && checkDimensions == other.checkDimensions;
     }
 
     public int getValueDependencyCount() {
@@ -62,5 +64,10 @@ public final class InstanceOf extends AbstractValue implements InstanceOperation
 
     public <T, R> R accept(final ValueVisitor<T, R> visitor, final T param) {
         return visitor.visit(param, this);
+    }
+
+    @Override
+    public Node getDependency() {
+        return dependency;
     }
 }
--- a/compiler/src/main/java/org/qbicc/graph/InstanceOf.java
+++ b/compiler/src/main/java/org/qbicc/graph/InstanceOf.java
@@ -4,12 +4,13 @@ import java.util.Objects;
 
 import org.qbicc.type.BooleanType;
 import org.qbicc.type.ObjectType;
+import org.qbicc.type.ReferenceType;
 import org.qbicc.type.definition.element.ExecutableElement;
 
 /**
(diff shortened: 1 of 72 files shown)


dmlloyd:
This should fix #543.

compiler/src/main/java/org/qbicc/graph/InstanceOf.java
```diff
@@ -4,20 +4,23 @@
...
 import org.qbicc.type.BooleanType;
 import org.qbicc.type.ObjectType;
+import org.qbicc.type.ReferenceType;
 import org.qbicc.type.definition.element.ExecutableElement;
 
 /**
  * A node that represents a check of the upper bound of the value against the given type.
  */
-public final class InstanceOf extends AbstractValue implements InstanceOperation {
+public final class InstanceOf extends AbstractValue implements InstanceOperation, OrderedNode, BooleanValue {
```
dgrove-oss, line 13: Ideally it would be nice if we could let InstanceOf nodes float more freely.  Is the OrderedNode addition needed because we eventually expand into a call in some cases or is it because of the BooleanValue and its interaction with phi nodes?

Should we open an issue to eventually get around to restoring these to not being OrderedNode?

compiler/src/main/java/org/qbicc/graph/SimpleBasicBlockBuilder.java
```diff
@@ -527,7 +532,7 @@ public Value classOf(Value typeId) {
         ValueType idType = typeIdType.getUpperBound();
         if (idType instanceof ReferenceType) {
             CompilationContext ctxt = classContext.getCompilationContext();
-            ctxt.warning(getLocation(), "TODO: class type ID argument is given as a reference type which is not allowed");
+//            ctxt.warning(getLocation(), "TODO: class type ID argument is given as a reference type which is not allowed");
```
dgrove-oss, line 535: I know this produces a ton of output, but did you intend to comment out the warning?  If so, maybe leave a TODO or open an issue so we don't lose track of it?

dgrove-oss (approved):

compiler/src/main/java/org/qbicc/graph/InstanceOf.java
dmlloyd: Right, converting it to control flow later causes some problems with scheduling.  We could fix the method call part using `callNoSideEffects` but the `if` stuff isn't as easy to fix (unless we use `select` instead, which is a possibility).

compiler/src/main/java/org/qbicc/graph/SimpleBasicBlockBuilder.java
dmlloyd: Yeah it's just way too noisy.  I'll open a follow-up issue.

dmlloyd: Done (#570).

DanHeidinga (commented):
Minor comments apart from the defender methods where I've been working examples and I don't think it's right yet.  Will post details on this in the #551

compiler/src/main/java/org/qbicc/type/definition/classfile/ClassFileImpl.java
```diff
@@ -1417,6 +1417,6 @@ ValueType getTypeOfVerificationInfo(int viTag, ExecutableElement element, ByteBu
     }
 
     ValueType nullable(ValueType v) {
-        return v instanceof ReferenceType ? ((ReferenceType) v).asNullable() : v;
+        return v instanceof ReferenceType ? ((ReferenceType) v) : v;
```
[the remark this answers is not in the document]
dmlloyd: Good catch. Auto-refactors sometimes leave artifacts like this.

compiler/src/main/java/org/qbicc/graph/ParameterValue.java
```diff
@@ -14,12 +14,14 @@
...
     private final String label;
     private final int index;
+    private final boolean nullable;
 
     ParameterValue(final Node callSite, final ExecutableElement element, final ValueType type, String label, final int index) {
         super(callSite, element, 0, -1);
         this.type = type;
         this.label = label;
         this.index = index;
+        nullable = label.equals("p");
```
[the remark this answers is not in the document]
dmlloyd: `p` is for parameter here.  All method parameters are potentially nullable (at least, for now).  But other implicit parameter types (like the current thread or the receiver) are not.

Eventually we'll have a more sophisticated analysis which can determine the nullability of parameters on an individual basis.

main/src/main/java/org/qbicc/main/Main.java
```diff
@@ -332,14 +333,17 @@ DiagnosticContext call0(BaseDiagnosticContext initialContext) {
                                 }
                                 builder.addBuilderFactory(Phase.ANALYZE, BuilderStage.INTEGRITY, ReachabilityBlockBuilder::new);
                                 builder.addElementVisitor(Phase.ANALYZE, new DotGenerator(Phase.ANALYZE, graphGenConfig));
-                                builder.addPostHook(Phase.ANALYZE, new ClassInitializerRegister());
+                                //builder.addPostHook(Phase.ANALYZE, new ClassInitializerRegister());
```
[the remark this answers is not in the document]
dmlloyd: Good question, I"ll give it a try.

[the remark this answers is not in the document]
dmlloyd: Yeah, it works now. I'll just delete the commit.

dmlloyd:
I've pushed up to rebase on #551. No other changes.

DanHeidinga:
I think both of these are ready to merge.  Does #551 need to be merged first?

PR #514, opened 2021-05-12
Annotate sys/types.h mode_t as integer

theresa-m:
I'm running into an issue where compiling [UnixConstants.java](https://github.com/qbicc/qbicc-class-library/blob/main/java.base/src/main/java/sun/nio/fs/UnixConstants.java) results in this error for lines 57-72:
```
UnixConstants.java:57: error: Cannot extend a signed integer of type s32 into a wider unsigned integer of type u32: either sign-extend to s32 first or cast to u32 first
```
[sys/types.h](https://pubs.opengroup.org/onlinepubs/009696899/basedefs/sys/types.h.html) indicates that `mode_t` is an integer type.


theresa-m:
Closing since I'm investigating a non workaround solution.

PR #521, opened 2021-05-20
Switch to public CI with a simple configuration

dmlloyd:


dmlloyd:
*Finally*.  I'm going to push this one in so we have CI again.

PR #522, opened 2021-05-20
Use schedule in copier

### LINKED ISSUE #517: Incorrect LLVM IR code is generated
Test case:
```
package hello.world;
import static org.qbicc.runtime.CNative.*;
public class Main {
    @extern
    public static native int putchar(int arg);

    static class Dummy {
        void dummy() {
                putchar('c');
        }
    }

    static class SubDummy extends Dummy {
        void sub1() {
                putchar('a');
        }
        void sub2() {
                putchar('b');
        }
    }

    public static Dummy getDummy(int i) {
            if (i == 0) {
                    return new SubDummy();
            } else {
                    return new Dummy();
            }
    }

    public static boolean check() {
        return true;
    }
        Dummy d = getDummy(args.length);
        if (check()) {
                if (d instanceof SubDummy) {
                        ((SubDummy)d).sub1();
                } else {
                        d.dummy();
                }
        } else {
                if (d instanceof SubDummy) {
                        ((SubDummy)d).sub2();
                } else {
                        d.dummy();
                }
        }
   }
}
```

Compiling the above results in following error:

```
./output/hello/world/Main.ll: error: `opt` invocation has failed: org.qbicc.machine.tool.CompilationFailureException: Compiler terminated with exit code 134
```

Actual error message generated by `opt` is:

```
$ opt -mtriple=x86_64-linux-gnu -rewrite-statepoints-for-gc -always-inline -o ./output/hello/world/Main.opt ./output/hello/world/Main.ll
opt: /lib64/libtinfo.so.5: no version information available (required by opt)
Instruction does not dominate all uses!
  %LD = icmp eq i1 %LC, false, !dbg !28
  br i1 %LD, label %Bf, label %B10, !dbg !32
LLVM ERROR: Broken module found, compilation aborted!
```

Problematic blocks are the `if-else` blocks with same `instanceof` checks in the `main` method. Although we initially create two different `InstanceOf` nodes, the equality as computed by `InstanceOf#equals(final InstanceOf other)` on the two nodes holds true and therefore in subsequent phases they get mapped to single `InstanceOf` node. This results in using the value computed in one branch to be used in another branch.

ashu-mehra:
Using Schedule when copying nodes ensures that the nodes are processed before their dependencies.

Fixes #517

### DIFF
--- a/compiler/src/main/java/org/qbicc/graph/CheckCast.java
+++ b/compiler/src/main/java/org/qbicc/graph/CheckCast.java
@@ -1,5 +1,6 @@
 package org.qbicc.graph;
 
+import java.util.NoSuchElementException;
 import java.util.Objects;
 
 import org.qbicc.type.ReferenceType;
@@ -10,7 +11,8 @@ import org.qbicc.type.definition.element.ExecutableElement;
  * This node is used to implement two distinct kinds of dynamic checks:
  * arraystorechecks and checkcasts.
  */
-public final class CheckCast extends AbstractValue implements CastValue {
+public final class CheckCast extends AbstractValue implements CastValue, OrderedNode {
+    private final Node dependency;
     /**
      * The input value, which must be of type ReferenceType
      */
@@ -41,14 +43,20 @@ public final class CheckCast extends AbstractValue implements CastValue {
         }
     }
 
-    CheckCast(final Node callSite, final ExecutableElement element, final int line, final int bci, final Value input, final Value toType,
+    CheckCast(final Node callSite, final ExecutableElement element, final int line, final int bci, final Node dependency, final Value input, final Value toType,
               final Value toDimensions, CastType kind, ReferenceType type) {
         super(callSite, element, line, bci);
+        this.dependency = dependency;
         this.input = input;
         this.toType = toType;
         this.toDimensions = toDimensions;
         this.type = type;
         this.kind = kind;
+    }
+
+    @Override
+    public Node getDependency() throws NoSuchElementException {
+        return dependency;
     }
 
     public Value getInput() {
@@ -98,6 +106,7 @@ public final class CheckCast extends AbstractValue implements CastValue {
 
     public boolean equals(final CheckCast other) {
         return this == other || other != null
+            && dependency.equals(other.dependency)
             && input.equals(other.input)
             && toType.equals(other.toType)
             && toDimensions.equals(other.toDimensions)
--- a/compiler/src/main/java/org/qbicc/graph/Node.java
+++ b/compiler/src/main/java/org/qbicc/graph/Node.java
@@ -26,6 +26,7 @@ import org.qbicc.graph.literal.SymbolLiteral;
 import org.qbicc.graph.literal.TypeLiteral;
 import org.qbicc.graph.literal.UndefinedLiteral;
 import org.qbicc.graph.literal.ZeroInitializerLiteral;
+import org.qbicc.graph.schedule.Schedule;
 import org.qbicc.type.definition.element.ExecutableElement;
 import io.smallrye.common.constraint.Assert;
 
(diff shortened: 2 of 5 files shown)


compiler/src/main/java/org/qbicc/graph/CheckCast.java
```diff
@@ -10,7 +11,8 @@
  * This node is used to implement two distinct kinds of dynamic checks:
  * arraystorechecks and checkcasts.
  */
-public final class CheckCast extends AbstractValue implements CastValue {
+public final class CheckCast extends AbstractValue implements CastValue, OrderedNode {
+    private final Node dependency;
```
dgrove-oss, line 15: I think you also need to add handling of `dependency` to `DotNodeVisitor.visit(final Appendable param, final CheckCast node)`
`

ashu-mehra:
`DotNodeVisitor` should also be using `Schedule` to identify the block to which a node belongs to. It would then give the correct view of the graph. I would do that in a separate PR.

compiler/src/main/java/org/qbicc/graph/CheckCast.java
ashu-mehra: Done

dgrove-oss (dismissed):
LGTM; but you might want to wait for @dmlloyd to also take a quick look.  I'm not up on subtle details of how the scheduler works.

compiler/src/main/java/org/qbicc/graph/Node.java
```diff
@@ -162,11 +165,27 @@ public BlockLabel copyBlock(BasicBlock original) {
...
+            }
+        }
+
         public Node copyNode(Node original) {
             if (original instanceof Value) {
                 return copyValue((Value) original);
             } else if (original instanceof Action) {
                 return copyAction((Action) original);
+            } else if (original instanceof ValueHandle) {
+                return copyValueHandle((ValueHandle) original);
```
dmlloyd, line 188: Good catch 😅

compiler/src/main/java/org/qbicc/graph/SimpleBasicBlockBuilder.java
```diff
@@ -363,7 +363,7 @@ public Value checkcast(final Value value, final Value toType, final Value toDime
                 throw new BlockEarlyTermination(unreachable());
             }
         }
-        return new CheckCast(callSite, element, line, bci, value, toType, toDimensions, kind, outputType);
+        return new CheckCast(callSite, element, line, bci, requireDependency(), value, toType, toDimensions, kind, outputType);
```
dmlloyd, line 366:
```suggestion
        return asDependency(new CheckCast(callSite, element, line, bci, requireDependency(), value, toType, toDimensions, kind, outputType));
```

compiler/src/main/java/org/qbicc/graph/schedule/Schedule.java
```diff
@@ -45,15 +56,6 @@ static Schedule forMethod(BasicBlock entryBlock) {
         BlockInfo root = new BlockInfo(entryBlock, 1);
         root.computeIndices(blockInfos, indexHolder);
         final int maxOneBasedIndex = indexHolder[0];
-        if (maxOneBasedIndex == 2) {
```
dmlloyd, line 48: I don't like losing the trivial schedule, but I understand why it's necessary.

dmlloyd (changes requested):
Looks mostly good, but I've suggested one inline change, and you'll need to also modify `org.qbicc.graph.Node.Copier.Terminus#visit(org.qbicc.graph.Node.Copier, org.qbicc.graph.CheckCast)` to add a `param.copyNode(node.getDependency())` beforehand.

compiler/src/main/java/org/qbicc/graph/SimpleBasicBlockBuilder.java
ashu-mehra: Fixed

ashu-mehra:
> you'll need to also modify org.qbicc.graph.Node.Copier.Terminus#visit(org.qbicc.graph.Node.Copier, org.qbicc.graph.CheckCast) to add a param.copyNode(node.getDependency()) beforehand.

Yup, missed that as well. Updated the commit with the suggested changes.

dmlloyd (approved):

Issue #524, opened 2021-05-21
Missing information about prerequisites for qbicc


rsvoboda:
Missing information about prerequisites for qbicc

 - LLVM / how to install it
 - JDK 8+ or 11+?
 - Maven 3.6+?

Note: qbicc-class-library requirement for case-sensitive FS is quite a bit pitta for macOS.

rsvoboda:
I have llvm intslled, qbicc-class-library compiled and installed, but still unable to successfully run `mvn clean install` for qbicc. I'm getting `Compiler terminated with exit code 1` but it's hard to see what is wrong.

2021-05-21 14:33:37.352 ERROR [o.q.t.i.u.QbiccDiagnosticLogger] (accept) error: Linker invocation failed: org.qbicc.machine.tool.CompilationFailureException: Compiler terminated with exit code 1

Was looking at https://github.com/qbicc/qbicc/blob/main/integration-tests/src/it/java/org/qbicc/tests/integration/utils/Qbicc.java. How can I tweak the code to see more details?

DanHeidinga:
Thanks for the feedback @rsvoboda.  There's definitely room for improvement in the documentation of qbicc!

Have you built the class libraries already?  There's a bit of a bootstrap dance where qbicc needs to be built to build the class libraries but non of the qbicc tests will pass till the class libraries are built....

Does something like the approach below get you passed the linker error?
```
$ cd qbicc
$ mvn clean install -DskipTests
$ cd ../qbicc-class-library
$ mvn clean install
$ cd ../qbicc
$ mvn install
```

dmlloyd:
> Note: qbicc-class-library requirement for case-sensitive FS is quite a bit pitta for macOS.

Yes it's quite annoying. Unfortunately there's not really any way to build the OpenJDK classes without it. ☹️

rsvoboda:
I think this can be closed thanks to https://github.com/qbicc/qbicc/pull/525 PR.

PR #525, opened 2021-05-21
Update readme based on user feedback

DanHeidinga:
related: #524


dmlloyd:
Heh, looks like we need to fix the docs no-op check.

dmlloyd (approved):

DanHeidinga:
Address docs no-op in https://github.com/qbicc/qbicc/pull/526

PR #528, opened 2021-05-25
Register constant as type expected by MethodParser

theresa-m:
Conversation details: https://qbicc.zulipchat.com/#narrow/stream/284393-qbicc-development/topic/extend.20failure/near/239635405

### DIFF
--- a/machine/probe/src/main/java/org/qbicc/machine/probe/CProbe.java
+++ b/machine/probe/src/main/java/org/qbicc/machine/probe/CProbe.java
@@ -30,6 +30,7 @@ import org.qbicc.machine.tool.process.InputSource;
 import io.smallrye.common.constraint.Assert;
 import org.qbicc.type.ArrayType;
 import org.qbicc.type.BooleanType;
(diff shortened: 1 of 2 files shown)


dmlloyd (approved):

theresa-m:
I should also add that to reproduce this error I added a ConcurrentHashMap.put(): https://github.com/qbicc/qbicc/commit/ae321be4f7d558758c4d5f6a9414629d8b00ae13#diff-3b61493c859a0eb82a00b6823e588938ed7e04e818e1b1fd33a1d3b378ea7d00R112

dmlloyd:
Ah I see the cause of the failure.  Constants which are only ever checked for defined-ness will have type `CNative.object` typically, which cannot be represented.  So you might have to keep both the old and new way, using the natural type of the constant if a specific type is not given for it.

theresa-m:
Thanks. I'm not able to reproduce this on my mac so I'll try a couple things using ci.

dmlloyd (approved):

PR #530, opened 2021-05-25
More <clinit> groundwork

DanHeidinga:
Original PR was https://github.com/qbicc/qbicc/pull/501, links broken when going public.  See that PR for initial reviews and discussion.

---

Related to #458

* Lower the class initialization status and function pointer to the `<clinit>` method into a new structure:
```
qbicc_clinit_state {
    u8[typeid] clinitState;  // 0 or 1
    void (i8 addrspace(1)*)*[typeid]// <clinit> function pointers
}
```

* Sketch out the class initialization protocol as defined in the JVMS

* Ensure the full set of <clinit> methods are enqueued

* Refactoring of emitting the `qbicc_typeid_array`

* Register `<clinit>` methods as entrypoints so they persist to code generation

* Emit the `qbicc_clinit_state` table

### DIFF
--- a/plugins/instanceof-checkcast/src/main/java/org/qbicc/plugin/instanceofcheckcast/ClassInitializerRegister.java
+++ b/plugins/instanceof-checkcast/src/main/java/org/qbicc/plugin/instanceofcheckcast/ClassInitializerRegister.java
@@ -1,0 +1,54 @@
+package org.qbicc.plugin.instanceofcheckcast;
+
+import java.util.function.Consumer;
+
+import org.qbicc.context.ClassContext;
+import org.qbicc.context.CompilationContext;
+import org.qbicc.plugin.reachability.RTAInfo;
+import org.qbicc.type.definition.DefinedTypeDefinition;
+import org.qbicc.type.definition.LoadedTypeDefinition;
+import org.qbicc.type.definition.element.InitializerElement;
+
+/**
+ * Register all reachable class's class initalizer methods as entrypoints
+ * so they survive to be lowered into functions.
+ * This is required as we lose info after ReachabilityBlockBuilder required
+ * to otherwise keep them alive.
+ * 
+ * Eventually, this will need to skip an class initializer that was run as
+ * part of the build process.
+ */
+public class ClassInitializerRegister implements Consumer<CompilationContext>  {
+
+    @Override
+    public void accept(CompilationContext ctxt) {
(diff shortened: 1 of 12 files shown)


plugins/instanceof-checkcast/src/main/java/org/qbicc/plugin/instanceofcheckcast/ClassInitializerRegister.java
```diff
@@ -0,0 +1,54 @@
...
+        DefinedTypeDefinition jloDef = classContext.findDefinedType("java/lang/Object");
+        LoadedTypeDefinition jlo = jloDef.load();
+
+        // Code below uses #hasMethodBody() to filter out the empty
+        // class initializers and to avoid processing internal arrays
+
+        // Register Object's <clinit> if it exists
+        InitializerElement object_clinit = jlo.getInitializer();
+        if (object_clinit != null && object_clinit.hasMethodBody()) {
+            ctxt.registerEntryPoint(object_clinit);
```
DanHeidinga, line 37: See https://github.com/qbicc/qbicc/pull/501#discussion_r636994590 for original discussion on registering entrypoints

DanHeidinga:
Build is failing with:
```
2021-05-25 17:59:52.009 WARN  [o.q.t.i.u.QbiccDiagnosticLogger] (accept) ./target/it/helloworld/native/org/qbicc/runtime/main/VMHelpers.ll: error: `opt` invocation has failed: org.qbicc.machine.tool.CompilationFailureException: Compiler terminated with exit code 134
```
which of course doesn't reproduce on my machine and we don't capture the `.ll` files in the build

DanHeidinga:
I've confirmed that the large `try / finally` that re-posts the Thread's interrupted state is causing the `opt` failure so I've left that part commented out until I can find a better way to debug the build failure

plugins/instanceof-checkcast/src/main/java/org/qbicc/plugin/instanceofcheckcast/ClassInitializerRegister.java
dmlloyd: The point of that discussion got lost but I'd rather see the reachable initializers be re-enqueued at every stage.  Really what I'd like to do eventually is to completely lock out the `registerEntryPoint` method once compilation begins, to force plugins to both know and control the reachable set at every stage, but we have a few cases where this would be problematic at present.

dmlloyd (approved):

PR #531, opened 2021-05-25
Method handles - API and implementation

dmlloyd:
Move over to a new invocation API, and make invocation targets be `ValueHandle` with a function type.

One key difference between instance field value handles and instance method value handles is that in the former, the instance is a value handle (because a field of an object is an offset into that same object).  In the latter case, the instance is a value rather than a value handle (because the instance of a method is an implicit value argument to the method).


dmlloyd:
This is certain to conflict with #530 so we should probably deal with that PR first, and then I'll rebase this.

DanHeidinga (commented):
A couple of minor questions but nothing stood out as problematic.  Definitely worth another set of eyes on this though

compiler/src/main/java/org/qbicc/graph/SimpleBasicBlockBuilder.java
```diff
@@ -581,14 +639,22 @@ public Node monitorExit(final Value obj) {
...
+
+    <N extends Node> N optionallyTry(N op, boolean ordered) {
         ExceptionHandler exceptionHandler = getExceptionHandler();
         // todo: temporarily disable until exception handlers are fixed
         if (true && exceptionHandler != null) {
             BlockLabel resume = new BlockLabel();
             BlockLabel handler = exceptionHandler.getHandler();
             BlockLabel setupHandler = new BlockLabel();
-            try_(op, resume, setupHandler);
+            //try_(op, resume, setupHandler);
```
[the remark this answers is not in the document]
dmlloyd: Yeah I have a follow-up commit coming - I forgot I was going to wrap this part up after Ashutosh finished his exception fixes.

compiler/src/main/java/org/qbicc/graph/CallNoReturn.java
```diff
@@ -0,0 +1,84 @@
...
+        return Objects.hash(CallNoReturn.class, dependency, target, arguments);
+    }
+
+    @Override
+    public boolean equals(Object other) {
+        return other instanceof CallNoReturn && equals((CallNoReturn) other);
+    }
+
+    public boolean equals(CallNoReturn other) {
+        return this == other || other != null && dependency.equals(other.dependency) && target.equals(other.target) && arguments.equals(other.arguments);
```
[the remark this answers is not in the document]
dmlloyd: I think it's probably redundant.  If the dependency is the same then the terminated block must also be the same, since a terminator can only depend on its block's `BlockEntry` or one of its successors.

dgrove-oss (commented):
Lots of changes, but looks reasonable to me.

One high-level comment is that the naming convention for `call` vs. `invoke` isn't clear to me.  I think the intent is that `call` implies no need for exception handling/block termination while `invoke` means it terminates a basic block and might raise an exception?  Maybe document somewhere?

Also a little murky which concrete call/invoke things should be expected to implement OrderedNode  (Why is Call Ordered while CallNoReturn is not?)

dmlloyd:
> One high-level comment is that the naming convention for `call` vs. `invoke` isn't clear to me. I think the intent is that `call` implies no need for exception handling/block termination while `invoke` means it terminates a basic block and might raise an exception? Maybe document somewhere?

The names `call` and `invoke` mirror the LLVM concepts of `call` and `invoke` which work the same way: `call` is a plain call (exceptions propagated out of the current frame) while `invoke` is a catching call (exceptions trapped in the catch block).

I'll make a note to improve the docs.

> Also a little murky which concrete call/invoke things should be expected to implement OrderedNode (Why is Call Ordered while CallNoReturn is not?)

Terminators are always at the end of the block (so nothing can depend on them) and always depend on the last ordered node, so they themselves are never (or always, depending on how you look at it) Ordered.

dmlloyd:
OK I've fixed the rebase problem and updated the documentation; ready for review (again).

dgrove-oss (dismissed):
LGTM

compiler/src/main/java/org/qbicc/graph/OrderedNode.java
```diff
@@ -1,26 +1,15 @@
...
 /**
  * A node which may be ordered after another node in the program order.
  */
 public interface OrderedNode extends Node {
-    /**
-     * Determine whether this node has a program-ordered predecessor.
-     *
-     * @return {@code true} if this node has a program-ordered predecessor, {@code false} otherwise
-     */
-    default boolean hasDependency() {
```
ashu-mehra, line 14: This was confusing. Glad it is removed and we have a separate node for "no-side-effects" calls.

compiler/src/main/java/org/qbicc/graph/ConstructorElementHandle.java
```diff
@@ -0,0 +1,47 @@
+package org.qbicc.graph;
+
+import org.qbicc.type.ValueType;
+import org.qbicc.type.definition.element.ConstructorElement;
+import org.qbicc.type.definition.element.ExecutableElement;
+
+/**
+ * A handle for an instance constructor.
+ */
+public final class ConstructorElementHandle extends Executable {
```
ashu-mehra, line 10: Shouldn't `ConstructorElementHandle` have a value dependency on `instance`?

dmlloyd: Oo, good catch, I thought I had found all of these.  When I add new nodes, missing the dependencies is generally the number one most common mistake that I make.  😕

dmlloyd:
All fixed.  In retrospect I'm not sure I'm liking the "force push clears reviews" setting... but, here we are.

dgrove-oss (approved):

plugins/dot/src/main/java/org/qbicc/plugin/dot/DotNodeVisitor.java
```diff
@@ -312,17 +306,194 @@ public String visit(final Appendable param, final StaticField node) {
...
 
-    public String visit(final Appendable param, final StaticInvocation node) {
+    // value handles
+
+    public String visit(Appendable param, ConstructorElementHandle node) {
+        String name = register(node);
+        appendTo(param, name);
+        nl(param);
+        attr(param, "label", "constructor\\n" + node.getExecutable());
+        nl(param);
```
ashu-mehra, line 316: I guess it now needs `addEdge(param, node, node.getInstance(), EdgeType.VALUE_DEPENDENCY);` as well.

ashu-mehra:
LGTM

dmlloyd:
There's one more bug here that comes up when more classes are pulled in: if an `Invoke` gets replaced with `Unreachable` during a copy, the return value of the `Invoke` doesn't get mapped. I have no idea how it's happening though, still figuring that out.

dmlloyd:
> There's one more bug here that comes up when more classes are pulled in: if an `Invoke` gets replaced with `Unreachable` during a copy, the return value of the `Invoke` doesn't get mapped. I have no idea how it's happening though, still figuring that out.

I've got that solved, but now there's a problem in the LLVM visitor where the block visiting order can be wrong 🙄

PR #532, opened 2021-05-26
Initial class init handling proposal 

DanHeidinga:
This is a replay of #458 (which already contains valuable discussion) due to PR links being broken by the change from private to public repos.


dgrove-oss:
Closing this as https://github.com/qbicc/qbicc/discussions/764 captures what we eventually ended up instead of what is discussed here.

PR #536, opened 2021-05-27
Mics cleanup of the code base

rsvoboda:
Mics cleanup of the code base

- Fix code listing formatting
- Remove unused imports
- Use enhanced for loop
- Simplify if-else with one line check

### DIFF
--- a/compiler/src/main/java/org/qbicc/graph/InsertElement.java
+++ b/compiler/src/main/java/org/qbicc/graph/InsertElement.java
@@ -3,7 +3,6 @@ package org.qbicc.graph;
 import java.util.Objects;
 
 import org.qbicc.type.ArrayType;
-import org.qbicc.type.ValueType;
 import org.qbicc.type.definition.element.ExecutableElement;
 
 /**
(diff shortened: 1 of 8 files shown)


plugins/llvm/src/main/java/org/qbicc/plugin/llvm/LLVMNodeVisitor.java
```diff
@@ -797,15 +794,15 @@ public LLValue visit(final Void param, final FunctionCall node) {
         LLValue llType = map(functionType);
         LLValue llTarget = map(node.getCallTarget());
         // two scans - once to populate the maps, and then once to emit the call in the right order
-        for (int i = 0; i < arguments.size(); i++) {
```
dmlloyd, line 800: These changes will conflict with #531 which removes this method.

rsvoboda:
Revoked changes in LLVMNodeVisitor, the file was changed as part of "unused imports" and "enhanced for loop" commits so it was not easy to squash new changes to one commit. I think all commits could be squash into one if preferred.

dmlloyd (approved):

rsvoboda:
Rebased, some changes were no longer relevant.

PR #537, opened 2021-05-27
Compile clinit methods before anything else

dmlloyd:
This change fixes the constant ordering issue.  It's based on #531 so it's not as big as it looks... only the last two commits belong to this change.

This might provide a basis to simplify #530 by allowing clinits to be recursively compiled, and maybe re-enqueued as well.

Use this link to see just the two commits: a9e56d6...f305662

### DIFF
--- a/compiler/src/main/java/org/qbicc/type/definition/element/InitializerElement.java
+++ b/compiler/src/main/java/org/qbicc/type/definition/element/InitializerElement.java
@@ -38,34 +38,14 @@ public final class InitializerElement extends BasicElement implements Executable
     }
 
     public MethodBody getMethodBody() {
-        return methodBody;
-    }
-
-    public MethodBody getOrCreateMethodBody() {
         MethodBody methodBody = this.methodBody;
         if (methodBody == null) {
-            MethodBodyFactory factory = this.methodBodyFactory;
-            if (factory != null) {
-                synchronized (this) {
-                    methodBody = this.methodBody;
-                    if (methodBody == null) {
-                        if (inProgress) {
-                            throw new IllegalStateException("Recursive method body creation");
-                        }
-                        inProgress = true;
-                        try {
-                            this.methodBody = previousMethodBody = methodBody = factory.createMethodBody(methodBodyFactoryIndex, this);
-                        } finally {
-                            inProgress = false;
-                        }
-                    }
-                }
-            }
+            throw new IllegalStateException("No method body is present on this element");
         }
         return methodBody;
     }
 
-    public void tryCreateMethodBody() {
+    public boolean tryCreateMethodBody() {
         MethodBody methodBody = this.methodBody;
         if (methodBody == null) {
             MethodBodyFactory factory = this.methodBodyFactory;
@@ -74,7 +54,7 @@ public final class InitializerElement extends BasicElement implements Executable
                     methodBody = this.methodBody;
                     if (methodBody == null) {
                         if (inProgress) {
-                            return;
+                            return true;
                         }
                         inProgress = true;
                         try {
@@ -84,8 +64,11 @@ public final class InitializerElement extends BasicElement implements Executable
                         }
                     }
                 }
+            } else {
+                return false;
             }
         }
+        return true;
     }
 
     public void replaceMethodBody(final MethodBody replacement) {
--- a/compiler/src/main/java/org/qbicc/type/definition/element/InitializerElement.java
+++ b/compiler/src/main/java/org/qbicc/type/definition/element/InitializerElement.java
@@ -19,6 +19,7 @@ public final class InitializerElement extends BasicElement implements Executable
     volatile MethodBody methodBody;
     final int minimumLineNumber;
     final int maximumLineNumber;
+    boolean inProgress;
 
     InitializerElement(Builder builder) {
         super(builder);
@@ -48,12 +49,43 @@ public final class InitializerElement extends BasicElement implements Executable
                 synchronized (this) {
                     methodBody = this.methodBody;
                     if (methodBody == null) {
-                        this.methodBody = methodBody = factory.createMethodBody(methodBodyFactoryIndex, this);
+                        if (inProgress) {
+                            throw new IllegalStateException("Recursive method body creation");
+                        }
+                        inProgress = true;
+                        try {
+                            this.methodBody = previousMethodBody = methodBody = factory.createMethodBody(methodBodyFactoryIndex, this);
+                        } finally {
+                            inProgress = false;
+                        }
                     }
                 }
             }
         }
         return methodBody;
+    }
+
+    public void tryCreateMethodBody() {
+        MethodBody methodBody = this.methodBody;
+        if (methodBody == null) {
+            MethodBodyFactory factory = this.methodBodyFactory;
+            if (factory != null) {
+                synchronized (this) {
+                    methodBody = this.methodBody;
+                    if (methodBody == null) {
+                        if (inProgress) {
+                            return;
+                        }
+                        inProgress = true;
+                        try {
+                            this.methodBody = previousMethodBody = factory.createMethodBody(methodBodyFactoryIndex, this);
+                        } finally {
+                            inProgress = false;
+                        }
+                    }
+                }
+            }
+        }
     }
 
     public void replaceMethodBody(final MethodBody replacement) {
--- a/plugins/verification/src/main/java/org/qbicc/plugin/verification/ClassInitializingBasicBlockBuilder.java
+++ b/plugins/verification/src/main/java/org/qbicc/plugin/verification/ClassInitializingBasicBlockBuilder.java
@@ -1,0 +1,109 @@
+package org.qbicc.plugin.verification;
+
+import org.qbicc.context.CompilationContext;
+import org.qbicc.graph.BasicBlockBuilder;
+import org.qbicc.graph.DelegatingBasicBlockBuilder;
+import org.qbicc.graph.Value;
+import org.qbicc.graph.ValueHandle;
+import org.qbicc.type.ClassObjectType;
+import org.qbicc.type.definition.DefinedTypeDefinition;
+import org.qbicc.type.definition.LoadedTypeDefinition;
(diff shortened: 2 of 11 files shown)


compiler/src/main/java/org/qbicc/type/definition/element/InitializerElement.java
```diff
@@ -48,14 +49,45 @@ public MethodBody getOrCreateMethodBody() {
                 synchronized (this) {
                     methodBody = this.methodBody;
                     if (methodBody == null) {
-                        this.methodBody = methodBody = factory.createMethodBody(methodBodyFactoryIndex, this);
+                        if (inProgress) {
+                            throw new IllegalStateException("Recursive method body creation");
```
DanHeidinga, line 53: This may be too restrictive given that mutually recursive initializers are valid (though can lead to deadlocks).

Is this to defend against the mutually recursive case? or against multi-threaded processing of the initializer? If it's the later, we'll need to implement more of the class initializion process at compile time to make the second attempt wait for the first

plugins/verification/src/main/java/org/qbicc/plugin/verification/ClassInitializingBasicBlockBuilder.java
```diff
@@ -0,0 +1,76 @@
...
+    @Override
+    public Value new_(ClassObjectType type) {
+        initialize(type.getDefinition());
+        return super.new_(type);
+    }
+
+    private void initialize(DefinedTypeDefinition definition) {
+        InitializerElement initializer = definition.load().getInitializer();
+        if (initializer != null) {
+            ctxt.enqueue(initializer);
```
DanHeidinga, line 72: This is insufficient as it misses all the super class and interface initializers.  See https://github.com/qbicc/qbicc/pull/530/commits/7f668e6c9c52be4a359b335a3bbb9205839ae7f0 for details on which initializers need to be enqueued

dmlloyd: Right, I didn't set out to solve that problem (yet) - to start I wanted to solve the immediate problem of constants being defined too late.  But I think this approach could easily be extended to cover those cases.

compiler/src/main/java/org/qbicc/type/definition/element/InitializerElement.java
dmlloyd: This method should only ever be called from `Driver` or in cases where we know that recursion isn't possible.  The  `tryCreateMethodBody` method covers the cycle case by returning when one is detected.  The key is that the *compilation* of a class initializer will be complete before the *compilation* of any access to its members, barring the cyclic case (in which some of one or more initializers might not be completely done initializing before the rest of them are).  In this way it mirrors how class initialization itself works, and the state machine is essentially similar, and I don't think it's really possible to do better than that without a dramatic increase in complexity.

Like class initialization, a deadlock is a very real possibility if two classes reference constants from each others' class in their initializer in the case that the two classes are simultaneously initialized from different threads.  But like the class initialization case, I think we don't have to care too much about that and can just let it fail.

DanHeidinga: > This method should only ever be called from `Driver` or in cases where we know that recursion isn't possible.

Is there some documentation (javadoc?) that indicates this contract?  It would be good to get that recorded somewhere.

> The key is that the _compilation_ of a class initializer will be complete before the _compilation_ of any access to its members, barring the cyclic case (in which some of one or more initializers might not be completely done initializing before the rest of them are).

I'm still unclear on why this is a requirement for most class initializers, apart from the benefits of peephole optimizing away the `getstatic` to drive early DCE.  The critical case in my mind is the `isWindows`-style constants  and, as suggested on the call, I think special casing those native probe classes to occur first makes a lot of the complexity in the rest of this code drop away.

> In this way it mirrors how class initialization itself works, and the state machine is essentially similar,

+1 and to not caring about the deadlock case, as that can be flagged at build time.

> I don't think it's really possible to do better than that without a dramatic increase in complexity.

I think there's less complexity in special casing the few native probe classes and their constants - as they are clearly special - and allowing the rest of this code to sort of fall out in the wash.  Otherwise, we'll be transitioning back and forth between the compiler and the interpreter to get initializers processed and drive the additional class loads.

dmlloyd: > > This method should only ever be called from `Driver` or in cases where we know that recursion isn't possible.
> 
> Is there some documentation (javadoc?) that indicates this contract? It would be good to get that recorded somewhere.

Yeah that's a good idea. I'll add some docs for this.

> > The key is that the _compilation_ of a class initializer will be complete before the _compilation_ of any access to its members, barring the cyclic case (in which some of one or more initializers might not be completely done initializing before the rest of them are).
> 
> I'm still unclear on why this is a requirement for most class initializers, apart from the benefits of peephole optimizing away the `getstatic` to drive early DCE. The critical case in my mind is the `isWindows`-style constants and, as suggested on the call, I think special casing those native probe classes to occur first makes a lot of the complexity in the rest of this code drop away.

It's all about early DCE. The more constant conditions we can detect, the faster we can compile (and of course the more chances we have for later DCE, since we don't iterate through phases more than one time).

I'm not really sure a special case would help here. There will always be another case which benefits from early DCE (first it's platform, then it could be CPU type, then it might be page size, or a constant array size value, or magic numbers like bitmasks, or even complex things like build-time application configuration such as Quarkus uses, etc.).  If you extend a special case to enough situations, eventually it becomes a general case.

> > I don't think it's really possible to do better than that without a dramatic increase in complexity.
> 
> I think there's less complexity in special casing the few native probe classes and their constants - as they are clearly special - and allowing the rest of this code to sort of fall out in the wash. Otherwise, we'll be transitioning back and forth between the compiler and the interpreter to get initializers processed and drive the additional class loads.

I'm not really sure that it'd be less complex in the end. I think it's *generally* easier to maintain these kinds of things as source code than as intrinsics or any other early replacement programmatic mechanism. It seems to me that we'll have to transition back and forth between the compiler and interpreter anyway if we support any kind of build-time execution, since execution of class initialization can cause new classes to be loaded and can call any method which can also load classes, call things reflectively, or even generate classes dynamically to load. I don't believe it would be possible (or advisable) to ensure that all possibly reachable code is pre-compiled. So at least we're already going to be talking about waiting for the compiler to allow initialization to proceed. And if we're supporting constant folded methods like GraalVM, then the compiler will have to execute the method in question during compilation already, so there's another case where the compiler should be able to call into the interpreter. I'm not sure we can reasonably separate these worlds forever.

plugins/verification/src/main/java/org/qbicc/plugin/verification/ClassInitializingBasicBlockBuilder.java
dmlloyd: I've pushed up a modification to initialize the superclass of any classes used.  I *think* that covers all the cases that we need to cover.

dmlloyd:
I pushed up a commit which makes the acquisition of the method body a bit less confusing.

plugins/verification/src/main/java/org/qbicc/plugin/verification/ClassInitializingBasicBlockBuilder.java
DanHeidinga: There's also the interfaces that define default methods that need also need to be initialized per the spec.

This code, copied from ReachabilityBlockBuilder, would cover the interfaces:
```
// Walk all interfaces implemented by the current class and its supers
            // and enqueue their <clinit> if they have a default method
            ltd.forEachInterfaceFullImplementedSet(i -> {
                if (i.declaresDefaultMethods()) {
                    InitializerElement initializer = i.getInitializer();
                    if (initializer != null) {
                        ctxt.enqueue(initializer);
                    }
                }
            });
```

dmlloyd: Ah, I figured I probably missed a case. I was thinking we only needed to init interfaces for static methods, and didn't even think of default methods.

dmlloyd: I updated for this case, but I had to do it a little bit differently because `initialize` is already recursive.

DanHeidinga: Looks right to me

plugins/verification/src/main/java/org/qbicc/plugin/verification/ClassInitializingBasicBlockBuilder.java
```diff
@@ -0,0 +1,97 @@
...
+
+    @Override
+    public ValueHandle instanceFieldOf(ValueHandle instance, FieldElement field) {
+        initialize(field.getEnclosingType());
+        return super.instanceFieldOf(instance, field);
+    }
+
+    @Override
+    public ValueHandle staticField(FieldElement field) {
+        initialize(field.getEnclosingType());
```
DanHeidinga, line 36: There are two cases for static fields and static methods - if the method is defined in an interface, we only initailize that interface.  If it's on a class, then we walk the superclasses and superinterfaces.

Is the intention that this over-registers code and we trim it later?  Should this avoid adding unnecessary initializers?

dmlloyd: No, that wasn't intended. I can tighten it up for static fields and methods of interfaces.

dmlloyd: Fixed.

DanHeidinga (approved):
lgtm

DanHeidinga:
@dmlloyd Anything else you wanted to do to this PR?  Otherwise, I'll merge as is

dmlloyd:
No, I'm happy with it, thanks.

PR #540, opened 2021-06-01
Better coloring for improving readbility of dot graph

kazunoriogata:
This change adds more styles for edges and nodes, so we can visually identify their attributes.

- Use different colors for _then_ and _else_ edges from an _if_ node
- Use different colors for value and cond dependency edges
- Use thick lines for control dependency, and thin lines for data and order dependency
- Use dotted lines for order dependency, and use solid lines for data dependency
- Put colors in start, return, and exception return nodes, so we can easily identify where the graph starts and exits


dmlloyd (approved):
Looks good to me. It will need a rebase however.

PR #545, opened 2021-06-03
RTA: improved reachability analysis for instance methods

### LINKED ISSUE #538: RTA: lazier method reachability
Per the discussion in https://qbicc.zulipchat.com/#narrow/stream/284393-qbicc-development/topic/Windows.20files.20being.20processed.20on.20Mac,  improve RTA by deferring the enqueuing of virtual methods until the defining class has been marked as live.

dgrove-oss:
Fixes #538.

### DIFF
--- a/plugins/reachability/src/main/java/org/qbicc/plugin/reachability/RTAInfo.java
+++ b/plugins/reachability/src/main/java/org/qbicc/plugin/reachability/RTAInfo.java
@@ -245,7 +245,9 @@ public class RTAInfo {
                 rtaLog.debugf("Adding method %s (directly invoked in %s)", target, originalElement);
                 invokableMethods.add(target);
                 ctxt.enqueue(target);
-                processInvokableInstanceMethod(target);
+                if (!target.isPrivate()) {
+                    propagateInvokabilityToOverrides(target);
+                }
             } else {
                 rtaLog.debugf("Deferring method %s (invoked in %s, but no instantiated receiver)", target, originalElement);
                 addReachableClass(definingClass);
@@ -416,12 +418,12 @@ public class RTAInfo {
         initializedTypes.add(type);
     }
 
-    private void processInvokableInstanceMethod(final MethodElement target) {
+    private void propagateInvokabilityToOverrides(final MethodElement target) {
         LoadedTypeDefinition definingClass = target.getEnclosingType().load();
 
         if (definingClass.isInterface()) {
             // Traverse the reachable extenders and implementors and handle as-if we just saw
-            // an invoke of their overriding/implementing method
+            // an invokevirtual/invokeinterface  of their overriding/implementing method
             visitReachableImplementors(definingClass, (c) -> {
                 MethodElement cand = null;
                 if (c.isInterface()) {
@@ -433,7 +435,7 @@ public class RTAInfo {
                     rtaLog.debugf("\tadding method (implements): %s", cand);
                     invokableMethods.add(cand);
                     ctxt.enqueue(cand);
-                    processInvokableInstanceMethod(cand);
+                    propagateInvokabilityToOverrides(cand);
                 }
             });
         } else {
--- a/plugins/reachability/src/main/java/org/qbicc/plugin/reachability/RTAInfo.java
+++ b/plugins/reachability/src/main/java/org/qbicc/plugin/reachability/RTAInfo.java
@@ -1,24 +1,88 @@
 package org.qbicc.plugin.reachability;
 
-import org.qbicc.context.AttachmentKey;
-import org.qbicc.context.CompilationContext;
-import org.qbicc.plugin.layout.Layout;
-import org.qbicc.type.definition.LoadedTypeDefinition;
-
+import java.util.ArrayDeque;
 import java.util.HashSet;
+import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.function.Consumer;
 
+import io.smallrye.common.constraint.Assert;
+import org.jboss.logging.Logger;
+import org.qbicc.context.AttachmentKey;
+import org.qbicc.context.CompilationContext;
+import org.qbicc.plugin.layout.Layout;
+import org.qbicc.type.ClassObjectType;
+import org.qbicc.type.InterfaceObjectType;
+import org.qbicc.type.ObjectType;
+import org.qbicc.type.definition.LoadedTypeDefinition;
+import org.qbicc.type.definition.element.ExecutableElement;
+import org.qbicc.type.definition.element.MethodElement;
 
+/**
+ * RTAInfo tracks the types and methods that have been determined to be reachable
+ * and/or instantiatable from a program entrypoint.
+ *
+ * A class or interface is called reachable if a reference to a
+ * static program element it defines or its java.lang.Class instance
+ * could be executed by an invokable method.
+ *
+ * A class is called instantiated if a `new` of the class (or one of its subclasses)
+ * could be executed by an invokable method.
+ *
+ * A reachable static method is considered to be invokable.
+ * A reachable interface method is considered to be invokable.
+ * A reachable instance method of a class is only considered to be invokable if its defining class is instantiated.
+ *
+ * If a class is reachable, its superclass is also considered to be reachable.
+ * Class reachability does not imply reachability of its implemented interfaces.
+ *
+ * If a class is instantiated, then its superclass is also considered to be instantiated.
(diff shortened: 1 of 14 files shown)


dgrove-oss:
The -rta-debug output looks reasonable, but I'm hitting a similar <clinit> becoming reachable after analyze as #511.  Will regroup in the morning...

plugins/reachability/src/main/java/org/qbicc/plugin/reachability/ReachabilityBlockBuilder.java
```diff
@@ -232,75 +210,95 @@ void enqueueClassInitializers(final LoadedTypeDefinition ltd) {
...
-    private void processInstantiatedClass(final LoadedTypeDefinition type, boolean directlyInstantiated, boolean arrayElement) {
-        RTAInfo info = RTAInfo.get(ctxt);
-        if (!info.isLiveClass(type)) {
+    private void processArrayElementType(ObjectType elemObjectType) {
+        if (elemObjectType instanceof ClassObjectType) {
+            LoadedTypeDefinition elemType = elemObjectType.getDefinition().load();
+            while (elemType != null && !info.isReachableClass(elemType)) {
+                info.addReachableClass(elemType);
+                elemType = elemType.getSuperClass();
+            }
```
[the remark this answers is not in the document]
dgrove-oss: Good catch....I forgot what my own helper method was doing 🤦

plugins/reachability/src/main/java/org/qbicc/plugin/reachability/RTAInfo.java
```diff
@@ -4,21 +4,50 @@
...
+ * A class is called instantiated if a `new` of the class (or one of its subclasses)
+ * could be executed by an invokable method.
+ * If a class is instantiated, then its superclass is also considered to be instantiated.
+ * If a class is instantiated, then its implemented interfaces are considered to be instantiated.
+ * If an interface is instantiated, then its super-interfaces are considered to be instantiated.
+ *
+ * If a class or interface is instantiated, it is also considered to be reachable.
+ *
+ * A reachable static method is considered to be invokable.
+ * A reachable instance method is only considered to be invokable if its defining class is instantiated.
```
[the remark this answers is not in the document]
dgrove-oss: I think this is correct.  When we are processing an instantiated class, we use the method look up routines from LTD, so we shouldn't need to worry about where the methods are coming from.

plugins/reachability/src/main/java/org/qbicc/plugin/reachability/ReachabilityBlockBuilder.java
```diff
@@ -202,150 +199,162 @@ public Value classOf(Value typeId) {
...
-            while (iterator != null) {
-                InitializerElement initializer = iterator.getInitializer();
-                if (initializer != null) {
-                    ctxt.enqueue(initializer);
+        }
+
+        // Annoyingly, because an intermediate interface could be marked initialized due to a static field
+        // access which doesn't cause the initialization of its superinterfaces, we can't short-circuit
+        // this walk up the entire interface hierarchy when we hit an already initialized interfaces.
+        ArrayDeque<LoadedTypeDefinition> worklist = new ArrayDeque<>(List.of(ltd.getInterfaces()));
```
DanHeidinga, line 219: I got tired of writing this code so I added a `void forEachInterfaceFullImplementedSet(Consumer<LoadedTypeDefinition> function);` helper to LoadedTypeDefinition.

I think this is equivalent to:
```java
ltd.forEachInterfaceFullImplementedSet( i -> { 
    if (i.declaresDefaultMethods() && !info.isInitializedType(i)) {
             info.addInitializedType(i);
             if (i.getInitializer() != null) {
                 ctxt.enqueue(i.getInitializer());
             }
     }
};
```

Feel free to ignore this suggestion

plugins/reachability/src/main/java/org/qbicc/plugin/reachability/ReachabilityBlockBuilder.java
```diff
@@ -202,150 +199,162 @@ public Value classOf(Value typeId) {
...
+                            }
                         }
                     }
                 }
             }
 
             // Extend the interface hierarchy
             for (LoadedTypeDefinition i: type.getInterfaces()) {
-                processInstantiatedInterface(info, i, false);
+                processInstantiatedInterface(i);
```
DanHeidinga, line 268: `type.getInterfaces()` returns the locally declared interfaces.  Does this need to walk the hierarchy to find all the interfaces this class can process?

Ah, I see now.  The solution is recursively built up so any interfaces declared on the super classes are handled by the call to `processInstantiatedClass` passing the superclass

plugins/reachability/src/main/java/org/qbicc/plugin/reachability/ReachabilityBlockBuilder.java
```diff
@@ -202,150 +199,162 @@ public Value classOf(Value typeId) {
...
                 info.addInterfaceEdge(type, i);
             }
 
-            // For every enqueued interface method, make sure my implementation of that method is also enqueued.
+            // For every reachable interface method, make sure my implementation of that method is invokable.
             for (LoadedTypeDefinition i: type.getInterfaces()) {
                 for (MethodElement sig: i.getInstanceMethods()) {
-                    if (ctxt.wasEnqueued((sig))) {
+                    if (info.isInvokableMethod(sig)) {
                         MethodElement impl = type.resolveMethodElementVirtual(sig.getName(), sig.getDescriptor());
```
DanHeidinga, line 276: Will this catch overrides of interface methods where the interface is only declared on the superclass?  

```
interface I {
  default void m() { .... }
}

class A implements I {}

class B extends A {
  void m() { /* B specific impl */ }
}

Somewhere: new B();

Somewhere else:  I.m();
```

My hesitation about this case comes from `getInterfaces()` only returning the locally defined interfaces

plugins/reachability/src/main/java/org/qbicc/plugin/reachability/ReachabilityBlockBuilder.java
dgrove-oss: I saw that method, put I didn't think it actually went up the superinterface hierarchy.  pretty sure LTD.getInterfaces() is only one level...not whole crawl up the interface extends hierarchy

plugins/reachability/src/main/java/org/qbicc/plugin/reachability/ReachabilityBlockBuilder.java
dgrove-oss: It should be caught on the superclass.   ie, the superclass should implement that method, therefore it should have an instance method (real or a synthetic abstract one) that this subclass is either using, or overriding.

dgrove-oss: For at least this fairly simple case, it works as expected.  I wrote two test cases, one where I structured the code so RTA found the `new B()` first and one where it found the `I.m()` first.  Looking at the RTA log output, it handled appropriately.

plugins/reachability/src/main/java/org/qbicc/plugin/reachability/RTAInfo.java
```diff
@@ -109,32 +193,262 @@ public void visitLiveImplementors(LoadedTypeDefinition type, Consumer<LoadedType
...
+        if (type.isInterface()) {
+            addReachableInterface(type);
+        } else {
+            addReachableClass(type);
+        }
+        initializedTypes.add(type);
+    }
+
+    private void processInvokableInstanceMethod(final MethodElement target) {
+        LoadedTypeDefinition definingClass = target.getEnclosingType().load();
```
dmlloyd, line 420: I had to make this change locally:

```suggestion
        if (target.isPrivate()) {
            return;
        }
        LoadedTypeDefinition definingClass = target.getEnclosingType().load();
```

dmlloyd: I'm not sure if it's OK to bail at this point though or if the method itself has to be added to `invokableMethods`...

dgrove-oss: I think we should bail out one level up (when we see the invoke);  240e2ef does that instead.   We want the propagation of what was originally an invoke of a non-private method to propagate through any private methods with the same selector it stumbles across as it walks through the class/interface hierarchy looking for overrides.

dmlloyd (approved):

PR #547, opened 2021-06-04
Fix copying of invoke and terminators in general now that we copy in schedule order

dmlloyd:


dgrove-oss (approved):
LGTM -- fixes a problem I was seeing on the branch for #545

compiler/src/main/java/org/qbicc/graph/Node.java
```diff
@@ -134,12 +134,7 @@ public BasicBlock copyProgram() {
                             BasicBlock copiedIncomingBlock = copiedTerminators.get(incomingTerminator);
                             // if this block is null, that means that the copied block can no longer flow into this block due to transformation
                             if (copiedIncomingBlock != null && copiedIncomingBlock.isSucceededBy(ourBlock)) {
-                                Value copiedValue = (Value) copiedNodes.get(val);
-                                // if this value is null, that means big problems
-                                if (copiedValue == null) {
-                                    throw new IllegalStateException("Incoming phi value was not copied");
-                                }
-                                copy.setValueForBlock(ctxt, blockBuilder.getCurrentElement(), copiedIncomingBlock, copiedValue);
+                                copy.setValueForBlock(ctxt, blockBuilder.getCurrentElement(), copiedIncomingBlock, copyValue(val));
```
ashu-mehra, line 137: I guess the call to `copyValue(val)` is to copy `Unschedulable` incoming values, now that `copyOutboundValues` is removed.
Can you please explain why was this broken?

dmlloyd: It breaks essentially because of `Invoke.ReturnValue` in certain kinds of looping situations involving block cancellation (IIRC).  This way the algorithm is simplified and we can rely on schedule order.

PR #548, opened 2021-06-04
Fix forEachInterfaceFullImplementedSet() to actually walk superinterfaces

DanHeidinga:
Dave correctly pointed out that the original implementation only
walked the directly implemented interfaces of the target class
and its superclasses.

It was missing the walk of the interfaces super-interfaces.

I've now corrected the implementation and tightend it to only visit
each interface once.

### DIFF
--- a/compiler/src/main/java/org/qbicc/type/definition/LoadedTypeDefinitionImpl.java
+++ b/compiler/src/main/java/org/qbicc/type/definition/LoadedTypeDefinitionImpl.java
@@ -124,6 +124,9 @@ final class LoadedTypeDefinitionImpl extends DelegatingDefinedTypeDefinition imp
             Collections.addAll(worklist, cur.getInterfaces());
             cur = cur.getSuperClass();
         }
+        if (this.isInterface()) {
+            worklist.add(this);
+        }
         // worklist contains all directly implemented interfaces in the super class chain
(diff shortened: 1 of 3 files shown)


compiler/src/main/java/org/qbicc/type/definition/LoadedTypeDefinitionImpl.java
```diff
@@ -115,13 +118,25 @@ public LoadedTypeDefinition getInterface(final int index) throws IndexOutOfBound
...
-            superClass.forEachInterfaceFullImplementedSet(function);
+        if (this.isInterface()) {
+            worklist.add(this);
+        }
+        // worklist contains all directly implemented interfaces in the super class chain
+        HashSet<LoadedTypeDefinition> visited = new HashSet<>();
+        while (!worklist.isEmpty()) {
+            LoadedTypeDefinition ltd = worklist.pop();
+            if (!visited.contains(ltd)) {
+                visited.add(ltd);
```
dmlloyd, line 135: This can be merged into:
```suggestion
            if (visited.add(ltd)) {
```

dmlloyd (approved):
I have just one minor comment but otherwise it's fine. You can either fix it or just merge, your call :)

PR #549, opened 2021-06-07
reduce clang warnings on MacOS

dgrove-oss:
1. push -pthread option down to the compiler/linker invokers
2. only give -std, -pthread, -input-charset when source language is C

### DIFF
--- a/machine/tool/clang/src/main/java/org/qbicc/machine/tool/clang/ClangCCompilerInvokerImpl.java
+++ b/machine/tool/clang/src/main/java/org/qbicc/machine/tool/clang/ClangCCompilerInvokerImpl.java
@@ -82,7 +82,11 @@ final class ClangCCompilerInvokerImpl extends AbstractClangInvoker implements Cl
         Platform platform = getTool().getPlatform();
         cmd.add("-target");
         cmd.add(platform.getCpu().toString() + "-" + platform.getOs().toString() + "-" + platform.getAbi().toString());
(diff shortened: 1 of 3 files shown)


machine/tool/clang/src/main/java/org/qbicc/machine/tool/clang/ClangCCompilerInvokerImpl.java
```diff
@@ -82,7 +82,11 @@ void addArguments(final List<String> cmd) {
         Platform platform = getTool().getPlatform();
         cmd.add("-target");
         cmd.add(platform.getCpu().toString() + "-" + platform.getOs().toString() + "-" + platform.getAbi().toString());
-        Collections.addAll(cmd, "-std=gnu11", "-f" + "input-charset=UTF-8", "-pipe");
+        if (sourceLanguage == SourceLanguage.C) {
+            Collections.addAll(cmd, "-std=gnu11", "-f" + "input-charset=UTF-8");
+            cmd.add("-pthread");
```
dgrove-oss, line 87: -pthread is definitely not used when compiling ASM files on MacOS.   I suspect this is true in general, but have not verified on other platforms.  It's not entirely clear from the documentation...

dmlloyd: It's weird that they ignore input charset for ASM files, but 🤷🏻‍♂️.

dmlloyd (approved):

PR #551, opened 2021-06-08
Introduce support for "Miranda" methods

dmlloyd:
This partially (but not completely) addresses the problem in #552 as well.

### DIFF
--- a/compiler/src/main/java/org/qbicc/type/definition/DefinedTypeDefinitionImpl.java
+++ b/compiler/src/main/java/org/qbicc/type/definition/DefinedTypeDefinitionImpl.java
@@ -7,6 +7,7 @@ import java.util.Arrays;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
+import java.util.ListIterator;
 import java.util.Map;
 import java.util.Objects;
 
@@ -433,8 +434,8 @@ final class DefinedTypeDefinitionImpl implements DefinedTypeDefinition {
 
     private int findInterfaceImplementations(LoadedTypeDefinition[] interfaces, HashMap<String, HashSet<MethodDescriptor>> visited, HashMap<String, HashMap<MethodDescriptor, List<MethodElement>>> toAdd, int depth) {
         int processed = 0;
-        for (int i = 0; i < interfaces.length; i ++) {
-            processed += findInterfaceImplementations(interfaces[i], visited, toAdd, depth);
+        for (LoadedTypeDefinition interface_ : interfaces) {
+            processed += findInterfaceImplementations(interface_, visited, toAdd, depth);
         }
         return processed;
     }
@@ -444,7 +445,7 @@ final class DefinedTypeDefinitionImpl implements DefinedTypeDefinition {
             return findInterfaceImplementations(currentInterface.getInterfaces(), visited, toAdd, depth - 1);
         } else {
             int cnt = currentInterface.getMethodCount();
-            for (int i = 0; i < cnt; i ++) {
+            search: for (int i = 0; i < cnt; i ++) {
                 MethodElement method = currentInterface.getMethod(i);
                 if (addMethodToVisitedSet(method, visited)) {
                     // we didn't have this one before
@@ -455,8 +456,20 @@ final class DefinedTypeDefinitionImpl implements DefinedTypeDefinition {
                     if (map1 != null) {
                         List<MethodElement> list = map1.get(method.getDescriptor());
                         if (list != null) {
+                            // first make sure that this method element does not override any of the already-known method elements
+                            ListIterator<MethodElement> iter = list.listIterator();
+                            while (iter.hasNext()) {
+                                MethodElement current = iter.next();
+                                if (method.overrides(current)) {
+                                    // remove old candidate
+                                    iter.remove();
+                                } else if (current.overrides(method)) {
+                                    // no action needed since a better choice already exists
+                                    continue search;
+                                }
+                            }
                             // we have another impl option for this
-                            list.add(method);
+                            iter.add(method);
                         }
                     }
                 }
--- a/compiler/src/main/java/org/qbicc/type/definition/DefinedTypeDefinitionImpl.java
+++ b/compiler/src/main/java/org/qbicc/type/definition/DefinedTypeDefinitionImpl.java
@@ -4,10 +4,21 @@ import java.lang.invoke.MethodHandles;
 import java.lang.invoke.VarHandle;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Map;
 import java.util.Objects;
 
 import org.qbicc.context.ClassContext;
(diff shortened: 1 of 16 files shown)


compiler/src/main/java/org/qbicc/type/definition/LoadedTypeDefinition.java
```diff
@@ -210,25 +230,27 @@ default MethodElement resolveMethodElementVirtual(String name, MethodDescriptor
...
             }
         }
 
         int interfaceCount = getInterfaceCount();
         for (int i = 0; i < interfaceCount; i ++) {
-            MethodElement interfaceCandidate = getInterface(i).resolveMethodElementInterface(name, descriptor);
+            MethodElement interfaceCandidate = getInterface(i).resolveMethodElementInterface(true, name, descriptor);
             if (interfaceCandidate != null) {
                 return interfaceCandidate;
             }
```
dmlloyd, line 256: Just an FYI on this bit of code. With this change, it should no longer be necessary, since we've copied all interface methods down on to the class.  I've kept it here for now just because it is an easy place to set a breakpoint to see if it is still being triggered.  So as a note to myself, this should be deleted before the PR is ready for review.

dgrove-oss (approved):
LGTM, with the disclaimer that I am not at language layer level of expertise on how default methods are supposed to work in modern Java.

compiler/src/main/java/org/qbicc/type/definition/DefinedTypeDefinitionImpl.java
```diff
@@ -254,6 +425,88 @@ public LoadedTypeDefinition load() throws VerifyFailedException {
...
+                    // check to see whether we need to add this one to our level
+                    HashMap<MethodDescriptor, List<MethodElement>> map1 = toAdd.get(method.getName());
+                    if (map1 != null) {
+                        List<MethodElement> list = map1.get(method.getDescriptor());
+                        if (list != null) {
+                            // we have another impl option for this
+                            list.add(method);
+                        }
+                    }
+                }
```
DanHeidinga, line 492: For default methods, the first rule regarding using the default implementation is that "the class always wins" so we can skip creating the interface sets if there is a concrete implementation from the current class or its superclasses.

compiler/src/main/java/org/qbicc/type/definition/DefinedTypeDefinitionImpl.java
```diff
@@ -192,9 +207,165 @@ public LoadedTypeDefinition load() throws VerifyFailedException {
...
+                            if (oneOfTheMethods == null) {
+                                oneOfTheMethods = element;
+                            }
+                            if (! element.isAbstract()) {
+                                // found a default method
+                                if (defaultMethod == null) {
+                                    defaultMethod = element;
+                                } else {
+                                    // conflict method! Synthesize a method that throws ICCE
+                                    MethodElement.Builder builder = MethodElement.builder();
```
DanHeidinga, line 256: I think we're also missing a "most specific" check here.  Finding more than 1 method is OK provided one is more specific than another.

Here's an example where a sub-interface is more specific:
```
interface I { default void m() {...} }
interface J { default void m() {...} }
interface K extends I, J { default void m() {...} }

class C extends I, J, K {}
```

We can handle that here or when building the `List<MethodElement>`.  In either case, we need to check each element in the list to see if it's more specific than existing elements and remove them if they are. 

After processing I and J, our list would have `{ I::m, J::m }` but when we add `K::m`, it's more specific than both and should remove both of them from the list.

This could also be handled in `findInterfaceImplementations`

DanHeidinga (commented):
There are two things:
* the list of methods a class provides, which is needed for method resolution
* the vtable-set of methods which is needed for selection
and we've kind of partially mixed them here.

We're increasing our memory pressure by having every class include all the super class methods, and new copies of abstract interface methods as well the newly created conflict methods.

Would it be more space efficient if we kept these concepts separate?  It would take less memory due to smaller sets at each class (though more code to walk the full set) and more closely match the JVM spec resolution rules.

Conversely, building the vtables / itables here would let RTA use the pre-built tables and mark items as "dead".

Right now it feels like we're either doing too much or too little here and they can be consolidated

compiler/src/main/java/org/qbicc/type/definition/DefinedTypeDefinitionImpl.java
dmlloyd: The `toAdd` set should only include methods that were not already present on the class or the superclas (this is done on line 228).  Any method found to already be in the visited set is skipped.  This code right here adds all the possible default method implementations that we've found - but only if we didn't already have an implementation for that class.

compiler/src/main/java/org/qbicc/type/definition/DefinedTypeDefinitionImpl.java
dmlloyd: Notice that the loop starting on 231 is a breadth-first search. Methods are detected by specificity, so in your example we'd hit K first, and then `m` would be in the visited set so it would be excluded when we hit `I` and `J`.

dmlloyd: OK yeah it looks like this case needs to be handled because the interfaces appear at multiple depths.  I just have to work out the algorithmic implication.

dmlloyd: I think handling it during the list building makes the most sense.

DanHeidinga: When I implemented this in OpenJ9, the potential conflict lists tended to be quite small so I didn't worry about the repeated processing of it.

dmlloyd: I have pushed a potential fix.

DanHeidinga: Looks good to me

dmlloyd:
> There are two things:
> 
> * the list of methods a class provides, which is needed for method resolution
> * the vtable-set of methods which is needed for selection
>   and we've kind of partially mixed them here.
> 
> We're increasing our memory pressure by having every class include all the super class methods, and new copies of abstract interface methods as well the newly created conflict methods.

I included superclass methods because I vaguely felt that I ought to.  But TBH I'm not sure there's any benefit to it at all - it would only include abstract methods, and we already have a solution for this in RTA that covers both abstract and non-abstract methods.  I'll remove that chunk.

> Conversely, building the vtables / itables here would let RTA use the pre-built tables and mark items as "dead".
> 
> Right now it feels like we're either doing too much or too little here and they can be consolidated

I think that there might be some consolidation possible.

DanHeidinga:
I'm OK with merging this now and any further changes / consolidation can be done in other PRs.  That work for you @dmlloyd ?

dmlloyd:
Yep that works for me. Thanks!

PR #558, opened 2021-06-09
Include test / snippet name in stderr & stdout output

DanHeidinga:
Output looks like:
```
2021-06-09 10:42:13.756 INFO  [o.q.t.i.u.NativeExecutable] (run) Process(Arrays) standard output:
```

Found this to be useful when trying to diagnose which Snippet test failed.


dgrove-oss (approved):
LGTM

PR #560, opened 2021-06-09
Heap serialization: configurable endianness

dgrove-oss:
Refactor heap serialization plugin to use a ByteBuffer
as the lowlevel output target.  This allows us to
select the endianness at compile time to match the
target platform.

### DIFF
--- a/main/src/main/java/org/qbicc/main/Main.java
+++ b/main/src/main/java/org/qbicc/main/Main.java
@@ -1,6 +1,7 @@
 package org.qbicc.main;
 
 import java.io.IOException;
+import java.nio.ByteOrder;
 import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.Iterator;
@@ -229,6 +230,8 @@ public class Main implements Callable<DiagnosticContext> {
                             // for now, type IDs == int32
                             tsBuilder.setTypeIdSize((int) probeResult.getTypeInfo(int32_t).getSize());
                             tsBuilder.setTypeIdAlignment((int) probeResult.getTypeInfo(int32_t).getAlign());
+                            // todo: endianness probe
+                            tsBuilder.setEndianness(ByteOrder.LITTLE_ENDIAN);
                             if (nogc) {
                                 new NoGcTypeSystemConfigurator().accept(tsBuilder);
                             }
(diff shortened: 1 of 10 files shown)


main/src/main/java/org/qbicc/main/Main.java
```diff
@@ -229,6 +230,8 @@ DiagnosticContext call0(BaseDiagnosticContext initialContext) {
                             // for now, type IDs == int32
                             tsBuilder.setTypeIdSize((int) probeResult.getTypeInfo(int32_t).getSize());
                             tsBuilder.setTypeIdAlignment((int) probeResult.getTypeInfo(int32_t).getAlign());
+                            // todo: endianness probe
+                            tsBuilder.setEndianness(ByteOrder.LITTLE_ENDIAN);
```
dmlloyd, line 234: You should be able to get this value from `probeResult.getByteOrder()` already.

plugins/serialization/src/main/java/org/qbicc/plugin/serialization/HeapOutputStream.java
```diff
@@ -0,0 +1,98 @@
...
+    HeapOutputStream(ByteOrder endianness) {
+        this.endianness = endianness;
+        buffer = new byte[CHUNK_SIZE];
+        out = ByteBuffer.wrap(buffer).order(endianness);
+    }
+
+    private void ensureCapacity(int desired) {
+        if (out.position() + desired >= out.limit()) {
+            byte[] newBuf = new byte[buffer.length + CHUNK_SIZE];
+            System.arraycopy(buffer, 0, newBuf, 0, out.position());
```
dmlloyd, line 27: This could be simplified to `byte[] newBuf = Arrays.copyOf(buffer, buffer.length + CHUNK_SIZE);`.

plugins/serialization/src/main/java/org/qbicc/plugin/serialization/HeapOutputStream.java
```diff
@@ -0,0 +1,98 @@
...
+        } else {
+            lastIndex += 1;
+            objects.put(obj, lastIndex);
+            return -1;
+        }
+    }
+
+    byte[] getBytes() {
+        byte[] ans = new byte[out.position()];
+        System.arraycopy(buffer, 0, ans, 0, out.position());
```
dmlloyd, line 46: This can use `Arrays.copyOfRange`.

main/src/main/java/org/qbicc/main/Main.java
dgrove-oss: done

plugins/serialization/src/main/java/org/qbicc/plugin/serialization/HeapOutputStream.java
dgrove-oss: done

plugins/serialization/src/main/java/org/qbicc/plugin/serialization/HeapOutputStream.java
dgrove-oss: done

dmlloyd (approved):

PR #569, opened 2021-06-10
Introduce interpreter implementation

dmlloyd:
Introduce interpreter implementation, and add a command-line switch to enable build time class initialization.


interpreter/src/main/java/org/qbicc/interpreter/impl/Frame.java
```diff
@@ -0,0 +1,1937 @@
...
+
+    @Override
+    public Object visit(VmThreadImpl param, ObjectLiteral node) {
+        return node.getValue();
+    }
+
+    @Override
+    public Object visit(VmThreadImpl thread, StringLiteral node) {
+        VmClassLoaderImpl classLoader = (VmClassLoaderImpl) element.getEnclosingType().getContext().getClassLoader();
+        return classLoader.getString(thread, node.getValue(), true);
```
DanHeidinga, line 1050: Does this globally intern the string?  All string constants need to be interned but we can add that as a follow on task as well

interpreter/src/main/java/org/qbicc/interpreter/impl/Frame.java
```diff
@@ -0,0 +1,1937 @@
...
+        @Override
+        public Memory visit(Frame frame, StaticField node) {
+            FieldElement variableElement = node.getVariableElement();
+            DefinedTypeDefinition enclosingType = variableElement.getEnclosingType();
+            VmThreadImpl thread = (VmThreadImpl) Vm.requireCurrentThread();
+            ClassContext cc = enclosingType.getContext();
+            VmClassLoaderImpl classLoader = (VmClassLoaderImpl) cc.getClassLoader();
+            if (classLoader == null) {
+                classLoader = thread.getVM().getBootstrapClassLoader();
+            }
```
DanHeidinga, line 1666: We probably want this to be wrapped in an api like `getClassLoaderIncludingBoot()` (with a better name) so we don't have to sprinkle the null check throughout the VM code.

interpreter/src/main/java/org/qbicc/interpreter/impl/VmClassLoaderImpl.java
```diff
@@ -0,0 +1,118 @@
...
+import org.qbicc.type.definition.DefinedTypeDefinition;
+import org.qbicc.type.definition.LoadedTypeDefinition;
+import org.qbicc.type.definition.classfile.ClassFile;
+import org.qbicc.type.descriptor.ClassTypeDescriptor;
+import org.qbicc.type.descriptor.MethodDescriptor;
+
+final class VmClassLoaderImpl extends VmObjectImpl implements VmClassLoader {
+    private final ClassContext classContext;
+    private final Map<ObjectType, VmClassImpl> classes = new ConcurrentHashMap<>();
+    private final Map<String, VmStringImpl> interned = new ConcurrentHashMap<>();
```
DanHeidinga, line 28: This should move the Vm class so it can be shared by all VmClassLoaderImpl

interpreter/src/main/java/org/qbicc/interpreter/impl/VmClassLoaderImpl.java
```diff
@@ -0,0 +1,118 @@
...
+        return findLoadedClass(definition.getType());
+    }
+
+    VmClassImpl loadNewClass(VmThreadImpl thread, VmString intName) {
+        VmClassImpl classLoaderClass = getVmClass();
+        LoadedTypeDefinition clDef = classLoaderClass.getTypeDefinition();
+        ClassTypeDescriptor stringDesc = ClassTypeDescriptor.synthesize(classContext, "java/lang/String");
+        ClassTypeDescriptor classDesc = ClassTypeDescriptor.synthesize(classContext, "java/lang/Class");
+        MethodDescriptor loadClassDesc = MethodDescriptor.synthesize(classContext, classDesc, List.of(stringDesc));
+        return (VmClassImpl) classLoaderClass.getOrCompile(clDef.getMethod(clDef.findMethodIndex("loadClass", loadClassDesc))).invoke(thread, this, intName);
```
DanHeidinga, line 101: I think we also need to record the successfully loaded classes in this classloader's `classes` hashtable to ensure we don't attempt to load the class a second time if we've successfully loaded it once

DanHeidinga (commented):
Some initial comments.  Overall, the approach seems reasonable and the apis look usable.

interpreter/src/main/java/org/qbicc/interpreter/impl/Frame.java
dgrove-oss: The BuildTimeHeap code will intern StringLiterals when it emits them into the native image.  Maybe that is good enough (if we define compile time `equals` between StringLiteral values appropriately)?

DanHeidinga: User code can trip up on this though.... with the interpreter not recognizing two strings as `==` that should be, or with the final native image treating two strings as `==` when it shouldn't.  It's annoying, but I think we need to follow the programmer's intent on this.

The BuildTimeHeap can intern common `char[]'s from strings with the same content but different identity, but it probably shouldn't remove the identity differences.

dgrove-oss: In BuildTimeHeap, I'm interring StringLiteral but not interning instances of java.lang.String that come through as ObjectLiterals.  My assumption was that a StringLiteral would only be coming from a string literal in a class file, and therefore should be interned by the system (while ObjectLiteral would be interned by the program).

DanHeidinga: You're right about BuildTimeHeap being OK to intern `.equals` `StringLiterals`.  The only code I can find that creates `StringLiterals` is in ClassFileImpl and only for constantpool entries.

The interpreter may still treat two intern'd Strings as different if they originated in different classloaders so this still needs to be addressed or `==` can change results between interpreter and compiled image

compiler/src/main/java/org/qbicc/interpreter/VmObject.java
```diff
@@ -10,27 +10,5 @@
...
-    void writeField(int fieldIndex, boolean value);
-
-    void writeField(int fieldIndex, int value);
-
-    void writeField(int fieldIndex, long value);
-
-    void writeField(int fieldIndex, float value);
-
-    void writeField(int fieldIndex, double value);
+    Memory getMemory();
```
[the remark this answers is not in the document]
dmlloyd: It's a per-object `Memory`, so it is GC'd independently from other objects using the host GC.

interpreter/src/main/java/org/qbicc/interpreter/impl/Frame.java
dmlloyd: Yeah only string literals are intentionally interned, and certain other specific cases like class names and so forth.  I do agree that string literals should be interned on a per-class-loader basis.

interpreter/src/main/java/org/qbicc/interpreter/impl/Frame.java
dmlloyd: Yeah I agree, this needs to be improved.

interpreter/src/main/java/org/qbicc/interpreter/impl/VmClassLoaderImpl.java
dmlloyd: Yeah this is a mess right now, and in fact this problem is the reason this is still in draft.  I'm trying out a couple different approaches and we'll see what fits best.

interpreter/src/main/java/org/qbicc/interpreter/impl/Frame.java
DanHeidinga: > The interpreter may still treat two intern'd Strings as different if they originated in different classloaders so this still needs to be addressed or `==` can change results between interpreter and compiled image

Sorry if I wasn't clear - this interpreter behaviour is a bug.  per-class-loader interning of strings is contrary to the spec

dmlloyd: > > The interpreter may still treat two intern'd Strings as different if they originated in different classloaders so this still needs to be addressed or `==` can change results between interpreter and compiled image
> 
> Sorry if I wasn't clear - this interpreter behaviour is a bug. per-class-loader interning of strings is contrary to the spec

Ah okay, well good to know.  I'll get that fixed...

dmlloyd:
Round 2.  Distinct subclasses for each kind of `Class`, more distinct subclasses for different kinds of `Object` with better capabilities.  Fixed class loading, more or less.  Lots of cleanup.

dmlloyd:
Some exciting progress: the interpreter is now running class initialization. There is still much to do but this is major progress.

dmlloyd:
This is now ready for full review. All classes without exception are build-time initialized for now. Enabling the `--init-build-time` switch does result in compilation failure - the majority of classes will fail to initialize due to missing native methods, intrinsics that are not designed for compile time, etc., however these problems need to be tackled incrementally.

plugins/layout/src/main/java/org/qbicc/plugin/layout/Layout.java
```diff
@@ -45,6 +46,18 @@ public static Layout get(CompilationContext ctxt) {
         return layout;
     }
 
+    public static Layout getForInterpreter(CompilationContext ctxt) {
```
dgrove-oss, line 49: Is the separate layout for the interpreter just used for static fields?  Or is it used to account for differences in the instance fields between qbicc and hostVM classes (where such differences exist)?

compiler/src/main/java/org/qbicc/graph/ValueHandleVisitorLong.java
```diff
@@ -0,0 +1,181 @@
+package org.qbicc.graph;
+
+/**
+ *
+ */
+public interface ValueHandleVisitorLong<T> {
```
dgrove-oss, line 6: I didn't see where this new visitor was used.  What am I missing?

dgrove-oss (approved):
Overall makes sense to me.

I believe this PR is building up a "heap" of VmObject instances in the Memory objects associated with the various initialized classes, but that this heap is not yet being written to the executable.  After this is merged, and in parallel with continuing to enhance the interpreter functionality, work can be done to BuildTimeHeap to replace the current set of reflective operations to read the host-VM objects with operations that instead use VmObject APIs to access the Memory for each build-time initialized object and write its contents into initialHeap.ll.  Is that correct, or did I miss something.

compiler/src/main/java/org/qbicc/graph/ValueHandleVisitorLong.java
dmlloyd: It's used by `org.qbicc.interpreter.impl.Frame#getOffset` to get a numeric result for a handle (in this case, what the offset is for the member compared to the base of the `Memory` object) without having to construct a box for the value.

plugins/layout/src/main/java/org/qbicc/plugin/layout/Layout.java
dmlloyd: The difference here is that it's possible for fields to be injected that the interpreter does not know about because the interpreter may need the layout info at an earlier point than would have otherwise been determined. It's possible we could reconcile this at some point though - I've ended up needing to force the injection of most of the interesting injected fields earlier anyway.

The static fields layout method is only used by the interpreter and the results are not cached so the actual Layout instance is not of particular importance for this use case.

compiler/src/main/java/org/qbicc/graph/ValueHandleVisitorLong.java
dgrove-oss: Got it; thanks!  Missed that in the 2132 lines & GitHub had helpfully elided the details of the big diffs when I was doing a search.

plugins/layout/src/main/java/org/qbicc/plugin/layout/Layout.java
dgrove-oss: thanks.  Makes sense.

dgrove-oss:
> That is my intention for how it would work. Maybe we should also look into dropping `StringLiteral` in favor of `ObjectLiteral` as well.

Yeah.  It seems like a easily separable first step would be changing the `StringLiteral.value` field from `java.lang.String` to `VmString`.

dmlloyd:
Do not merge yet; the integration tests are failing but not reporting the failure for some reason.

dmlloyd:
I can't determine why the reachability graph is so much smaller. It may be due to a change in timing or something more nefarious but as the results have been consistent and correct AFAICT, I'm fine with this going in unless anyone has an objection.

ashu-mehra:
@dmlloyd  why do we have the build time init flag disabled by default?

dmlloyd:
> @dmlloyd why do we have the build time init flag disabled by default?

Because if it's enabled, compilation always fails with multiple exceptions in initializers. I will follow up with initializer fixes, and once "hello world" is working with build time init, we can switch the default.

PR #571, opened 2021-06-10
Add llvm-link stage

ashu-mehra:
This patch adds a new stage in the compilation process.
The new stage uses llvm-link to combine all the bitcode files into
a single bitcode file.
This bitcode file is then compiled into assembly, then to an object file
and finally linked to get the executable.
The single bitcode file is named as "qbicc-app.bc" and is created in the
output directory specified by the user.

This step is required so that instruction offset for the stackmap
records are in increasing order, which would allow us to correctly
identify the callsite instruction.

### DIFF
--- a/machine/tool/llvm/src/main/java/org/qbicc/tool/llvm/LlvmLinkInvoker.java
+++ b/machine/tool/llvm/src/main/java/org/qbicc/tool/llvm/LlvmLinkInvoker.java
@@ -1,0 +1,19 @@
+package org.qbicc.tool.llvm;
+
+import org.qbicc.machine.tool.MessagingToolInvoker;
+
+import java.nio.file.Path;
+
+public interface LlvmLinkInvoker extends LlvmInvoker {
+    void addBitcodeFile(Path path);
+
+    default void addBitcodeFiles(Iterable<Path> paths) {
+        for (Path path : paths) {
+            addBitcodeFile(path);
+        }
+    }
+
+    int getBitcodeFileCount();
+
+    Path getBitcodeFile(int index) throws IndexOutOfBoundsException;
+}
--- a/machine/tool/llvm/src/main/java/org/qbicc/tool/llvm/LlvmLinkInvokerImpl.java
+++ b/machine/tool/llvm/src/main/java/org/qbicc/tool/llvm/LlvmLinkInvokerImpl.java
@@ -1,0 +1,44 @@
+package org.qbicc.tool.llvm;
+
+import io.smallrye.common.constraint.Assert;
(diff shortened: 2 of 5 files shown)


dmlloyd:
I want to see how this performs when the big nullability change is in, since this will remove parallelism from a major portion of the build stage.

The other thing that bothers me here is that we're losing per-class assembly output, which was a useful thing. I want to be sure there are no other practical approaches to this problem.

ashu-mehra:
@dmlloyd It seems like we can use partial linking (using `ld`'s `-r` option) to create a fat object file from all the individual object files. This way both the problems would be solved.

ashu-mehra:
There is difference in the disk size of the big fat object file and the final executable in the two approaches.
With partial linking disk size of the files is more.

ashu-mehra:
Most of the difference in disk size is due to debug sections. With debug sections the final exe has a difference of 1 MB. After stripping the debug sections (using `strip --strip-debug`) it comes down to 40 KB.

ashu-mehra:
@dmlloyd this is the [commit](https://github.com/ashu-mehra/qbicc/commit/ef21c5365ead60a3dbda26bff35458bdd756a899) for partial linking. I think this is better than using `llvm-link` as it solves both the issues mentioned [above](https://github.com/qbicc/qbicc/pull/571#issuecomment-859530702). If this looks fine, I can close this and open a new PR.

dmlloyd:
Aren't the instruction offsets relative to the beginning of the function?

ashu-mehra:
yes, the instruction offsets in stackmaps are relative to the function, that is why we need the address (or offset from start of the segment) of the functions in the executable so that we can compute the instruction pointer (address or offset) corresponding to the callsite in the stackmap.

dmlloyd:
It seems to me that we don't actually need the segment offset.  If we know that the offset is relative to the function address, and we know the order that functions are emitted in each object file, and we know the order that the object files will be linked in the final executable, then we know transitively everything we need to know.

In order to construct the tables we need to map IP to function, we mostly need to know the order of the functions so that we can efficiently binary-search the table.  But this ordering does not depend on the address of each instruction overall in the executable; rather it depends on the order that the functions occur in the executables.  The table entries themselves would already be relative to the function's base; for example, we'd emit a constant value of `somefunc + offset` for the IP value rather than (for example) writing an absolute address or a segment offset. The only real ordering requirements are that we emit the function entries in the same order that the functions appear in the executable, which in turn is going to be the order they occur in the object files plus the order of the object files.  Does that make sense?

ashu-mehra:
These change are not required as per the above comment. Closing it.

PR #574, opened 2021-06-14
Enable runtime class initialization

DanHeidinga:
With this commit, runtime class initialization works.  The "is_initialized()" checks are inserted before `get/putstatic` and `invokestatic` bytecodes as well as at `new` bytecodes.

The state machine correctly marks classes as initialized and everything seems to work OK with it.


dmlloyd (approved):
This seems correct to me; you're right, the new node approach does look good.

PR #580, opened 2021-06-15
Introduce core classes plugin; remove corresponding code from Layout

dmlloyd:
This is a cleanup to move the non-layout related code from `Layout`, thereby simplifying that class substantially.  Also adds a number of convenience methods.  The initial PR doesn't change existing usages of the methods on `Layout` which now bridge to `CoreClasses`, so as to minimize disruption.  We can clean them as we go, and then do a final cutover sometime later.

### DIFF
--- a/main/pom.xml
+++ b/main/pom.xml
@@ -78,6 +78,10 @@
         </dependency>
         <dependency>
             <groupId>${project.groupId}</groupId>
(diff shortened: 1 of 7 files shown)


DanHeidinga (approved):
looks ok to me

Issue #578, opened 2021-06-15
Local variable support


dmlloyd:
Introduce support for local variables.  This comes with several pieces.

- [x] Introduce local variable elements
- [x] Introduce value handles for local variables
- [ ] Create local variable elements corresponding to the frame state table in each method (based on slot and type combination)
- [ ] Use load and store operations to populate local variables
- [ ] Introduce LLVM lowering for allocation of variable values at scope-appropriate times
- [ ] Introduce LLVM lowering for loading and storing variable values
- [ ] Map variable debugging information to actual variables
- [ ] Establish "lexical scopes" for variables that are compatible with DWARF
- [ ] Introduce LLVM lowering for local variable debug information

PR #583, opened 2021-06-15
Initial local variable support

dmlloyd:
See #578.  This PR creates elements for each local variable by type and slot. The local variables are accessed via loads and stores, and the backend will stack allocate them, allowing the address of local variables to be taken.  Local variable debug information is added to the output, however the lifetime of the local variables is fixed to the lifetime of the method (meaning that variables will contain garbage until they're initialized).


DanHeidinga:
This seems pretty reasonable to me and my limited knowledge of LLVM debug info.  Once the conflicts are resolved, I'll mark as reviewed and we can get this merged

galderz (approved):

galderz:
The conflicts are now resolved and it looks good to me, so as per @DanHeidinga's comment above, I'm integrating this.

PR #584, opened 2021-06-16
Generate initial heap as collection of globals

dgrove-oss:
An approach to an initial heap that pushes most of the heavy lifting onto llvm & the linker.  We generate a global for each object in the initial heap.  "Pointers" within the initial heap are represented by global variable references that the linker resolves for us.

The actual literal data that is being generated in InitialHeap.ll looks plausible, but I'm struggling with figuring out what the right LLVM types for various things should be (and how to get qbicc to generate them...).  Posting this PR with some snippets of output to get guidance...

In InitialHeap.ll here's a relevant excerpt
```
...
%T.struct.qbicc_initial_heap_obj_183_type = type { i32, i32, [4 x i8] }
...
@qbicc_initial_heap_obj_182 = global %T.java.lang.String { i32 87, i8 0, [3 x i8] zeroinitializer, i8 addrspace(1)* bitcast (%T.struct.qbicc_initial_heap_obj_183_type addrspace(1)* @qbicc_initial_heap_obj_183 to i8 addrspace(1)*), i32 3392903 }, align 8
...
@qbicc_initial_heap_obj_183 = global %T.struct.qbicc_initial_heap_obj_183_type { i32 12, i32 4, [4 x i8] c"null" }, align 8
...
```
This is referenced in AbstractStringBuilder.ll to get to a StringLiteral:
```
@qbicc_initial_heap_obj_182 = external global %T.java.lang.String addrspace(1)*
...
  %L1 = phi i8 addrspace(1)* [ %p0, %B0 ], [ bitcast (%T.java.lang.String addrspace(1)* @qbicc_initial_heap_obj_182 to i8 addrspace(1)*), %B2 ], !dbg !145 ; AbstractStringBuilder.java:631 bci@7
...
```

InitialHeap.ll compiles (although I think the types aren't actually quite what we want....seems like these should be values in addrspace(1), but I can't figure out how to make a Data end up with that annotation).

AbstractStringBuilder.ll fails compilation with the error message:
```
java/lang/AbstractStringBuilder.ll:417:89: error: '@qbicc_initial_heap_obj_182' defined with type '%T.java.lang.String addrspace(1)**' but expected
      '%T.java.lang.String addrspace(1)*'
  %L1 = phi i8 addrspace(1)* [ %p0, %B0 ], [ bitcast (%T.java.lang.String addrspace(1)* @qbicc_initial_heap_obj_182 to i8 addrspace(1)*), %B2 ], !dbg !145 ; A...
```
Which is true, but by manually editing AbstractStringBuilder.ll in a bunch of ways, I still can't figure out what a typing the LLVM is happy with.....so time to ask for help.


plugins/serialization/src/main/java/org/qbicc/plugin/serialization/BuildtimeHeap.java
```diff
@@ -35,39 +47,187 @@ public static BuildtimeHeap get(CompilationContext ctxt) {
...
+        ));
+        SymbolLiteral arrayName = lf.literalOfSymbol(myName, literalCT.getPointer().asCollected());
+        initialHeap.put(arrayName, arrayLiteral);
+
+        return arrayName;
     }
 
-    public List<InitializedField> getHeapRoots() {
-        return heapRoots;
+    private SymbolLiteral serializeArray(boolean[] array) {
```
dgrove-oss, line 194: Leaving the other 8 array cases stubbed out until we get the byte[] case which is used by String to actually do what we want.

dmlloyd:
Addressing the easiest problem first - we should have a flag on `Section` that we can set indicating that it is a heap section.  Then the LLVM generator should examine that flag and, if set, add `addrspace(1)` on to every global variable declared or defined against that section.

I suspect the problem with the pointer types is that the declaration is wrong.  It's using, I guess, the type of the `SymbolLiteral` which is the type of the value of the variable when it is used in the program, not the type of the global symbol.  To work around the problem (until we can nuke `SymbolLiteral`), when you generate the data declaration, you'd have to use the pointee type of the global - or, since it's a reference type, the `ObjectType` (which *should* work... I *think*...).

dgrove-oss:
Thanks for the guidance.  I ended up adding a `addrspace` as a settable attribute on ProgramObject (to parallel threadlocal and linkage).   Programs are compiling and running correctly now.  Doing some cleanup, but should be ready for review soonish...

plugins/serialization/src/main/java/org/qbicc/plugin/serialization/HeapSerializer.java
```diff
@@ -1,426 +1,44 @@
...
-        BlockLabel errorBlock = new BlockLabel();
-        for (int i=0; i<numCases; i++) {
-            targets[i] = new BlockLabel();
-            typeIds[i] = classes[i].getTypeId();
-        }
-
-        bb.begin(new BlockLabel());
-        bb.switch_(original.getParameterValue(0), typeIds, targets, errorBlock);
+        LoadedTypeDefinition ih = ctxt.getBootstrapClassContext().findDefinedType("org/qbicc/runtime/main/InitialHeap").load();
+        Section heapSection = ctxt.getOrAddProgramModule(ih).getOrAddSection(ctxt.IMPLICIT_SECTION_NAME); // TODO: use ctxt.INITIAL_HEAP_SECTION_NAME
```
dgrove-oss, line 36: If I try to use the INITIAL_HEAP_SECTION_NAME here, I get an error from clang backend:
```
fatal error: error in backend: Global variable 'qbicc_initial_heap_obj_0' has an invalid section specifier 'QBICC_I_HEAP': mach-o section specifier
      requires a segment and section separated by a comma.
```

runtime/main/src/main/java/org/qbicc/runtime/main/InitialHeap.java
```diff
@@ -0,0 +1,5 @@
+package org.qbicc.runtime.main;
+
+class InitialHeap {
```
dgrove-oss, line 3: Using this class as a convenient way to get somewhere to put the section with the initial heap.  Could do something else if there's a better way to accomplish this.  It's a massive amount of data, so don't want to put it in a `.ll` file with anything else we might want to look at.

dgrove-oss:
Here's what the generated code looks like now.

A use of a String literal in a `.ll` file:
```
@qbicc_initial_heap_obj_84 = external addrspace(1) global %T.java.lang.String
...
  %L1 = phi i8 addrspace(1)* [ %p0, %B0 ], [ bitcast (%T.java.lang.String addrspace(1)* @qbicc_initial_heap_obj_84 to i8 addrspace(1)*), %B2 ], !dbg !145 ; AbstractStringBuilder.java:631 bci@7
```

The relevant pieces of InitialHeap.ll
```
%T.java.lang.String = type { i32, i8, [3 x i8], i8 addrspace(1)*, i32 }
%T.struct.internal_array_B_4 = type { i32, i32, [4 x i8] }
...
@qbicc_initial_heap_obj_84 = addrspace(1) global %T.java.lang.String { i32 458, i8 0, [3 x i8] zeroinitializer, i8 addrspace(1)* bitcast (%T.struct.internal_array_B_4 addrspace(1)* @qbicc_initial_heap_obj_85 to i8 addrspace(1)*), i32 3392903 }, align 8
@qbicc_initial_heap_obj_85 = addrspace(1) global %T.struct.internal_array_B_4 { i32 12, i32 4, [4 x i8] c"null" }, align 8
```

runtime/main/src/main/java/org/qbicc/runtime/main/InitialHeap.java
dmlloyd: *Not for this PR* but I think the solution here would be to make the enclosing type definition optional on program modules so we can create standalone modules that don't relate to a class.

plugins/serialization/src/main/java/org/qbicc/plugin/serialization/HeapSerializer.java
dmlloyd: I think we can look into this independently of this PR.

dmlloyd (approved):
A very nice diff stat (+284/-1,856)!  Looks good.

dmlloyd:
I'd love to get rid of those bitcasts at some point.  Either we will figure out a way to strongly type everything in LLVM, or we'll move to opaque pointers with LLVM 13 (#586).

PR #585, opened 2021-06-16
Fix bugs in dot graph generation

dmlloyd:
- Avoid stack overflows by giving up on graphs that are too big (print a warning)
- Add missing support for `ClassInitCheck`


dgrove-oss (approved):
LGTM

DanHeidinga (approved):
Thanks for fixing the missing `ClassInitCheck` node in Dot.

PR #593, opened 2021-06-23
Add startMethod BasicBlockBuilder callback

### LINKED ISSUE #591: Add startMethod for capturing method entry
Add a `startMethod` call, symmetrical to `finish`, that captures the method entry invocation. This can be useful for optimisations such as escape analysis, that summarise methods independent of the call sites.

Right now the closest you can get is to catch `begin` and keep a flag to see if it's the first block.

galderz:
Closes #591

### DIFF
--- a/compiler/src/main/java/org/qbicc/graph/SimpleBasicBlockBuilder.java
+++ b/compiler/src/main/java/org/qbicc/graph/SimpleBasicBlockBuilder.java
@@ -55,6 +55,7 @@ final class SimpleBasicBlockBuilder implements BasicBlockBuilder, BasicBlockBuil
     private final ExecutableElement rootElement;
     private Node callSite;
     private BasicBlock terminatedBlock;
+    private boolean started;
 
     SimpleBasicBlockBuilder(final ExecutableElement element, final TypeSystem typeSystem) {
         this.element = element;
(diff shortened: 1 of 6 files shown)


compiler/src/main/java/org/qbicc/type/definition/classfile/ClassFileImpl.java
```diff
@@ -1224,6 +1224,9 @@ public MethodBody createMethodBody(final int index, final ExecutableElement elem
             parameters = ParameterValue.NO_PARAMETER_VALUES;
             currentVarTypes = NO_TYPES;
         }
+
+        gf.startMethod(Arrays.asList(parameters));
```
dmlloyd, line 1228: Just a minor nit, but I'd prefer `List.of()`.

dmlloyd (commented):
There are a couple of call sites missing in `DefinedTypeDefinition`.  Also I'd like to see an exception thrown in `SimpleBasicBlockBuilder` if `begin` is called before `startMethod` was called.

compiler/src/main/java/org/qbicc/graph/SimpleBasicBlockBuilder.java
```diff
@@ -121,6 +121,12 @@ public void setExceptionHandlerPolicy(final ExceptionHandlerPolicy policy) {
         this.policy = policy;
     }
 
+    public void startMethod(List<ParameterValue> arguments) {
+        if (currentBlock != null) {
+            throw new IllegalStateException("begin() called before startMethod()");
+        }
```
dmlloyd, line 127: I'd rather put the check on `begin` because in most cases where this bug will appear, `startMethod` will never be called at all.

dmlloyd (approved):

PR #596, opened 2021-06-25
Fix class init bug and add tests

DanHeidinga:
Fix bug related to mixing up `getElement()` with the more accurate
`getVariableElement()` and `getExecutable()`.

Maybe we need a `getDeclaringClass()` method to make this clearer?

Also, ensure that ObjectModel is treated as initialized at buildtime.

Creating a LoadedTypeDefinitionImpl copies all the superclass and
superinterface methods that aren't overriden by the current class
into the set of instance methods.

The analysis to find locally declared default methods wasn't correctly
filtering out the ones defined in super-interfaces


DanHeidinga:
@theresa-m Thanks for pointing out the issue with the `getElement()` vs `getExecutable()` and proposing the initial fix.  The class init check on `StaticField` had the same kind of issue (also now fixed).

theresa-m (approved):

dmlloyd (approved):

PR #597, opened 2021-06-25
Use unsigned byte for array dimensions

### LINKED ISSUE #581: Array dimensions field should be unsigned byte
The array dimensions field should be an unsigned byte.  The corresponding `VMHelpers` methods would have to be modified to accept a `uint8_t`, which in turn changes their signature for the purposes of lookup.

evacchi:
Should fix #581 :)

### DIFF
--- a/plugins/core-classes/src/main/java/org/qbicc/plugin/coreclasses/CoreClasses.java
+++ b/plugins/core-classes/src/main/java/org/qbicc/plugin/coreclasses/CoreClasses.java
@@ -108,7 +108,7 @@ public final class CoreClasses {
         builder.setEnclosingType(jlcDef);
         builder.setDescriptor(BaseTypeDescriptor.V);
         builder.setSignature(BaseTypeSignature.V);
-        builder.setType(ts.getSignedInteger32Type());
+        builder.setType(ts.getUnsignedInteger8Type());
         field = builder.build();
         jlc.injectField(field);
         classDimensionField = field;
@@ -189,11 +189,11 @@ public final class CoreClasses {
         FieldElement.Builder fieldBuilder = FieldElement.builder();
         fieldBuilder.setEnclosingType(enclosing);
         // TODO: This should be a 8 bit unsigned field. (max dimensions is 255 from multianewarray)
-        fieldBuilder.setDescriptor(BaseTypeDescriptor.I);
-        fieldBuilder.setSignature(BaseTypeSignature.I);
+        fieldBuilder.setDescriptor(BaseTypeDescriptor.V);
+        fieldBuilder.setSignature(BaseTypeSignature.V);
         fieldBuilder.setIndex(index);
         fieldBuilder.setName("dims");
-        fieldBuilder.setType(enclosing.getContext().getTypeSystem().getSignedInteger32Type());
+        fieldBuilder.setType(enclosing.getContext().getTypeSystem().getUnsignedInteger8Type());
         fieldBuilder.setModifiers(ClassFile.ACC_FINAL | ClassFile.ACC_PRIVATE | ClassFile.I_ACC_NO_REFLECT | ClassFile.I_ACC_NO_RESOLVE);
         return fieldBuilder.build();
     }
--- a/runtime/main/src/main/java/org/qbicc/runtime/main/VMHelpers.java
+++ b/runtime/main/src/main/java/org/qbicc/runtime/main/VMHelpers.java
@@ -3,12 +3,14 @@ package org.qbicc.runtime.main;
 import org.qbicc.runtime.CNative;
 import org.qbicc.runtime.NoSideEffects;
 import org.qbicc.runtime.stdc.Stddef;
+import org.qbicc.runtime.stdc.Stdint;
 
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 
 import static org.qbicc.runtime.CNative.*;
 import static org.qbicc.runtime.posix.PThread.*;
+import static org.qbicc.runtime.stdc.Stdint.*;
 import static org.qbicc.runtime.stdc.Stdlib.*;
 
 /**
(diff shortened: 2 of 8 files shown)


runtime/main/src/main/java/org/qbicc/runtime/main/VMHelpers.java
```diff
@@ -69,13 +71,14 @@ private static boolean isAssignableTo(Object value, type_id toTypeId, int toDime
...
+        int intDim = toDimensions.intValue();
+        if (intDim == 0) {
             return isAssignableToLeaf(valueTypeId, toTypeId);
         } else if (ObjectModel.is_reference_array(valueTypeId)) {
-            int valueDims = ObjectModel.dimensions_of(value);
+            uint8_t valueDims = ObjectModel.dimensions_of(value);
             if (valueDims == toDimensions) {
                 return isAssignableToLeaf(ObjectModel.element_type_id_of(value), toTypeId);
-            } else if (valueDims > toDimensions) {
+            } else if (valueDims.intValue() > toDimensions.intValue()) {
```
evacchi, line 81: not sure this is correct, I could not think of a better way

evacchi:
```
2021-06-25T15:41:57.2749398Z org.qbicc.machine.tool.ToolExecutionFailureException: Failed building: `DynamicTypeTests`: org.qbicc.machine.tool.ToolExecutionFailureException: Tool output collection failed: java.io.IOException: Process returned exit code 139

2021-06-25T15:41:57.3251323Z org.qbicc.machine.tool.ToolExecutionFailureException: Failed building: `Arrays`: org.qbicc.machine.tool.ToolExecutionFailureException: Tool output collection failed: java.io.IOException: Process returned exit code 139
```

🤔

evacchi:
ouch. It looks like the native executable is dying 

```
qbicc on  issue-581 is 📦 0.1.0-SNAPSHOT via ☕ v11.0.11
❯ integration-tests/target/it/snippets/DynamicTypeTests/native/a.out
zsh: bus error  integration-tests/target/it/snippets/DynamicTypeTests/native/a.out

qbicc on  issue-581 is 📦 0.1.0-SNAPSHOT via ☕ v11.0.11
❯ integration-tests/target/it/snippets/Arrays/native/a.out
zsh: bus error  integration-tests/target/it/snippets/Arrays/native/a.out
```

ashu-mehra:
I think you are missing couple of changes here:
1. in `CoreIntrinsics` where `get_dimensions_from_class` is registered. Its method descriptor needs to be updated as well:
   `intrinsics.registerIntrinsic(Phase.LOWER, objModDesc, "get_dimensions_from_class", clsInt, getDimFromClass);`
   
2. `get_dimensions_from_class` actually referring to `CoreClasses#classDimensionField` which is also 32-bit int type.

ashu-mehra:
I guess that would also not be sufficient. I see there are two `dimension` fields - `classDimensionField` which is added in `j.l.Class` and another `refArrayDimensionsField` which is added in reference array type. I believe the change of type to `unsigned byte` should apply to both of them.

runtime/main/src/main/java/org/qbicc/runtime/main/VMHelpers.java
```diff
@@ -69,13 +71,14 @@ private static boolean isAssignableTo(Object value, type_id toTypeId, int toDime
...
         // printInt(toDimensions);
         // putchar('\n');
-        if (toDimensions == 0) {
+        int intDim = toDimensions.intValue();
+        if (intDim == 0) {
             return isAssignableToLeaf(valueTypeId, toTypeId);
         } else if (ObjectModel.is_reference_array(valueTypeId)) {
-            int valueDims = ObjectModel.dimensions_of(value);
+            uint8_t valueDims = ObjectModel.dimensions_of(value);
             if (valueDims == toDimensions) {
```
DanHeidinga, line 79: I'd be tempted to keep VMHelpers and ObjectModel viewing array dimensions as an int as they are both "runtime" code and it makes it easier to reason about what they mean.

I highlighted this code as from a pure java view of the world, `==` isn't the right operation here as we're comparing two `uint8_t` objects but it's OK in qbicc because we lower them to the same kind of data.  If we still treat them as `int` in this code, we allow the underlying model to store them as smaller values (good for footprint) and this code to be more obviously correct.

It's only at the boundary - in the ObjectModel intrinsics - we'd need to translate between the two views.

@dmlloyd Any concerns with that approach?

evacchi:
@ashu-mehra I have applied the changes you suggest (I think) but the generated executables still fail in the same way :/ (ERROR 138)

runtime/main/src/main/java/org/qbicc/runtime/main/VMHelpers.java
dmlloyd: I don't think it makes much difference.  The APIs are designed specifically to be able to do this for cases like this, so it's more of a stylistic question at that point.

runtime/main/src/main/java/org/qbicc/runtime/main/VMHelpers.java
dmlloyd: This is correct.  The dimensions values are always between `0` and `256` so the widening casts can be optimized away.

dmlloyd:
You'll probably have to run it in a debugger and see where the bus error is (probably some unaligned access).  The backtrace should pinpoint the line number, unless it's due to an uncaught exception in which case you'll probably have to step through the program to find the failure.

plugins/core-classes/src/main/java/org/qbicc/plugin/coreclasses/CoreClasses.java
```diff
@@ -193,7 +193,7 @@ private static FieldElement makeDimensionsField(final int index, final DefinedTy
         fieldBuilder.setSignature(BaseTypeSignature.I);
```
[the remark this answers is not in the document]
dmlloyd: Probably set them to `V`; that's becoming a convention indicating "Java doesn't care about these types".

runtime/main/src/main/java/org/qbicc/runtime/main/VMHelpers.java
DanHeidinga: I don't feel strongly about this though my preference is to work with Java's primitive types here.  Either way is OK.

ashu-mehra (approved):

ashu-mehra:
lgtm

plugins/core-classes/src/main/java/org/qbicc/plugin/coreclasses/CoreClasses.java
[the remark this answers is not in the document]
dmlloyd: This is my only outstanding comment; other than this it looks OK.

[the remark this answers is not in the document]
evacchi: whoops, forgot about that one!

dmlloyd (approved):

PR #599, opened 2021-06-28
Runtime.availableProcessors intrinsic

theresa-m:
I'm adding this intrinsic to resolve an UnsatisfiedLinkError for progress with object monitors.

### DIFF
--- a/plugins/intrinsics/src/main/java/org/qbicc/plugin/intrinsics/core/CoreIntrinsics.java
+++ b/plugins/intrinsics/src/main/java/org/qbicc/plugin/intrinsics/core/CoreIntrinsics.java
@@ -72,6 +72,7 @@ public final class CoreIntrinsics {
         registerJavaLangObjectIntrinsics(ctxt);
         registerJavaLangNumberIntrinsics(ctxt);
         registerJavaLangFloatDoubleMathIntrinsics(ctxt);
+        registerJavaLangRuntimeIntrinsics(ctxt);
         registerOrgQbiccRuntimeCNativeIntrinsics(ctxt);
         registerOrgQbiccObjectModelIntrinsics(ctxt);
         registerOrgQbiccRuntimeMainIntrinsics(ctxt);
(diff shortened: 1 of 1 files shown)


DanHeidinga:
@theresa-m did you mean to close this?

theresa-m:
Yes. I realized there is a `targetPlatform` that gets passed to the compilation context so I'm going to try to figure out how to get the information from there instead.

dmlloyd:
If this is preventing you from making progress, feel free to just have it return a constant like `4` for now.

theresa-m:
Sounds good. I've just added a default into Cpu for now to make it easier to find in the future.

machine/arch/src/main/java/org/qbicc/machine/arch/Cpu.java
```diff
@@ -28,10 +28,19 @@
     public static final Cpu PPC64 = new Cpu(8, "ppc64");
 
     private final int wordSize;
+    // TODO add an option to get this information from the user
+    private final int numProcessors;
+
+    Cpu(final int wordSize, final int numProcessors, final String name, final String... aliases) {
+        super(name, aliases);
+        this.wordSize = wordSize;
+        this.numProcessors = numProcessors;
```
dmlloyd, line 37: This would not be the correct place for this; the `Cpu` here gives information about the target CPU architecture, not the specific chip or run time configuration.  In practice, the available processors of the system will be governed by the operating system and can vary based on not only the physical device installed in the system but also the configuration of things like CGroups, etc.

The implementation of the `availableProcessors` method in qbicc will lazily compute the constant in the same manner that OpenJDK does.

dmlloyd (changes requested):

machine/arch/src/main/java/org/qbicc/machine/arch/Cpu.java
theresa-m: okay. reverting to my original workaround.

dmlloyd (approved):

PR #600, opened 2021-06-29
Increase CI build timeout from 15 to 20 minutes :(

dmlloyd:


dgrove-oss (approved):
Won't say LGTM, but clearly we need it 😿

DanHeidinga (approved):
lgtm - noticing the same thing (but worse) on my mac.

Is 20 min enough?

DanHeidinga:
Clearly, this is something we're all feeling :)

dmlloyd:
CI just started; the maven step is just barely tipping 15 minutes now.  At some point soon we might want to investigate ways to speed up LLC though: perhaps we're doing more optimizations than we need to be doing.

PR #602, opened 2021-07-01
BuildTimeHeap:  complete array support

dgrove-oss:
1. Implement support for the 7 other primitive array types and ref array
2. Somewhat reduce usage of SymbolLiteral in BuildTimeHeap
3. Temporarily serialize a small dummy object graph as a side-effect of serializing the first String for testing purposes.


main/src/main/java/org/qbicc/main/Main.java
```diff
@@ -378,7 +377,6 @@ DiagnosticContext call0(BaseDiagnosticContext initialContext) {
                                 }
                                 builder.addBuilderFactory(Phase.LOWER, BuilderStage.INTEGRITY, LowerVerificationBasicBlockBuilder::new);
 
-                                builder.addPreHook(Phase.GENERATE, new HeapSerializer());
```
dgrove-oss, line 381: BuildTimeHeap now generates the Data declarations incrementally into the right Section.  So we don't need an extra hook just for that.

DanHeidinga (approved):
looks reasonable to me

PR #603, opened 2021-07-02
improvements to RTA reachability logging & statistics

dgrove-oss:


compiler/src/main/java/org/qbicc/type/definition/element/FieldElement.java
```diff
@@ -16,6 +16,14 @@
         this.initialValue = builder.initialValue;
     }
 
+    public String toString() {
+        final String packageName = getEnclosingType().getDescriptor().getPackageName();
+        if (packageName.isEmpty()) {
+            return getEnclosingType().getDescriptor().getClassName()+"."+getName();
+        }
+        return packageName+"."+getEnclosingType().getDescriptor().getClassName()+"."+getName();
```
DanHeidinga, line 24:
```suggestion
        final ClassTypeDescriptor desc = getEnclosingType().getDescriptor();
        final String packageName = desc.getPackageName();
        if (packageName.isEmpty()) {
            return desc.getClassName()+"."+getName();
        }
        return packageName + "." + desc.getClassName() + "." + getName();
```

DanHeidinga: Treat this is as a suggestion - willing to merge without it as well

DanHeidinga (approved):

DanHeidinga:
Sorry, my proposed change missed adding the import to FieldElement:
```java
import org.qbicc.type.descriptor.ClassTypeDescriptor;
```

PR #609, opened 2021-07-13
Use a proper map for getVirutalFunctions

michihirohorie:
This change uses the virtualFunctions map for the getVirtualFunctions().


dmlloyd (approved):
The change is fine as it is, however `getVirtualFunction` is no longer used since we've moved to vtable dispatch instead.

PR #610, opened 2021-07-13
Embed pthread mutex in object

theresa-m:
Embed pthread mutex in every object as a 64 bit slot. Set will be done with CAS when it can be lowered in a separate PR.


theresa-m:
fyi @DanHeidinga this is what I have so far to get class initialization and object monitors working nicely together at startup. I'm not sure about the best way to solve the third bullet point though.

DanHeidinga:
Discussed with Theresa - the bootstrapping issues in using a hashmap are becoming more of a burden than a help.  We're pivoting to add an object monitor slot to the object header to hold the monitor pointer.  This will avoid the boot cycles where the data structures (HashMap, etc) used to implement `synchronized` support are in turn needed for synchronization in `VMHelpers::initialize_class`

theresa-m:
Updated to embed a slot for a pthread mutex in each object. In progress list in description.

plugins/intrinsics/src/main/java/org/qbicc/plugin/intrinsics/core/CoreIntrinsics.java
```diff
@@ -962,6 +965,34 @@ static void registerOrgQbiccObjectModelIntrinsics(final CompilationContext ctxt)
...
+        // public static native PThread.pthread_mutex_t_ptr nativeObjectMonitor_of(Object reference);
+        MethodDescriptor nomOfDesc = MethodDescriptor.synthesize(classContext, pthreadMutexDesc, List.of(objDesc));
+        StaticIntrinsic nomOf = (builder, target, arguments) -> {
+            Value mutexSlot = builder.load(builder.instanceFieldOf(builder.referenceHandle(arguments.get(0)), nativeObjectMonitorField), MemoryAtomicityMode.NONE);
+            PointerType returnType = (PointerType)target.getType().getReturnType();
+            return builder.valueConvert(mutexSlot, returnType);
+        };
+        intrinsics.registerIntrinsic(Phase.LOWER, objModDesc, "nativeObjectMonitor_of", nomOfDesc, nomOf);
+
+        // public static native Stdint.uint8_t set_nativeObjectMonitor(Object object, PThread.pthread_mutex_t_ptr nom);
```
DanHeidinga, line 979: Why not return a java boolean here?

runtime/main/src/main/java/org/qbicc/runtime/main/ObjectModel.java
```diff
@@ -174,4 +175,21 @@ public static boolean has_default_methods(CNative.type_id typeId) {
...
+    public static native PThread.pthread_mutex_t_ptr nativeObjectMonitor_of(Object reference);
+
+    /**
+     * Set the native object monitor (mutex) for the referenced object. This method is atomic and will return true on success.
+     *
+     * @param reference the object reference (must not be {@code null})
+     * @param nom mutex for the referenced object (must not be {@code null})
+     * @return true if successful
+     */
+    public static native boolean set_nativeObjectMonitor(Object reference, PThread.pthread_mutex_t_ptr nom);
```
DanHeidinga, line 194: Can the two functions be named consistently?  I think it makes more sense to have 
```
nativeObjectMonitor_get
nativeObjectMonitor_set
```
then the current `nom_of` and `set_nom` mismatch

runtime/main/src/main/java/org/qbicc/runtime/main/VMHelpers.java
```diff
@@ -111,48 +104,71 @@ private static boolean isAssignableToLeaf(type_id valueTypeId, type_id toTypeId)
...
+                }
+                ptr<pthread_mutexattr_t> attr = (ptr<pthread_mutexattr_t>) castPtr(attrVoid, pthread_mutexattr_t.class);
 
-        Stddef.size_t mutexSize = sizeof(pthread_mutex_t.class);
-        ptr<?> mVoid = malloc(word(mutexSize.longValue()));
-        if (attrVoid.isNull()) {
-            throw new OutOfMemoryError(/*"Allocation failed"*/);
+                Stddef.size_t mutexSize = sizeof(pthread_mutex_t.class);
+                nom = malloc(word(mutexSize.longValue()));
+                ptr<?> mVoid = malloc(word(mutexSize.longValue()));
```
DanHeidinga, line 136: Why are there two mallocs here?  Only one is null checked and it looks like the one that isn't checked is always leaked

runtime/main/src/main/java/org/qbicc/runtime/main/VMHelpers.java
```diff
@@ -111,48 +104,71 @@ private static boolean isAssignableToLeaf(type_id valueTypeId, type_id toTypeId)
...
-        ptr<?> attrVoid = malloc(word(mutexAttrSize.longValue()));
-        if (attrVoid.isNull()) {
-            throw new OutOfMemoryError(/*"Allocation failed"*/);
+        if (object == null) {
+            /* TODO skip for now. Object should never be null except that
+                classof_from_typeid is not currently implemented. */
+            return;
         }
-        ptr<pthread_mutexattr_t> attr = (ptr<pthread_mutexattr_t>)castPtr(attrVoid, pthread_mutexattr_t.class);
+        while (true) { /* facilitate restart if atomic set operation fails */
```
DanHeidinga, line 125: We don't need to restart here if the CAS fails.  The usual pattern is:
```
previousValue = CAS(location, oldValue, newValue);
if (previousValue != oldValue) {
  // someone else installed a different value for oldValue
  newValue = previousValue;
}
use newValue as the result
```
Different encodings of CAS have different semantics on how they return the result but we just need a way to tell if the mutex slot has transitioned from null to something as the something should stick in that slot for the life of the object.  (Though later we can look at deflating and removing monitors as well)

plugins/core-classes/src/main/java/org/qbicc/plugin/coreclasses/CoreClasses.java
```diff
@@ -89,6 +90,18 @@ private CoreClasses(final CompilationContext ctxt) {
         jlo.injectField(field);
         objectTypeIdField = field;
 
+        // inject a field to hold the object pthread_mutex_t
+        builder = FieldElement.builder();
+        builder.setModifiers(ClassFile.ACC_PUBLIC | ClassFile.ACC_FINAL | ClassFile.I_ACC_NO_REFLECT | ClassFile.I_ACC_NO_RESOLVE);
```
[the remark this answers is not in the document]
theresa-m: thanks for catching. thats a mistake.

plugins/intrinsics/src/main/java/org/qbicc/plugin/intrinsics/core/CoreIntrinsics.java
theresa-m: It does now. This comment is out of date.

runtime/main/src/main/java/org/qbicc/runtime/main/VMHelpers.java
theresa-m: thanks. forgot to remove this

theresa-m:
Thanks for the review @DanHeidinga . I'll move this out of draft mode and disabled object monitors for now since lowering CAS will be a separate pull request.

theresa-m:
CAS-ing a new pthread mutex into an objects slot is working with https://github.com/qbicc/qbicc/pull/619 and https://github.com/qbicc/qbicc/pull/620

runtime/main/src/main/java/org/qbicc/runtime/main/VMHelpers.java
```diff
@@ -111,48 +104,68 @@ private static boolean isAssignableToLeaf(type_id valueTypeId, type_id toTypeId)
...
+            ptr<pthread_mutexattr_t> attr = (ptr<pthread_mutexattr_t>) castPtr(attrVoid, pthread_mutexattr_t.class);
 
-        Stddef.size_t mutexSize = sizeof(pthread_mutex_t.class);
-        ptr<?> mVoid = malloc(word(mutexSize.longValue()));
-        if (attrVoid.isNull()) {
-            throw new OutOfMemoryError(/*"Allocation failed"*/);
+            Stddef.size_t mutexSize = sizeof(pthread_mutex_t.class);
+            ptr<?> mVoid = malloc(word(mutexSize.longValue()));
+            if (mVoid.isNull()) {
+                throw new OutOfMemoryError(/*"Allocation failed"*/);
```
DanHeidinga, line 136:
```suggestion
                free(attrVoid);
                throw new OutOfMemoryError(/*"Allocation failed"*/);
```

theresa-m: Thanks. committed.

DanHeidinga:
@theresa-m the CI build is failing.  Can you take a look?  It had timed out earlier and I restarted it.  If it's another timeout, there maybe something in the locking?

dgrove-oss:
The CI runs I was doing yesterday were also failing out about 50% of the time.  Sadly, I think we need to up the time limits again.  I took a quick look in the GitHub UI,  but didn't see a way for me to change it.

dgrove-oss:
> The CI runs I was doing yesterday were also timing out about 50% of the time. Sadly, I think we need to up the time limits again. I took a quick look in the GitHub UI, but didn't see a way for me to change it.

Never mind.  Forgot about the .github directory.  #625

theresa-m:
Thanks @dgrove-oss ! Looks like its passing now.

DanHeidinga (approved):

PR #612, opened 2021-07-29
Get an entry at requires_version_index

michihirohorie:
This fix uses the requires_version_index in the requires table of the module attribute to get the version of the specified module.


ashu-mehra (approved):

ashu-mehra:
LGTM. Fix is straight forward, so I am merging it.

PR #613, opened 2021-07-30
Take overall alignment into consideration when determining size of CompoundType

ashu-mehra:
Take overall alignment into consideration when determining size of CompoundType

### DIFF
--- a/compiler/src/main/java/org/qbicc/type/CompoundType.java
+++ b/compiler/src/main/java/org/qbicc/type/CompoundType.java
@@ -281,7 +281,7 @@ public final class CompoundType extends ValueType {
                 throw new IllegalStateException("CompoundType has no members");
             }
             if (completeType == null) {
-                int size = offset;  // Offset points to the end of the structure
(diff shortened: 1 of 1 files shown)


compiler/src/main/java/org/qbicc/type/CompoundType.java
```diff
@@ -281,7 +281,7 @@ public CompoundType build() {
                 throw new IllegalStateException("CompoundType has no members");
             }
             if (completeType == null) {
-                int size = offset;  // Offset points to the end of the structure
+                int size = (offset + (overallAlign - 1)) & -overallAlign;  // Offset points to the end of the structure, align the size to overall alignment
```
DanHeidinga, line 284: I had to look up the C structure sizing and alignment rules to convince myself that this was correct (and it is).  Including the tail padding in the size helps ensure optimal alignment for arrays so they avoid unaligned accesses on oddly sized structs.

DanHeidinga (approved):

PR #615, opened 2021-08-04
Use StackMap section name specific to objectfile format

ashu-mehra:
https://llvm.org/docs/StackMaps.html#stack-map-section states Stack Map section name is different for ELF and MachO file formats. This PR makes changes to use the correct name for the Stack Map section based on the file format.


ashu-mehra:
This is also ready for review.

DanHeidinga (approved):

PR #617, opened 2021-08-05
Search all sections in getSection() for named one

DanHeidinga:
The original implementation was using the Segment->Section mapping
and was trying to use the same name for both the segment and the
section.

This failed when looking up the `__llvm_stackmaps` section on macho
as the segment is named `__LLVM_STACKMAPS`.

New algorithm searches the segments list until it finds one of the
matching name and uses that to provide an org.qbicc.machine.object.Section
instances.

Co-authored-by: Ashutosh Mehra <asmehra@redhat.com>


DanHeidinga:
@ashu-mehra can you review this?

ashu-mehra (approved):

PR #619, opened 2021-08-12
CompareAndSwap lowering

theresa-m:
Lowers the CmpAndSwap value to the LLVM IR cmpxchg instruction. 

cmpxchg returns a "literal" structure type and I believe only "identified" llvm structure types are currently supported. The next step will be to correctly lower `extractvalue` to use a literal structure so the result of cmpxchg can be accessed. 
update: https://github.com/qbicc/qbicc/pull/620

### DIFF
--- a/compiler/src/main/java/org/qbicc/graph/CmpAndSwap.java
+++ b/compiler/src/main/java/org/qbicc/graph/CmpAndSwap.java
@@ -41,10 +41,33 @@ public final class CmpAndSwap extends AbstractValue implements OrderedNode {
         if (! target.isReadable()) {
             throw new IllegalArgumentException("Handle is not readable");
         }
+
+        /* expected and update value must have the same type, target must be a pointer to that type. */
+        if (!expectedValue.getType().equals(updateValue.getType())
+             || !expectedValue.getType().equals(target.getPointerType().getPointeeType())
+        ) {
+            throw new IllegalArgumentException("The target, expected and new value types must agree.");
+        };
+
+        /* ordering params must be at least MONOTONIC */
+        if ((0 > successAtomicityMode.compareTo(MemoryAtomicityMode.MONOTONIC))
+            || (0 > failureAtomicityMode.compareTo(MemoryAtomicityMode.MONOTONIC))
+        ) {
+            throw new IllegalArgumentException("Mode must be at least monotonic");
+        }
+        /* volatile does not map to an ordering constraint */
+        if (successAtomicityMode.equals(MemoryAtomicityMode.VOLATILE) || failureAtomicityMode.equals(MemoryAtomicityMode.VOLATILE)) {
+            throw new IllegalArgumentException("volatile does not map to a LLVM ordering constraint");
+        }
+        /* failure ordering must not be RELEASE or ACQUIRE_RELEASE */
+        if (failureAtomicityMode.equals(MemoryAtomicityMode.RELEASE) || failureAtomicityMode.equals(MemoryAtomicityMode.ACQUIRE_RELEASE)) {
+            throw new IllegalArgumentException("Failure mode cannot be release or acquire_release");
+        }
+
     }
 
     int calcHashCode() {
-        return Objects.hash(CmpAndSwap.class, dependency, target, expectedValue, updateValue, successAtomicityMode, failureAtomicityMode);
+        return Objects.hash(CmpAndSwap.class, dependency, target, expectedValue, updateValue, resultType, successAtomicityMode, failureAtomicityMode);
     }
 
     public CompoundType getType() {
@@ -86,7 +109,8 @@ public final class CmpAndSwap extends AbstractValue implements OrderedNode {
     public boolean equals(final CmpAndSwap other) {
         return this == other || other != null && dependency.equals(other.dependency) && target.equals(other.target)
             && expectedValue.equals(other.expectedValue) && updateValue.equals(other.updateValue)
-            && successAtomicityMode == other.successAtomicityMode && failureAtomicityMode == other.failureAtomicityMode;
+            && resultType.equals(other.resultType) && successAtomicityMode == other.successAtomicityMode
+            && failureAtomicityMode == other.failureAtomicityMode;
     }
 
     public int getValueDependencyCount() {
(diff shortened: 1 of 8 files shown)


DanHeidinga (commented):
Seems pretty reasonable to me.  I think down the line we'll want versions of CAS that return the value rather than a boolean but this is a good first step

compiler/src/main/java/org/qbicc/graph/CmpAndSwap.java
```diff
@@ -113,12 +146,30 @@ public static CompoundType getResultType(CompilationContext ctxt, ValueType valu
         CompoundType compoundType = map.get(valueType);
         if (compoundType == null) {
             TypeSystem ts = ctxt.getTypeSystem();
-            compoundType = ts.getCompoundType(CompoundType.Tag.NONE, null, -1, -1, () -> List.of(ts.getCompoundTypeMember(null, valueType, -1, -1), ts.getCompoundTypeMember(null, ts.getBooleanType(), -1, -1)));
-            CompoundType appearing = map.putIfAbsent(valueType, compoundType);
-            if (appearing != null) {
-                compoundType = appearing;
-            }
```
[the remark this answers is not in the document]
theresa-m: Yes thanks! I'll add those back.

theresa-m:
@DanHeidinga is this ready to merge?

compiler/src/main/java/org/qbicc/graph/CmpAndSwap.java
```diff
@@ -113,12 +146,34 @@ public static CompoundType getResultType(CompilationContext ctxt, ValueType valu
         CompoundType compoundType = map.get(valueType);
         if (compoundType == null) {
             TypeSystem ts = ctxt.getTypeSystem();
-            compoundType = ts.getCompoundType(CompoundType.Tag.NONE, null, -1, -1, () -> List.of(ts.getCompoundTypeMember(null, valueType, -1, -1), ts.getCompoundTypeMember(null, ts.getBooleanType(), -1, -1)));
+            compoundType = CompoundType.builder(ts)
+                .setTag(CompoundType.Tag.NONE)
```
DanHeidinga, line 150: How does this connect to the new `CompoundType.TAG.INLINE` added in #620?

theresa-m: This line would be replaced with creating an INLINE compound type. Probably makes sense to remove the change from this pull request all together.

theresa-m: I've removed this change since it will be correctly modified in #620

plugins/intrinsics/src/main/java/org/qbicc/plugin/intrinsics/core/CoreIntrinsics.java
```diff
@@ -1017,9 +1022,50 @@ static void registerOrgQbiccRuntimeValuesIntrinsics(final CompilationContext ctx
...
+                Value update = arguments.get(2);
+                Value result = builder.cmpAndSwap(target, expect, update, successMode, failureMode, isVolatile);
+                Value resultValue = builder.extractMember(result, ((CmpAndSwap)result).getResultValueType());
+                /* set was successful when expected value is returned */
+                return builder.isEq(resultValue, expect);
+            }
+        }
+
+        StaticIntrinsic compareAndSwapVolatile = new CompareAndSwapIntrinsic(MemoryAtomicityMode.MONOTONIC, MemoryAtomicityMode.MONOTONIC, true);
+        StaticIntrinsic compareAndSwapAcquire = new CompareAndSwapIntrinsic(MemoryAtomicityMode.ACQUIRE, MemoryAtomicityMode.ACQUIRE, false);
```
DanHeidinga, line 1054: Is ACQUIRE correct for both success and failure?  I'm having some trouble mapping this to the equivalent Unsafe operations but I would have expected the same failure mode for this case and the RELEASE case below.

plugins/intrinsics/src/main/java/org/qbicc/plugin/intrinsics/core/CoreIntrinsics.java
```diff
@@ -1017,9 +1022,50 @@ static void registerOrgQbiccRuntimeValuesIntrinsics(final CompilationContext ctx
...
         intrinsics.registerIntrinsic(valsDesc, "isAlwaysFalse", boolBoolDesc, isAlwaysFalse);
 
-        // todo: compareAndSwap*
 
-        // getAndSet*
+        // compareAndSwap*
+        class CompareAndSwapIntrinsic implements StaticIntrinsic {
+            private final MemoryAtomicityMode successMode;
+            private final MemoryAtomicityMode failureMode;
+            private final boolean isVolatile;
```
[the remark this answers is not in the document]
theresa-m: That makes sense.

plugins/intrinsics/src/main/java/org/qbicc/plugin/intrinsics/core/CoreIntrinsics.java
theresa-m: I'm not sure about this one but agree it makes sense to be consistent.

plugins/intrinsics/src/main/java/org/qbicc/plugin/intrinsics/core/CoreIntrinsics.java
```diff
@@ -1017,9 +1022,48 @@ static void registerOrgQbiccRuntimeValuesIntrinsics(final CompilationContext ctx
...
+                Value expect = arguments.get(1);
+                Value update = arguments.get(2);
+                Value result = builder.cmpAndSwap(target, expect, update, successMode, failureMode);
+                Value resultValue = builder.extractMember(result, ((CmpAndSwap)result).getResultValueType());
+                /* set was successful when expected value is returned */
+                return builder.isEq(resultValue, expect);
+            }
+        }
+
+        // TODO remove is volatile and always set to true
```
DanHeidinga, line 1050: Is this TODO already handled?

DanHeidinga (approved):
apart from the comment nitpick, this looks reasonable to me

plugins/intrinsics/src/main/java/org/qbicc/plugin/intrinsics/core/CoreIntrinsics.java
theresa-m: yes :) thanks. removed

PR #620, opened 2021-08-17
Lower literal structs for extractvalue

theresa-m:
LLVM literal and identified structure types are not interchangeable. A literal type is needed to correctly interpret the return value of cmpxchg.

change relates to: https://github.com/qbicc/qbicc/pull/619

### DIFF
--- a/compiler/src/main/java/org/qbicc/type/CompoundType.java
+++ b/compiler/src/main/java/org/qbicc/type/CompoundType.java
@@ -200,6 +200,7 @@ public final class CompoundType extends ValueType {
         NONE("untagged"),
         STRUCT("struct"),
         UNION("union"),
+        INLINE("inline"), /* LLVM literal structure type */
         ;
         private final String string;
 
@@ -254,6 +255,10 @@ public final class CompoundType extends ValueType {
             }
             overallAlign = align;
             return this;
+        }
+
+        public Builder addNextMember(final ValueType type) {
+            return addNextMember("", type, type.getAlign());
         }
 
         public Builder addNextMember(final String name, final ValueType type) {
--- a/machine/llvm/src/main/java/org/qbicc/machine/llvm/impl/ExtractValueImpl.java
+++ b/machine/llvm/src/main/java/org/qbicc/machine/llvm/impl/ExtractValueImpl.java
@@ -41,11 +41,8 @@ final class ExtractValueImpl extends AbstractYieldingInstruction implements Extr
         aggregateType.appendTo(target);
         target.append(' ');
         aggregate.appendTo(target);
-        ArgImpl lastArg = this.lastArg;
-        if (lastArg != null) {
-            lastArg.appendTo(target);
-        }
-        return target;
+        lastArg.appendTo(target);
+        return appendTrailer(target);
     }
 
     static final class ArgImpl extends AbstractEmittable {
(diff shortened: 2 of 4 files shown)


machine/llvm/src/main/java/org/qbicc/machine/llvm/impl/ExtractValueImpl.java
```diff
@@ -41,11 +41,8 @@ public Appendable appendTo(final Appendable target) throws IOException {
         aggregateType.appendTo(target);
         target.append(' ');
         aggregate.appendTo(target);
-        ArgImpl lastArg = this.lastArg;
-        if (lastArg != null) {
```
theresa-m, line 45: a minimum of one index is required.

compiler/src/main/java/org/qbicc/type/CompoundType.java
```diff
@@ -200,6 +200,7 @@ public int compareTo(final Member o) {
         NONE("untagged"),
         STRUCT("struct"),
         UNION("union"),
+        INLINE("inline"), /* LLVM literal structure type */
```
DanHeidinga, line 203: Are there other instances of INLINE structures in LLVM?  I'm not clear on whether this is a special case for CAS or a more general feature

theresa-m: A _literal_ structure is a general LLVM feature although I'm not sure where else it would be used since `cmpxchg` is the only LLVM instruction that explicitly returns one. The other enum types here will be lowered to structures with names making them _identified_ structures in LLVM. The difference is mentioned here [1].
If an identified type is used to extract the return value from `cmpxchg` there is a compiler error saying it expects the literal type. I believe this is because only literal types can be uniqued.
1. https://llvm.org/docs/LangRef.html#structure-types

DanHeidinga: My preference would be to not expose INLINE as a TAG that can be added in general as it seems like a fairly special case.  Unfortunately, I don't see a clean way to do that without completely changing how we've modelled the CAS instruction.

To hide the INLINE type in the CAS instruction, we'd need to mandate that the return value is the previous value from memory (`P`).  If `P == oldValue`, then it'd be success.  We'd lose the boolean returned in the inlined structure.

I've been slow with the review while looking for an alternative way to do this and I think what you've done here is the best alternative we're going to find. 

tldr: I'm OK with exposing INLINE as the best practical solution to move forward.

compiler/src/main/java/org/qbicc/type/CompoundType.java
```diff
@@ -256,6 +257,10 @@ public Builder setOverallAlignment(int align) {
             return this;
         }
 
+        public Builder addNextMember(final ValueType type) {
+            return addNextMember("", type, type.getAlign());
```
DanHeidinga, line 261: Should this only be applicable to INLINE tagged CompoundTypes?  This is not a leading question - I don't know whether that's a useful restriction or not.

DanHeidinga (approved):
This looks reasonable to me.  There's one pending question about API restrictions but it won't change the overall approach so approving now

compiler/src/main/java/org/qbicc/type/CompoundType.java
theresa-m: I think it makes sense to restrict this to INLINE only. Otherwise getMember(String name) may not retrieve the expected result. I will create a new pr for that.

theresa-m: Here is the change: https://github.com/qbicc/qbicc/pull/624

PR #623, opened 2021-08-19
Ensure 16 byte alignment of struct_UnwindException

dgrove-oss:
Fixes a GPF when throwing exceptions on BigSur because the unwindException struct element of the VMThread object was only being aligned to 8 bytes, not 16 as required by the libunwind specification.

### DIFF
--- a/plugins/native/src/main/java/org/qbicc/plugin/native_/NativeInfo.java
+++ b/plugins/native/src/main/java/org/qbicc/plugin/native_/NativeInfo.java
@@ -24,6 +24,7 @@ import org.qbicc.type.CompoundType;
 import org.qbicc.type.TypeSystem;
 import org.qbicc.type.ValueType;
 import org.qbicc.type.annotation.Annotation;
+import org.qbicc.type.annotation.IntAnnotationValue;
 import org.qbicc.type.annotation.StringAnnotationValue;
 import org.qbicc.type.annotation.type.TypeAnnotation;
 import org.qbicc.type.annotation.type.TypeAnnotationList;
@@ -128,6 +129,7 @@ final class NativeInfo {
                         String simpleName = null;
                         Qualifier q = Qualifier.NONE;
                         boolean incomplete = false;
+                        int annotatedAlign = 0;
                         for (Annotation annotation : definedType.getVisibleAnnotations()) {
                             ClassTypeDescriptor annDesc = annotation.getDescriptor();
                             if (ProbeUtils.processCommonAnnotation(pb, annotation)) {
@@ -138,6 +140,11 @@ final class NativeInfo {
                                     simpleName = ((StringAnnotationValue) annotation.getValue("value")).getString();
                                 } else if (annDesc.getClassName().equals(Native.ANN_INCOMPLETE)) {
                                     incomplete = true;
+                                } else if (annDesc.getClassName().equals(Native.ANN_ALIGN)) {
+                                    annotatedAlign = ((IntAnnotationValue)annotation.getValue("value")).intValue();
+                                    if (annotatedAlign == Integer.MAX_VALUE) {
+                                        annotatedAlign = ctxt.getTypeSystem().getMaxAlignment();
+                                    }
                                 }
                             }
                         }
@@ -189,13 +196,14 @@ final class NativeInfo {
                                 if (result != null) {
                                     CProbe.Type.Info typeInfo = result.getTypeInfo(probeType);
                                     long size = typeInfo.getSize();
+                                    int align = annotatedAlign != 0 ? annotatedAlign : (int) typeInfo.getAlign();
                                     if (typeInfo.isFloating()) {
                                         if (size == 4) {
                                             resolved = ts.getFloat32Type();
                                         } else if (size == 8) {
                                             resolved = ts.getFloat64Type();
                                         } else {
-                                            resolved = ts.getCompoundType(tag, simpleName, size, (int) typeInfo.getAlign(), List::of);
+                                            resolved = ts.getCompoundType(tag, simpleName, size, align, List::of);
                                         }
                                     } else if (typeInfo.isSigned()) {
                                         if (size == 1) {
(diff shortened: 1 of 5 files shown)


plugins/native/src/main/java/org/qbicc/plugin/native_/NativeInfo.java
```diff
@@ -189,13 +193,14 @@ ValueType resolveNativeType(final DefinedTypeDefinition definedType) {
                                 if (result != null) {
                                     CProbe.Type.Info typeInfo = result.getTypeInfo(probeType);
                                     long size = typeInfo.getSize();
+                                    int align = annotatedAlign != 0 ? annotatedAlign : (int) typeInfo.getAlign();
```
DanHeidinga, line 196: Should we validate that the annotatedAlign is at least as large as `typeInfo.getAlign()`?

dgrove-oss: Not sure if we have to give it much weight, but the comment on the @align annotation was:
```
    /**
     * Explicitly specify the exact alignment of an object.  The alignment will not be probed.  Different alignments can be specified
     * for different platforms.
     */
```
So, I was trying to implement those semantics....if there's an align annotation we just do what it says.  Whether or not that's the right semantics 🤷‍♂️

ashu-mehra:
FWIW I was trying to see the alignment that CProbe computes for `_Unwind_Exception` and it correctly computes it as 16 bytes.
I used following code:
```
#include <stdio.h>
#include <stddef.h>
#include <stdalign.h>
#include <unwind.h>

int main(void)
{
    printf("Alignment of UnwindExceptoin using _Alignof = %zu\n", _Alignof(struct _Unwind_Exception));
    return 0;
}
```

CProbe also uses `_Alignof` operator to get the alignment. This is the code generate by cprobe for `_Unwind_Exception`:

```

#include <stddef.h>
#include <limits.h>
#include <unwind.h>
...
unsigned long tp_overall_align0 = _Alignof(struct _Unwind_Exception);
...
```

Looks like for some reason this technique is not working on BigSur. I am wondering if there are any other native structs that may suffer similar issue on BigSur.

plugins/native/src/main/java/org/qbicc/plugin/native_/NativeInfo.java
DanHeidinga: That makes sense.

dgrove-oss:
@ashu-mehra -- that is quite worrisome.  I tried your C program:

Daves-MacBook-Pro:tests dgrove$ clang probe.c 
Daves-MacBook-Pro:tests dgrove$ ./a.out 
Alignment of UnwindExceptoin using _Alignof = 8

I can verify it isn't working with clang on Big Sur.

ashu-mehra:
Also see https://github.com/oracle/graal/issues/3356

ashu-mehra:
IIUC the graal issue above says the alignment on Catalina would also be 8 bytes but since it did not use `xmm` instructions, it does not suffer from this problem. BigSur is using `xmm` instruction which requires 16-byte alignment and hence the GPF.

ashu-mehra (approved):

ashu-mehra:
I think it makes sense to force the alignment in this exceptional case.

dgrove-oss:
Right.   From the actual C types being declared as elements of the structure in libunwind.h, an 8 byte alignment seems justified.  But there's an external constraint that isn't captured in declaration in unwind.h 😭

dgrove-oss:
ok.  I think Apple screwed this up.  So we have to hack it ourselves.  On the plus side, its not likely to be a general problem with Cprobe.

The libgcc version of this struct is:
```
struct _Unwind_Exception {
 ... elided field declarations ... 
 
/* @@@ The IA-64 ABI says that this structure must be double-word aligned.
Taking that literally does not make much sense generically.  Instead we
provide the maximum alignment required by any type for the machine.  */
} __attribute__((__aligned__));
```

Apple decided to innovate and instead do:
```
struct _Unwind_Exception {
        uint64_t                   exception_class;
        void                       (*exception_cleanup)(int, int);
        uintptr_t                  private_1;        // non-zero means forced unwind
        uintptr_t                  private_2;        // holds sp that phase1 found for phase2 to use
#if !__LP64__
        // The gcc implementation of _Unwind_Exception used attribute mode on the above fields
        // which had the side effect of causing this whole struct to round up to 32 bytes in size.
        // To be more explicit, we add pad fields added for binary compatibility.  
        uint32_t                                reserved[3];
#endif
```

I changed @ashu-mehra test program to the following, just to verify that the missing __attribute was the problem.

```
#include <stdio.h>
#include <stdint.h>
#include <stddef.h>
#include <stdalign.h>

struct _Unwind_Exception {
        uint64_t                   exception_class;
        void                       (*exception_cleanup)(int, int);
        uintptr_t                  private_1;        // non-zero means forced unwind
        uintptr_t                  private_2;        // holds sp that phase1 found for phase2 to use
#if !__LP64__
        // The gcc implementation of _Unwind_Exception used attribute mode on the above fields
        // which had the side effect of causing this whole struct to round up to 32 bytes in size.
        // To be more explicit, we add pad fields added for binary compatibility.  
        uint32_t                                reserved[3];
#endif
} __attribute((__aligned__));

int main(void)
{
    printf("Alignment of UnwindExceptoin using _Alignof = %zu\n", _Alignof(struct _Unwind_Exception));

    return 0;
}
```
Compiling this and running on Big Sur gets the desired output of 16.
```
Daves-MacBook-Pro:tests dgrove$ ./a.out 
Alignment of UnwindExceptoin using _Alignof = 16
```

dgrove-oss:
Fooling around to see if I can get a less hacky way of getting to the platform's max_align than slapping in 16... I think it should be doable; then I'll merge.

dgrove-oss:
For the record, I did submit a bug report to Apple reporting the inconsistent alignment assumptions of libunwind and unwind.h on BigSur.

PR #624, opened 2021-08-19
CompoundType members must have a name except INLINE

theresa-m:

### DIFF
--- a/compiler/src/main/java/org/qbicc/type/CompoundType.java
+++ b/compiler/src/main/java/org/qbicc/type/CompoundType.java
@@ -1,5 +1,7 @@
 package org.qbicc.type;
 
+import io.smallrye.common.constraint.Assert;
+
 import java.util.ArrayList;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


compiler/src/main/java/org/qbicc/type/CompoundType.java
```diff
@@ -258,6 +259,7 @@ public Builder setOverallAlignment(int align) {
         }
 
         public Builder addNextMember(final ValueType type) {
+            assert this.tag.equals(Tag.INLINE);
```
DanHeidinga, line 262: These asserts will only fire if running java with `-ea` (enable assertions).  I think this would be better done either as an `io.smallrye.common.constraint.Assert`.  Applies above as well

theresa-m: Thanks. I've made that change.

DanHeidinga (approved):

PR #625, opened 2021-08-20
increase CI timeout from 20 minutes to 25 minutes

dgrove-oss:


DanHeidinga (approved):
Approving.  Let's hope 25 min is long enough

PR #628, opened 2021-08-24
Serialization of java.lang.Class instances

dgrove-oss:
Initial steps; should be enough for executing static synchronized methods using the nativeObjectMonitor field of the appropriate class object.
- Emit the array of j.l.C instances indexed by typeId
- Implement VM_Helpers.classof_from_typeid
- Add compiler pass to "serialize" j.l.C instances for all initialized types
- Initialize a couple of the instance fields (name, id); leave the rest zero initialized.

### DIFF
--- a/plugins/serialization/src/main/java/org/qbicc/plugin/serialization/BuildtimeHeap.java
+++ b/plugins/serialization/src/main/java/org/qbicc/plugin/serialization/BuildtimeHeap.java
@@ -8,11 +8,11 @@ import java.util.List;
 import java.util.Map;
 import java.util.function.Supplier;
 
+import io.smallrye.common.constraint.Assert;
 import org.qbicc.context.AttachmentKey;
 import org.qbicc.context.CompilationContext;
 import org.qbicc.graph.literal.Literal;
 import org.qbicc.graph.literal.LiteralFactory;
-import org.qbicc.graph.literal.SymbolLiteral;
 import org.qbicc.object.Data;
 import org.qbicc.object.Linkage;
 import org.qbicc.object.Section;
@@ -23,7 +23,9 @@ import org.qbicc.type.CompoundType;
 import org.qbicc.type.TypeSystem;
 import org.qbicc.type.WordType;
 import org.qbicc.type.definition.LoadedTypeDefinition;
+import org.qbicc.type.definition.element.ExecutableElement;
 import org.qbicc.type.definition.element.FieldElement;
+import org.qbicc.type.definition.element.GlobalVariableElement;
 import org.qbicc.type.descriptor.BaseTypeDescriptor;
 
 public class BuildtimeHeap {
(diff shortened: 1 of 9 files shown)


dgrove-oss:
This is ready for review at some point; done polishing it.

plugins/serialization/src/main/java/org/qbicc/plugin/serialization/BuildtimeHeap.java
```diff
@@ -82,12 +93,29 @@ public synchronized Data serializeStringLiteral(String value) {
...
+
     public synchronized Data serializeObject(Object obj) {
         if (objects.containsKey(obj)) {
             return objects.get(obj);
         }
 
         Class<?> cls = obj.getClass();
+        // java.lang.Class instances are their own special thing; can't just reflectively write them.
+        if (obj instanceof Class<?>) {
+            LoadedTypeDefinition ltd = ctxt.getBootstrapClassContext().findDefinedType(cls.getName().replace('.', '/')).load();
```
DanHeidinga, line 115: Will this work for non-bootstrap classes?  I think this will work for now but once the interpreter is up and running, we may need to revisit this as there may be other ClassContexts.

plugins/serialization/src/main/java/org/qbicc/plugin/serialization/BuildtimeHeap.java
```diff
@@ -118,6 +146,38 @@ public synchronized Data serializeObject(Object obj) {
...
+        }
+
+        // Object header
+        memberMap.put(jlcLayout.getMember(layout.getObjectTypeIdField()), lf.literalOf(jlc.getTypeId()));
+
+        // Next, initialize instance fields of the qbicc jlc type that we can/want to set at build time.
+        CompoundType.Member name = jlcType.getMember("name");
+        String externalName = type.getInternalName().replace('/', '.');
+        if (externalName.startsWith("internal_array_")) {
+            externalName = externalName.replaceFirst("internal_array_", "[");
```
DanHeidinga, line 167: nice catch on remapping the internal array names!

plugins/serialization/src/main/java/org/qbicc/plugin/serialization/BuildtimeHeap.java
dgrove-oss: Expect it won't work as is.  We'll need a way to map from the ClassLoader of the Class to the ClassContext we are supposed to use to load the qbicc type.  Assuming this will be possible, but we don't have it yet.

dmlloyd: The interpreter has an API for loading classes, so this would be how it is done once that is ready.

DanHeidinga (approved):

PR #630, opened 2021-08-24
Missing monitor_enter cleanup

theresa-m:
I realized that in an attempt to make this readable I forgot a whole bunch of "frees". I added some extra functions to try to mimic how fall through goto statements would be used to destroy resources in c.

### DIFF
--- a/runtime/main/src/main/java/org/qbicc/runtime/main/VMHelpers.java
+++ b/runtime/main/src/main/java/org/qbicc/runtime/main/VMHelpers.java
@@ -100,23 +100,9 @@ public final class VMHelpers {
         return null; // TODO: Implement this! (or perhaps implement it inline; it should take less code than a call).
     }
 
-    @NoSideEffects
-    private static void omError(c_int nativeErrorCode) throws IllegalMonitorStateException {
-        int errorCode = nativeErrorCode.intValue();
-        if (0 != errorCode) {
-//            putchar('O');
-//            putchar('M');
-//            putchar('E');
-//            putchar('R');
-//            putchar('R');
-//            putchar(':');
-//            printInt(errorCode);
-            throw new IllegalMonitorStateException("error code: " + errorCode);
-        }
-    }
-
     // TODO: mark this with a "NoInline" annotation
-    static void monitor_enter(Object object) throws IllegalMonitorStateException {
+    static void monitor_enter(Object object) throws Throwable {
+        int result;
         if (object == null) {
             /* TODO skip for now. Object should never be null except that
                 classof_from_typeid is not currently implemented. */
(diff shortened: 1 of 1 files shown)


runtime/main/src/main/java/org/qbicc/runtime/main/VMHelpers.java
```diff
@@ -100,23 +100,9 @@ private static boolean isAssignableToLeaf(type_id valueTypeId, type_id toTypeId)
...
-//            putchar('R');
-//            putchar(':');
-//            printInt(errorCode);
-            throw new IllegalMonitorStateException("error code: " + errorCode);
-        }
-    }
-
     // TODO: mark this with a "NoInline" annotation
-    static void monitor_enter(Object object) throws IllegalMonitorStateException {
+    static void monitor_enter(Object object) throws Throwable {
```
DanHeidinga, line 104: Would it be cleaner to use some nested `try / finally` blocks here?  Something roughly like:
```java
   // TODO: mark this with a "NoInline" annotation
    static void monitor_enter(Object object) throws Throwable {
        int result;
        if (object == null) {
            /* TODO skip for now. Object should never be null except that
                classof_from_typeid is not currently implemented. */
            return;
        }
        pthread_mutex_t_ptr nom = ObjectModel.get_nativeObjectMonitor(object);
        if (nom == null) {

            ptr<?> attrVoid = null; 
	    try {
		    ptr<?> attrVoid = malloc(sizeof(pthread_mutexattr_t.class));
		    if (attrVoid.isNull()) {
			throw new OutOfMemoryError();
		    }
		    ptr<pthread_mutexattr_t> attr = (ptr<pthread_mutexattr_t>) castPtr(attrVoid, pthread_mutexattr_t.class);

		    result = pthread_mutexattr_init((pthread_mutexattr_t_ptr) attr).intValue();
		    if (0 != result) {
			    throw new IllegalMonitorStateException("error code: " + result);
		    }
		    try {
			    result = pthread_mutexattr_settype((pthread_mutexattr_t_ptr) attr, PTHREAD_MUTEX_RECURSIVE).intValue();
			    if (0 != result) {
				    throw new IllegalMonitorStateException("error code: " + result);
			    }

			    Stddef.size_t mutexSize = sizeof(pthread_mutex_t.class);
			    ptr<?> mVoid = malloc(word(mutexSize.longValue()));
			    if (mVoid.isNull()) {
				    throw new OutOfMemoryError(/*"Allocation failed"*/);
			    }
			    ptr<pthread_mutex_t> m = (ptr<pthread_mutex_t>) castPtr(mVoid, pthread_mutex_t.class);

			    result = pthread_mutex_init((pthread_mutex_t_ptr) m, (const_pthread_mutexattr_t_ptr) attr).intValue();
			    if (0 != result) {
				    free(m);
				    throw new IllegalMonitorStateException("error code: " + result);
			    }
		    } finally {
			ptr<pthread_mutexattr_t> attr = (ptr<pthread_mutexattr_t>) castPtr(attrVoid, pthread_mutexattr_t.class);
			pthread_mutexattr_destroy((pthread_mutexattr_t_ptr) attr);
		    }
            } finally {
		if (attrVoid != null) {
                   /* attr is no longer needed */
                   monitor_enter_cleanup_destroyMutexAttr(attrVoid, null);
		}
	    }

            nom = (pthread_mutex_t_ptr) m;
            if (!ObjectModel.set_nativeObjectMonitor(object, nom)) {
                /* atomic assignment failed, mutex has already been initialized for object. */
                pthread_mutex_destroy(nom);
                free(mVoid);
                mVoid = null;
                nom = ObjectModel.get_nativeObjectMonitor(object);
            }
        }
        result = pthread_mutex_lock(nom).intValue();
        if (0 != result) {
            throw new IllegalMonitorStateException("error code: " + result);
        }
    }
```

DanHeidinga:
If we keep the separate function approach, my preference is that the extra functions **both** throw the exception and are declared to return a Throwable.

ie:
```
Throwable some_helper(Object value, Throwable t) {
    ....
    throw t;
}
```
so that callers can write:
```
if (!valid) {
    throw some_helper(....);
}
```

This makes it clear to the reader of the code that execution doesn't continue after the call to `some_helper()` and the compiler can generate better code as it knows that it doesn't need to merge the code following the `some_helper()` call back to the mainline path

runtime/main/src/main/java/org/qbicc/runtime/main/VMHelpers.java
theresa-m: I like this idea. Having nested methods could be difficult and confusing to maintain as the code changes.

theresa-m: I updated the method to use try/finally.

PR #639, opened 2021-09-03
Create j.l.Class instance for array classes

ashu-mehra:


compiler/src/main/java/org/qbicc/type/definition/classfile/ClassFileImpl.java
```diff
@@ -324,7 +325,13 @@ public Literal getConstantValue(int idx, TypeParameterContext paramCtxt) {
         }
         int constantType = getConstantType(idx);
         switch (constantType) {
-            case CONSTANT_Class: return setIfNull(literals, idx, literalFactory.literalOfType(getTypeConstant(idx, paramCtxt)));
+            case CONSTANT_Class: {
+                ValueType valueType = getTypeConstant(idx, paramCtxt);
```
dmlloyd, line 329: I think we want to forbid `getTypeConstant` from returning `ReferenceType`.  Instead, in places that require a reference type, it's easier to produce a reference type from an object type than the reverse.

compiler/src/main/java/org/qbicc/type/definition/classfile/ClassFileImpl.java
```diff
@@ -345,6 +352,14 @@ public Literal getConstantValue(int idx, TypeParameterContext paramCtxt) {
         }
     }
 
+    private ValueType getReferencedObjectType(ReferenceType refType) {
+        if (refType.getInterfaceBounds().size() > 0) {
```
dmlloyd, line 356: Just a note that this only works because in this particular case you'll never have a type with both a class bound and an interface bound, and you'll never have a type with more than one interface bound.  At least, I think that is the case. In any event it's better not to have this logic at all.

ashu-mehra:
@dmlloyd this PR's first [commit](https://github.com/qbicc/qbicc/pull/639/commits/9ea2ae228a473b8ed8d3973372f0db5542237a53) attempts to fix the class literal issue that was discussed [here](https://qbicc.zulipchat.com/#narrow/stream/284393-qbicc-development/topic/handling.20class.20literals), but I am not sure if this is what you were suggesting.

runtime/main/src/main/java/org/qbicc/runtime/main/ObjectModel.java
```diff
@@ -28,7 +33,43 @@
...
+    /**
+     * Creates java.lang.Class instance for array class
+     *
+     * @param componentClass leaf element of the array
+     * @param dimensions dimensions of the array
+     * @return instance of java.lang.Class
+     */
+    public static synchronized Class<?> get_or_create_class_for_refarray(Class<?> componentClass, Stdint.uint8_t dimensions) {
+        if (classMap == null) {
+            classMap = new HashMap<>();
```
ashu-mehra, line 47: `HashMap` can be replaced with `ConcurrentHashMap` to avoid making this method `synchronized` but `ConcurrentHashMap` uses `Unsafe` APIs which are not yet supported in qbicc.

plugins/instanceof-checkcast/src/main/java/org/qbicc/plugin/instanceofcheckcast/InstanceOfCheckCastBasicBlockBuilder.java
```diff
@@ -189,15 +189,22 @@ public Value instanceOf(final Value input, final ObjectType expectedType, int ex
...
             ValueType valueType = ((TypeLiteral) typeId).getValue();
             if (valueType instanceof ReferenceType) {
                 ctxt.error(getLocation(), "Class of reference type");
                 return ctxt.getLiteralFactory().zeroInitializerLiteralOfType(methodElement.getType().getReturnType());
+            } else if (valueType instanceof ReferenceArrayObjectType) {
+                ReferenceArrayObjectType referenceArrayType = (ReferenceArrayObjectType) valueType;
+                Value elementTypeId = ctxt.getLiteralFactory().literalOfType(referenceArrayType.getLeafElementType());
+                int dims = referenceArrayType.getDimensionCount();
+                List<Value> args = List.of(elementTypeId, ctxt.getLiteralFactory().literalOf(ctxt.getTypeSystem().getUnsignedInteger8Type(), dims));
+                return notNull(getFirstBuilder().call(getFirstBuilder().staticMethod(methodElement), args));
```
dmlloyd, line 203: A type literal can never carry a type of `ReferenceArrayObjectType`.  This is because the type ID of _any_ reference array class is always going to be the same: the type ID of the special internal reference array implementation class.  That means that the `typeIdOf` node won't actually work correctly for that case.

The logic for actually finding the array type should probably reside completely within the helper method.

runtime/main/src/main/java/org/qbicc/runtime/main/ObjectModel.java
dmlloyd: We definitely do not want to use a map here at all. Each `Class` should contain a field for the array type of that `Class`, which is lazily initialized by this helper method; thus no synchronization is necessary (just a CAS operation after the array `Class` is constructed).

compiler/src/main/java/org/qbicc/type/definition/classfile/ClassFileImpl.java
ashu-mehra: I am realizing it now what you meant when you said it is a bit nontrivial ;-)

plugins/lowering/src/main/java/org/qbicc/plugin/lowering/VMHelpersSetupHook.java
```diff
@@ -6,7 +6,7 @@
 import java.util.function.Consumer;
 
 /**
- * Unconditionally register VMHelper methods that we may not actually refer to until lowering.
+ * Unconditionally register VMHelper or ObjectModel methods that we may not actually refer to until lowering.
  */
 public class VMHelpersSetupHook implements Consumer<CompilationContext> {
```
ashu-mehra, line 11: We should probably rename `VMHelpersSetupHook` to `RuntimeHelpersSetupHook`.

runtime/main/src/main/java/org/qbicc/runtime/main/ObjectModel.java
```diff
@@ -28,7 +28,97 @@
...
+     * @param componentClass
+     * @return instance of java.lang.Class
+     */
+    public static Class<?> get_or_create_array_class(Class<?> componentClass) {
+        Class<?> arrayClass = get_array_class_of(componentClass);
+        if (arrayClass == null) {
+            String className;
+            if (is_reference_array_class(componentClass)) {
+                className = "[" + componentClass.getName();
+                arrayClass = create_class(className, get_reference_array_typeid(), get_dimensions_from_class(componentClass));
```
ashu-mehra, line 75: I have intentionally duplicated this statement in both `if` and `else` block.
Putting it at the end is resulting in some issue with LLVM code generation (LLVM IR looks fine I think) which results in incorrect `className` set for the newly created instance of j.l.Class when the code flows through `if` block.
I will try to create a simplified test case that reproduces this issue.

ashu-mehra:
Finally I got this working after addressing the review comments :relieved: 
It should now be reading for next round of review.

dmlloyd (approved):

dmlloyd:
I'm good with this as it is. Do you want to merge now or is there anything else you wanted to do first?

plugins/instanceof-checkcast/src/main/java/org/qbicc/plugin/instanceofcheckcast/SupersDisplayTables.java
```diff
@@ -338,6 +338,11 @@ public int getNumberOfBytesInInterfaceBitsArray() {
     public int getInterfaceByteIndex(LoadedTypeDefinition cls) {
         Assert.assertTrue(cls.isInterface());
         IdAndRange idRange = typeids.get(cls);
+        try {
+            int rc = idRange.implementedInterfaceByteIndex();
+        } catch (NullPointerException e) {
+            System.out.println("NPE caught for cls: " + cls.getInternalName());
```
dgrove-oss, line 344: Debugging code that should be removed?  Or if we are going to catch a NPE then flag as an error

dgrove-oss (commented):
Other than suppressing the NPE, it looks fine to me

plugins/instanceof-checkcast/src/main/java/org/qbicc/plugin/instanceofcheckcast/SupersDisplayTables.java
ashu-mehra: aah! thanks for pointing this out. I will update the commit to remove it.

ashu-mehra:
> Do you want to merge now or is there anything else you wanted to do first?

Nothing more to do in this PR if there are no comments.

dmlloyd:
OK! Once CI passes, let's merge.

ashu-mehra:
Tests are timing out. I guess news tests is causing CI to run longer than expected.

ashu-mehra:
Restarting the jobs one more time.

ashu-mehra:
Passed this time. Merging it.

PR #644, opened 2021-09-13
Increase CI timeout to 40 minutes

dmlloyd:


DanHeidinga (approved):

DanHeidinga:
Sigh... unfortunately but necessary at this stage

dmlloyd:
Yeah I'm not happy about it.  At some point after the interpreter is integrated we should take a closer look at the LLVM problem.

PR #645, opened 2021-09-14
Primitive types and class literals

ashu-mehra:
This PR is an attempt to give java primitive types a "home".
It builds on that and creates class objects for primitive types,
and fixes few issues to make class literals for primitive types and their arrays work correctly.

### DIFF
--- a/compiler/src/main/java/org/qbicc/type/BooleanType.java
+++ b/compiler/src/main/java/org/qbicc/type/BooleanType.java
@@ -50,4 +50,8 @@ public final class BooleanType extends WordType {
     public StringBuilder toFriendlyString(final StringBuilder b) {
         return b.append("bool");
     }
+
+    public Primitive asPrimitive() {
+        return Primitive.BOOLEAN;
+    }
 }
--- a/compiler/src/main/java/org/qbicc/type/Primitive.java
+++ b/compiler/src/main/java/org/qbicc/type/Primitive.java
@@ -15,8 +15,8 @@ public enum Primitive {
     VOID("void");
 
     private final String name;
-    private int typeId;
-    private ValueType type;
+    volatile private int typeId;
+    volatile private ValueType type;
 
     Primitive(String name) {
         this.name = name;
--- a/compiler/src/main/java/org/qbicc/type/Primitive.java
+++ b/compiler/src/main/java/org/qbicc/type/Primitive.java
@@ -1,0 +1,86 @@
+package org.qbicc.type;
+
+import java.util.function.Consumer;
+
+public enum Primitive {
+    // Predefine the set of primitive types in Java
+    BOOLEAN("boolean"),
+    BYTE("byte"),
+    SHORT("short"),
+    CHAR("char"),
+    INT("int"),
+    FLOAT("float"),
+    LONG("long"),
+    DOUBLE("double"),
+    VOID("void");
+
+    private final String name;
+    private int typeId;
+    private ValueType type;
+
+    Primitive(String name) {
+        this.name = name;
+    }
+
+    public String getName() {
+        return name;
+    }
+
+    public int getTypeId() {
+        return typeId;
+    }
+
+    public void setTypeId(int typeId) {
+        this.typeId = typeId;
+    }
+
+    public ValueType getType() {
+        return this.type;
+    }
+
+    public void setType(ValueType type) {
+        this.type = type;
+    }
+
+    public static void forEach(Consumer<Primitive> function) {
+        for (Primitive type : Primitive.values()) {
+            function.accept(type);
+        }
+    }
+
+    public static Primitive getPrimitiveFor(String typeName) {
+        Primitive type;
+        switch(typeName) {
+            case "byte":
+                type = BYTE;
+                break;
+            case "short":
+                type = SHORT;
+                break;
+            case "int":
+                type = INT;
+                break;
+            case "long":
+                type = LONG;
+                break;
+            case "char":
+                type = CHAR;
+                break;
+            case "float":
+                type = FLOAT;
+                break;
+            case "double":
+                type = DOUBLE;
+                break;
+            case "boolean":
+                type = BOOLEAN;
+                break;
+            case "void":
+                type = VOID;
+                break;
+            default:
+                throw new IllegalStateException("Unexpected primitive type: " + typeName);
+        }
+        return type;
+    }
+}
--- a/integration-tests/src/it-in/snippets/ClassLiteralTests.java
+++ b/integration-tests/src/it-in/snippets/ClassLiteralTests.java
@@ -22,6 +22,7 @@ public class ClassLiteralTests {
         } else {
             putchar('F');
         }
+        putchar('#');
     }
 
     static void testArrayOfPrimitiveType() {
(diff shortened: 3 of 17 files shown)


integration-tests/src/it-in/snippets/ClassLiteralTests.java
```diff
@@ -92,16 +97,12 @@ static void testMultiDimensionalArrayOfConcreteType() {
         } else {
             putchar('F');
         }
+        putchar('#');
     }
     public static void main(String args[]) {
-        /**
-         * Tests for class literals of primitive type are disabled as
```
ashu-mehra, line 98: Just for the record - this comment was incorrect. Before this PR these tests were failing and the real reason for failure was that we are not creating j.l.Class object for primitive types.

compiler/src/main/java/org/qbicc/type/PrimitiveType.java
```diff
@@ -0,0 +1,88 @@
+package org.qbicc.type;
+
+import java.util.function.Consumer;
+
+public class PrimitiveType {
```
dmlloyd, line 5: The idea here I think is really good and cleans up a lot of ugly stuff.  A couple minor notes though:

- The fact that this class ends in `...Type` and resides in the `.type` package makes it appear as if this is part of the type system, but it is not.  Maybe just call it `Primitive`?
- Would this be better done as an `enum`?

compiler/src/main/java/org/qbicc/type/BooleanType.java
```diff
@@ -50,4 +50,8 @@ public StringBuilder toString(final StringBuilder b) {
     public StringBuilder toFriendlyString(final StringBuilder b) {
         return b.append("bool");
     }
+
+    public PrimitiveType asJavaType() {
```
dmlloyd, line 54: `asJavaType` is a bit at odds with the class name `PrimitiveType` IMO; they should be more consistent.

plugins/instanceof-checkcast/src/main/java/org/qbicc/plugin/instanceofcheckcast/SupersDisplayBuilder.java
```diff
@@ -33,17 +34,9 @@ public void accept(CompilationContext ctxt) {
...
-        // primitives
-        // [1] boolean.class
-        // [2] byte.class
-        // [3] short.class
-        // [4] char.class
-        // [5] int.class
-        // [6] float.class
-        // [7] long.class
-        // [8] double.class
-        // [9] void.class
```
dmlloyd, line 45: The class for primitives does not match this ordering, so it is very important to ensure that we don't have these values hard-coded anywhere. I think every such place would be marked with `TODO` or `XXX` though...

compiler/src/main/java/org/qbicc/type/PrimitiveType.java
ashu-mehra: Renamed `PrimitiveType` to `Primitive` and changed it to enum.

compiler/src/main/java/org/qbicc/type/BooleanType.java
ashu-mehra: Renamed `asJavaType` to `asPrimitive`.

plugins/instanceof-checkcast/src/main/java/org/qbicc/plugin/instanceofcheckcast/SupersDisplayBuilder.java
ashu-mehra: Sorry, I didn't get it. Can you please point out where is the mismatch in the ordering?

DanHeidinga: I think the the primitives do match but it's much more implicit now.  The primitives are added with `assignTypeIdToPrimitives()`, which does `PrimitiveType.forEach()` which in turn iterates the `PrimitiveType.TYPES` array using `for (PrimitiveType type : PrimitiveType.TYPES) {`.

As long as the order of elements in the `PrimitiveType.TYPES` doesn't change, this should match.  But it's super subtle.

Best to check everywhere using this and ensure it doesn't rely on the ordering.  (ie: that sub-int types typeid < int.typeid}

dmlloyd: Sorry, I thought you had `boolean` last, but maybe I just read it wrong without my glasses on.

ashu-mehra: still not clear - why are we worried about order of primitives anyway? is there relationship between typeid of primitives?

Are there any other places where we hard-code the typeid for primitives? Is that what we are are worried about - that the hard-coded values may not match with the order in `PrimitiveType`?

ashu-mehra:
Updated commits as per the suggestions.

plugins/instanceof-checkcast/src/main/java/org/qbicc/plugin/instanceofcheckcast/SupersDisplayBuilder.java
dmlloyd: Yes, exactly; there may be places where type IDs for primitives were hard-coded so we should ensure that those all are replaced with references to `Primitive`.

ashu-mehra: ok, only place I could find was in `LLVMModuleNodeVisitor` where we lower the `TypeLiteral`. Do you know any other place that needs to be updated.

DanHeidinga: I don't see any other places - the ones that typically need to be changed are the ObjectModel helpers but so far they only care about where Object starts, where the interfaces start, and the arrays.

ashu-mehra: @DanHeidinga thanks for confirming

compiler/src/main/java/org/qbicc/type/Primitive.java
```diff
@@ -0,0 +1,86 @@
...
+    CHAR("char"),
+    INT("int"),
+    FLOAT("float"),
+    LONG("long"),
+    DOUBLE("double"),
+    VOID("void");
+
+    private final String name;
+    private int typeId;
+    private ValueType type;
```
dmlloyd, line 19: Since these are set late and are shared variables, I think they should be `volatile`.

dmlloyd (approved):
I only have one more comment; once that's addressed I think this can be merged.

compiler/src/main/java/org/qbicc/type/Primitive.java
ashu-mehra: Done

ashu-mehra:
This is ready now.

DanHeidinga (approved):

PR #646, opened 2021-09-14
Remove deserialization jar from example command

theresa-m:


ashu-mehra:
While you are it, can you please dependency in top-lelve pom.xml as well:
https://github.com/qbicc/qbicc/blob/05ddc11b122a91d3896c0c847a8db97d929ad5ee/pom.xml#L83

dgrove-oss (approved):
Thanks.  I missed fixing this when I got rid of the runtime-deserialization plugin

PR #648, opened 2021-09-16
Install maven manually on macos

theresa-m:

### DIFF
--- a/docs/README.adoc
+++ b/docs/README.adoc
@@ -61,6 +61,8 @@ Please ensure the follow are installed and available on your path:
 
 === MacOS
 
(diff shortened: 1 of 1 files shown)


docs/README.adoc
```diff
@@ -61,6 +61,8 @@ Please ensure the follow are installed and available on your path:
 
 === MacOS
 
+To prevent Java versioning errors it is best to install Maven manually rather than with homebrew (see https://maven.apache.org/install.html).
```
DanHeidinga, line 64: Can you add some additional detail to this about what thee "versioning errors" looks like?  Even a sentence that says "This typically manifests as Maven using the wrong version of Java when building qbicc."

theresa-m: Thanks. Updated.

DanHeidinga (approved):

PR #650, opened 2021-09-21
Miscellaneous interpreter and compiler fixes

dmlloyd:


dmlloyd:
Easiest reviewed commit by commit.

plugins/intrinsics/src/main/java/org/qbicc/plugin/intrinsics/core/CoreIntrinsics.java
```diff
@@ -263,6 +264,14 @@ public static void registerJavaLangSystemIntrinsics(CompilationContext ctxt) {
...
 
         intrinsics.registerIntrinsic(systemDesc, "identityHashCode", objectToIntDesc, identityHashCode);
+
+        // registerNatives
+
+        Literal voidLiteral = ctxt.getLiteralFactory().zeroInitializerLiteralOfType(ctxt.getTypeSystem().getVoidType());
+
+        StaticIntrinsic registerNatives = (builder, target, arguments) -> voidLiteral;
+
+        intrinsics.registerIntrinsic(systemDesc, "registerNatives", emptyToVoid, registerNatives);
```
DanHeidinga, line 274: It's trivial code so probably doesn't matter, but there's multiple copies of the empty `registerNatives` intrinsic.  If we need to add another, it might be better to commonize so there's only 1 copy of it

plugins/intrinsics/src/main/java/org/qbicc/plugin/intrinsics/core/CoreIntrinsics.java
```diff
@@ -1588,5 +1588,13 @@ private static void registerJdkInternalMiscUnsafeIntrinsics(final CompilationCon
         };
 
         intrinsics.registerIntrinsic(unsafeDesc, "isBigEndian0", emptyToBool, isBigEndian0);
+
+        InstanceIntrinsic unalignedAccess0 = (builder, instance, target, arguments) -> {
+            ClassContext c = builder.getCurrentElement().getEnclosingType().getContext();
+            LiteralFactory lf = c.getLiteralFactory();
+            return lf.literalOf(false);
```
DanHeidinga, line 1595: Isn't this platform dependent?  aarch64/x86/s390/ppc would all say true.  Others false.  

False is probably safe but may be pessimistic for some use cases

DanHeidinga (approved):
Overall, this looks fine.  A couple of questions but nothing that should block merging

plugins/intrinsics/src/main/java/org/qbicc/plugin/intrinsics/core/CoreIntrinsics.java
dmlloyd: I agree

plugins/intrinsics/src/main/java/org/qbicc/plugin/intrinsics/core/CoreIntrinsics.java
dmlloyd: #651

PR #652, opened 2021-09-21
Process generic types for functional interface

theresa-m:
Return function interface type with correct generics. For example: function_ptr<UnaryOperator<void_ptr>>

### DIFF
--- a/plugins/native/src/main/java/org/qbicc/plugin/native_/FunctionTypeResolver.java
+++ b/plugins/native/src/main/java/org/qbicc/plugin/native_/FunctionTypeResolver.java
@@ -92,7 +92,8 @@ public class FunctionTypeResolver implements DescriptorTypeResolver.Delegating {
                                     functionType = ts.getFunctionType(ts.getVoidType());
                                 } else {
                                     if (definedType.isInterface()) {
-                                        return NativeInfo.get(ctxt).getTypeOfFunctionalInterface(definedType);
+                                        List<TypeArgument> functionalInterfaceArguments = ((ClassTypeSignature)bound.getBound()).getTypeArguments();
+                                        return NativeInfo.get(ctxt).getTypeOfFunctionalInterface(definedType, functionalInterfaceArguments);
                                     } else {
                                         ctxt.error("Function interface type \"%s\" is not an interface", name);
                                         functionType = ts.getFunctionType(ts.getVoidType());
--- a/plugins/native/src/main/java/org/qbicc/plugin/native_/NativeInfo.java
+++ b/plugins/native/src/main/java/org/qbicc/plugin/native_/NativeInfo.java
@@ -5,6 +5,7 @@ import java.nio.ByteBuffer;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.Comparator;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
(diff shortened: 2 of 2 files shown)


theresa-m:
Combining with https://github.com/qbicc/qbicc/pull/654 to resolve dependency

plugins/native/src/main/java/org/qbicc/plugin/native_/FunctionTypeResolver.java
```diff
@@ -92,7 +92,8 @@ public ValueType resolveTypeFromDescriptor(final TypeDescriptor descriptor, Type
                                     functionType = ts.getFunctionType(ts.getVoidType());
                                 } else {
                                     if (definedType.isInterface()) {
-                                        return NativeInfo.get(ctxt).getTypeOfFunctionalInterface(definedType);
+                                        List<TypeArgument> functionalInterfaceArguments = ((ClassTypeSignature)bound.getBound()).getTypeArguments();
+                                        return NativeInfo.get(ctxt).getTypeOfFunctionalInterface(definedType, functionalInterfaceArguments);
```
dmlloyd, line 96: It *should* be the case that you can use the `TypeParameterContext paramCtxt` argument to resolve the actual types of any type variables found in the given signature.  Without that, I'm not sure you're going to be able to correctly resolve them.

dmlloyd:
Sorry, I started reviewing this before it was closed. I did have one comment so far; I'll review the follow up in a bit.

theresa-m:
> Sorry, I started reviewing this before it was closed. I did have one comment so far; I'll review the follow up in a bit.

Thanks! I'll apply it on the new pr.

plugins/native/src/main/java/org/qbicc/plugin/native_/FunctionTypeResolver.java
dmlloyd: The context exists for cases like this:

```java
class Foo<T> {
    function_ptr<Consumer<T>> tConsumerFn;
}
```
Or I guess maybe like this:
```java
abstract class Foo<T> {
    abstract function_ptr<Consumer<T>> getTheFunction();
}

class Bar extends Foo<void_ptr> {
    function_ptr<Consumer<void_ptr>> getTheFunction() { ... }
}
```

I am not quite sure what five-days-ago-me had in mind but upon reflection I guess it might be a good idea to simply disallow this case and any other case where there is even a slight type ambiguity.  This would be consistent with the restrictions already in place in this class which prevent non-invariant arguments from being used.

theresa-m: The problem with ignoring the non-invariant case (if I'm understanding this right) is at least for UnaryOperator the `apply` method that the function type is derived from is inherited from java.util.function.Function so there needs to be some way to understand what the generic types for Function are supposed to be.

dmlloyd: I think you'd still have invariant type arguments in this case so it'd be okay.

TBH though at this point I don't really know what it is I was arguing for so if this solution works then I'm fine with it. There are some holes here and there in our handling of generic types generally but I don't expect them there be analyzed, discovered, and resolved here.

theresa-m: Okay sounds good, I should have looked at it sooner while it was still fresh.

theresa-m: I'm not sure I understand how invariance is working in qbicc. Shouldn't java.util.function.UnaryOperator be covariant then as a sub interface of java.util.function.Function?

dmlloyd: It is a subtype, yes, you are right that the type itself is covariant, but I meant that its generic type arguments would have to be invariant (for the purposes of understanding function types).

Sorry, I really could have been more clear on that. The code base doesn't refer to variance by name much (if at all) outside of handling of generics so my brain just didn't go there. :)

PR #654, opened 2021-09-21
java.lang.Thread.start() natives

theresa-m:


plugins/intrinsics/src/main/java/org/qbicc/plugin/intrinsics/core/CoreIntrinsics.java
```diff
@@ -368,12 +370,106 @@ public static void registerJavaLangThreadIntrinsics(CompilationContext ctxt) {
...
+
+        /* VMHelpers.java - threadWrapper: helper method for java.lang.Thread.start0 */
+        MethodDescriptor threadWrapperNativeDesc = MethodDescriptor.synthesize(classContext, voidPtrDesc, List.of(voidPtrDesc));
+        StaticIntrinsic threadWrapperNative = (builder, target, arguments) -> {
+            Value threadVoidPtr = arguments.get(0);
+
+            DefinedTypeDefinition jlt = classContext.findDefinedType("java/lang/Thread");
+            LoadedTypeDefinition jltVal = jlt.load();
+            ValueType jltType = jltVal.getType().getReference();
+            Value threadObject = builder.bitCast(threadVoidPtr, (WordType)jltType);
```
DanHeidinga, line 406: This makes sense as a starting point but will be unsafe once we have a GC.  We can't pass object pointers through native functions without having either pinned them or used some level of indirection (ie: a jni handle) so the GC can update the real pointer.

It looks like you've already started down this path by adding `saveNativeThread()`.  In J9, we kept a circular linked list of J9VMThread pointers in the VM structure that the GC could use to find all threads to ensure they were coalesced when updating the heap.

We'll need a similar list and the list elements can do double duty - act as a list of roots for the GC and provide the indirection here so we don't lose the thread object

theresa-m: That is the plan for `saveNativeThread`, to add that information to a global list. That will be a future pull request.

DanHeidinga: And to not pass the "naked" object pointer into this function?  Then I'm OK with the approach

theresa-m: Ah yes and pass in the list element. Sorry I think I was getting confused about the meaning of "native functions" for this. You mean don't pass direct object pointers to things like c libraries & exported functions since they will not be covered by the JNI level of indirection?

plugins/intrinsics/src/main/java/org/qbicc/plugin/intrinsics/core/CoreIntrinsics.java
```diff
@@ -368,12 +370,106 @@ public static void registerJavaLangThreadIntrinsics(CompilationContext ctxt) {
...
+
+        /* public final native boolean isAlive(); */
+        InstanceIntrinsic isAlive = (builder, instance, target, arguments) -> {
+            ValueHandle threadStatusHandle = builder.instanceFieldOf(builder.referenceHandle(instance), jltDesc, "threadStatus", BaseTypeDescriptor.I);
+            Value threadStatus = builder.load(threadStatusHandle, MemoryAtomicityMode.NONE);
+            Value aliveState = ctxt.getLiteralFactory().literalOf(threadAlive);
+            Value isThreadAlive = builder.and(threadStatus, aliveState);
+            return builder.isEq(isThreadAlive, aliveState);
+        };
+        intrinsics.registerIntrinsic(jltDesc, "isAlive", booleanDesc, isAlive);
```
dmlloyd, line 469: Again this is OK for now, but the basic problem with using intrinsics for methods that are public like this are that the intrinsic doesn't populate the method's body. This means that if this method were called reflectively we'd have a problem.

It doesn't need to change for now but expect that we'll have to go back and change this to be an actual method at some point.

plugins/intrinsics/src/main/java/org/qbicc/plugin/intrinsics/core/CoreIntrinsics.java
```diff
@@ -549,6 +645,11 @@ public static void registerJavaLangObjectIntrinsics(CompilationContext ctxt) {
         };
 
         intrinsics.registerIntrinsic(objDesc, "clone", MethodDescriptor.synthesize(classContext, objDesc, List.of()), clone);
+
+        // stub - public final native void wait(long timeoutMillis)
+        MethodDescriptor waitDesc = MethodDescriptor.synthesize(classContext, BaseTypeDescriptor.V, List.of(BaseTypeDescriptor.J));
+        InstanceIntrinsic wait = (builder, instance, target, arguments) -> ctxt.getLiteralFactory().zeroInitializerLiteralOfType(target.getType().getReturnType());
+        intrinsics.registerIntrinsic(objDesc, "wait", waitDesc, wait);
```
dmlloyd, line 652: This is OK temporarily (see above) but this method should be implemented directly in `java.lang.Object` in `qbicc-class-library` once we have monitors sorted out.

plugins/native/src/main/java/org/qbicc/plugin/native_/NativeInfo.java
```diff
@@ -303,50 +309,111 @@ private boolean isArgConst(DefinedTypeDefinition definedType) {
         return false;
     }
 
-    public ValueType getTypeOfFunctionalInterface(final DefinedTypeDefinition definedType) {
-        MethodElement method = getFunctionalInterfaceMethod(definedType);
-        if (method == null) {
+    public class FunctionalInterfaceData {
```
dmlloyd, line 312: Should be `static` (and `final`, as a stylistic consideration) and doesn't need to be `public` (it's inside of a non-`public` class).

runtime/api/src/main/java/org/qbicc/runtime/CNative.java
```diff
@@ -937,6 +938,7 @@ public void set(int arrayIdx, T newVal) {
         public native F getInvokable();
     }
 
+    public static final class void_ptr_unaryoperator_function_ptr extends ptr<function<UnaryOperator<void_ptr>>> {}
```
dmlloyd, line 941: Ohh, yeah I didn't think about how reifying the pointer types would affect function pointers.  I will probably have to think about that some more.

plugins/intrinsics/src/main/java/org/qbicc/plugin/intrinsics/core/CoreIntrinsics.java
DanHeidinga: > You mean don't pass direct object pointers to c libraries (pthread_create) since they will not be covered by the JNI level of indirection?

Exactly.  The GC loses track of the direct object pointers once they go off into c libraries.  So we need something - the list element or a JNI handle - to allow the GC to track and update the object pointer regardless of where it may end up in native (c) code.

theresa-m: Got it! Yes I will update the threadWrapper parameter to be the list element instead when that exists.

compiler/src/main/java/org/qbicc/graph/literal/FunctionParameterLiteral.java
```diff
@@ -0,0 +1,46 @@
+package org.qbicc.graph.literal;
+
+import org.qbicc.graph.ValueVisitor;
+import org.qbicc.type.PointerType;
+import org.qbicc.type.ValueType;
+
+public class FunctionParameterLiteral extends Literal {
+    private final String name;
+    private final PointerType type;
```
dmlloyd, line 9: My only real complaint of this iteration is this type, which represents a pointer literal derived from a symbol, which is essentially exactly what `SymbolLiteral` is.  A secondary quibble is that this doesn't really represent a function parameter (assuming I understand the expected usage correctly), so it would have been better named `FunctionPointerLiteral` in which case its type should have been covariantly `PointerType`.

But either way I think it might be better to go directly to the more canonical `AddressOf(FunctionDeclarationHandle)` for cases where you need a `Value` that refers to a native function rather than introducing another variant of `SymbolLiteral`.

dmlloyd (commented):
This looks really good, but the function literal type is the one thing that should be fixed before this goes in.

plugins/intrinsics/src/main/java/org/qbicc/plugin/intrinsics/core/CoreIntrinsics.java
```diff
@@ -368,12 +370,106 @@ public static void registerJavaLangThreadIntrinsics(CompilationContext ctxt) {
...
+
+            /* call "run" method of thread object */
+            VirtualMethodElementHandle runHandle = (VirtualMethodElementHandle)builder.virtualMethodOf(threadObject, jltDesc, "run", voidDesc);
+            builder.call(runHandle, List.of());
+
+            /* set java.lang.Thread.threadStatus to terminated */
+            ValueHandle threadStatusHandle = builder.instanceFieldOf(threadObjectHandle, jltDesc, "threadStatus", BaseTypeDescriptor.I);
+            builder.store(threadStatusHandle, ctxt.getLiteralFactory().literalOf(threadTerminated), MemoryAtomicityMode.NONE);
+
+            return ctxt.getLiteralFactory().zeroInitializerLiteralOfType(target.getType().getReturnType()); /* return null */
```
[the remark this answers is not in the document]
theresa-m: Agreed. Hoping that will help with readability as well as this gets kindof complicated following all the back and forth between VMHelpers and CoreIntrinsics. Is there an issue open for that support yet or is it still a ways off?

plugins/intrinsics/src/main/java/org/qbicc/plugin/intrinsics/core/CoreIntrinsics.java
theresa-m: As well as its annoying to debug 😆 as a future pr it would be helpful (to me at least) just to have even a Thread$_native method that just calls this intrinsic

plugins/intrinsics/src/main/java/org/qbicc/plugin/intrinsics/core/CoreIntrinsics.java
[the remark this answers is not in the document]
dmlloyd: That functionality exists but there are a few problems with it. My current PR includes changes to do the native method binding "correctly". Once that's merged then it should be possible to make this change.  There are a couple of trivial examples already in the `qbicc-class-library` project:

```java
./java.base/src/main/java/java/lang/System$_native.java
./java.base/src/main/java/sun/nio/ch/IOUtil$_native.java
```

runtime/api/src/main/java/org/qbicc/runtime/CNative.java
theresa-m: Yea there may be a lot of these eventually. Wasn't this format intended to be temporary?

dmlloyd: Yeah, I'd really like to go back to inferring them since it's a much nicer experience.

compiler/src/main/java/org/qbicc/graph/literal/FunctionParameterLiteral.java
theresa-m: Oh yes I can basically replace this whole thing with `SymbolLiteral` I see.

I do think a `Literal` solution makes more sense since at the LLVM level the function pointer must be passed in similarly to a constant. Since we want something like:
```
call void @JLT_start0(i8* (i8*)* @threadWrapper)
```
When I have tried treating the function pointer as a`Value` it is usually lowered treating the function pointer as though it can be assigned to a variable like this which doesn't compile:
```
%1 = i8* (i8*)* @threadWrapper
call void @JLT_start0(%1)
```

dmlloyd: > When I have tried treating the function pointer as a`Value` it is usually lowered treating the function pointer as though it can be assigned to a variable like this which doesn't compile:
> 
> ```
> %1 = i8* (i8*)* @threadWrapper
> call void @JLT_start0(%1)
> ```

Hmm that's definitely not what it is supposed to do.  I'll have a look in the LLVM backend to see what is going on there.

theresa-m: This was my most recent hacky test to produce that result. See if thats what you had in mind.
https://github.com/theresa-m/qbicc/commit/f8aa10d8ee3b3a1660c6303f9513997cf3dfc593

theresa-m: Thanks for your PR this morning. I have removed the FunctionParameterLiteral class and am passing threadWrapper as a pointer to a FunctionDeclaration now.

plugins/intrinsics/src/main/java/org/qbicc/plugin/intrinsics/core/CoreIntrinsics.java
```diff
@@ -369,12 +370,105 @@ public static void registerJavaLangThreadIntrinsics(CompilationContext ctxt) {
         Intrinsics intrinsics = Intrinsics.get(ctxt);
         ClassContext classContext = ctxt.getBootstrapClassContext();
 
+        final int threadAlive = 0x0001;
+        final int threadTerminated = 0x0002;
+        final int threadRunnable = 0x0004;
```
dmlloyd, line 375: These could be `static`.

PR #657, opened 2021-09-23
Make class initialization failure a warning... for now

dmlloyd:
Allowing compilation to complete even with failing class init will allow us to move more quickly with getting something working.  The most expedient way to make this work is to disable run time init when build time init is enabled, but this is not strictly correct.  If a class fails to initialize during build, then any class init check should be replaced with a throw of an exception-in-initializer error (which will result in pruning much of the program graph).

The next step would be to determine how the build and run time init modes could mix - if they can at all.


dgrove-oss (approved):
Sounds like a reasonable tactical move to me.

PR #659, opened 2021-09-24
Convert BuildTimeHeap to using VmObject

dgrove-oss:
Support serialization of VmObject instances to the initial heap.
Currently being used to serialize String literals and java.lang.Class
instances.  Should support arbitrary VmObject graphs as soon as
we manage to get them all the way to the backend.

### DIFF
--- a/plugins/serialization/src/main/java/org/qbicc/plugin/serialization/BuildtimeHeap.java
+++ b/plugins/serialization/src/main/java/org/qbicc/plugin/serialization/BuildtimeHeap.java
@@ -1,6 +1,5 @@
 package org.qbicc.plugin.serialization;
 
-import java.lang.reflect.Field;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.IdentityHashMap;
@@ -11,16 +10,26 @@ import java.util.function.Supplier;
 import io.smallrye.common.constraint.Assert;
 import org.qbicc.context.AttachmentKey;
 import org.qbicc.context.CompilationContext;
+import org.qbicc.graph.MemoryAtomicityMode;
 import org.qbicc.graph.literal.Literal;
 import org.qbicc.graph.literal.LiteralFactory;
+import org.qbicc.interpreter.Memory;
+import org.qbicc.interpreter.VmArray;
+import org.qbicc.interpreter.VmObject;
 import org.qbicc.object.Data;
 import org.qbicc.object.Linkage;
 import org.qbicc.object.Section;
 import org.qbicc.plugin.coreclasses.CoreClasses;
 import org.qbicc.plugin.layout.Layout;
 import org.qbicc.type.ArrayType;
+import org.qbicc.type.ClassObjectType;
 import org.qbicc.type.CompoundType;
+import org.qbicc.type.FloatType;
+import org.qbicc.type.IntegerType;
+import org.qbicc.type.PhysicalObjectType;
 import org.qbicc.type.Primitive;
+import org.qbicc.type.PrimitiveArrayObjectType;
+import org.qbicc.type.ReferenceArrayObjectType;
 import org.qbicc.type.TypeSystem;
 import org.qbicc.type.ValueType;
 import org.qbicc.type.WordType;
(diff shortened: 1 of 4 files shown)


plugins/serialization/src/main/java/org/qbicc/plugin/serialization/BuildtimeHeap.java
```diff
@@ -84,18 +102,25 @@ public GlobalVariableElement getAndRegisterGlobalClassArray(ExecutableElement or
...
+        if (vmObjects.containsKey(value)) {
+            return vmObjects.get(value);
         }
-        LoadedTypeDefinition jls = ctxt.getBootstrapClassContext().findDefinedType("java/lang/String").load();
-        Data sl = serializeObject(jls, value);
-        stringLiterals.put(value, sl);
+        PhysicalObjectType ot = value.getObjectType();
+        Data sl;
+        if (ot instanceof ClassObjectType) {
+            // TODO: Right here we need to detect if value is a java.lang.Class instance and handle it specially by calling serializeClassObject!
```
dgrove-oss, line 112: Are we reifying java.lang.Class instances into VmObjects in the interpreter?  Or do we do something else if the build time heap tries to instantiate a Class instance?

dgrove-oss: Cool.  I'll make that adjustment this morning.

dgrove-oss: Added the logic for VmClass.  Ready for review

plugins/serialization/src/main/java/org/qbicc/plugin/serialization/BuildtimeHeap.java
```diff
@@ -84,18 +104,37 @@ public GlobalVariableElement getAndRegisterGlobalClassArray(ExecutableElement or
...
         }
-        LoadedTypeDefinition jls = ctxt.getBootstrapClassContext().findDefinedType("java/lang/String").load();
-        Data sl = serializeObject(jls, value);
-        stringLiterals.put(value, sl);
+        PhysicalObjectType ot = value.getObjectType();
+        Data sl;
+        if (ot instanceof ClassObjectType) {
+            if (value instanceof VmClass) {
+                if (value instanceof VmArrayClass) {
+                    if (((VmArrayClass)value).getInstanceObjectType().getElementType() instanceof ReferenceArrayObjectType) {
```
DanHeidinga, line 116: Trying to think about which code would trigger this case - is it something like:
```
Class MyClass {
    Class<?> arrayClass = MyClass[].class;
}
```

The restriction seems reasonable to start with but trying to get the implications clear in my head

dgrove-oss: Yeah.  It's a staging thing.  We do support getting these class objects at runtime thanks to #639, but I didn't want to think about how to connect the build time creation of one of them to the runtime data structure yet.

DanHeidinga (approved):
I'm good with it.  @dmlloyd any other concerns?

plugins/serialization/src/main/java/org/qbicc/plugin/serialization/BuildtimeHeap.java
```diff
@@ -84,18 +104,37 @@ public GlobalVariableElement getAndRegisterGlobalClassArray(ExecutableElement or
...
+            sl = serializeRefArray((ReferenceArrayObjectType) ot, (VmArray)value);
+        } else {
+            sl = serializePrimArray((PrimitiveArrayObjectType) ot, (VmArray)value);
+        }
+        vmObjects.put(value, sl);
         return sl;
     }
 
-    public Data serializeClassObject(Primitive primitive) {
+    public synchronized Data serializeClassObject(Primitive primitive) {
```
dmlloyd, line 137: It looks like this method is still not being called; was that an oversight or is there still more to do up above? Or is it being called from someplace I can't see?

dgrove-oss: It's being called from ClassObjectSerializer (Analyze post hook).  Was already being called, so it doesn't show up in this PR.

PR #660, opened 2021-09-27
More improvements to interpreter and early init

dmlloyd:
This includes first part of implementing `System.nanoTime()` and `System.currentTimeMillis()`.  This PR is interdependent on qbicc/qbicc-class-library#20.


compiler/src/main/java/org/qbicc/graph/OffsetOfField.java
```diff
@@ -0,0 +1,51 @@
+package org.qbicc.graph;
+
+import java.util.Objects;
+
+import org.qbicc.type.SignedIntegerType;
+import org.qbicc.type.definition.element.ExecutableElement;
+import org.qbicc.type.definition.element.FieldElement;
+
+/**
+ * The class object for a given type ID value.
```
dgrove-oss, line 10: cut & paste comment mistake?

plugins/intrinsics/src/main/java/org/qbicc/plugin/intrinsics/core/CoreIntrinsics.java
```diff
@@ -1596,5 +1621,128 @@ private static void registerJdkInternalMiscUnsafeIntrinsics(final CompilationCon
...
+        StaticIntrinsic initializeFromArchive = (builder, target, arguments) -> voidLiteral;
+
+        intrinsics.registerIntrinsic(vmDesc, "initializeFromArchive", classToVoid, initializeFromArchive);
+
+        StaticIntrinsic initialize = (builder, target, arguments) -> voidLiteral;
+
+        intrinsics.registerIntrinsic(vmDesc, "initialize", emptyToVoid, initialize);
+    }
+
+    private static void registerJavaNetInet4AddressIntrinsics(final CompilationContext ctxt) {
```
dgrove-oss, line 1719: Defining the init methods as no-ops?  Is this all we need in the long term, or is it a stub to be filled in later?

dgrove-oss (approved):
two minor comments, but otherwise looks plausible

compiler/src/main/java/org/qbicc/graph/OffsetOfField.java
dmlloyd: Oops, yep.

plugins/intrinsics/src/main/java/org/qbicc/plugin/intrinsics/core/CoreIntrinsics.java
dmlloyd: This is all we need. The init methods of these classes are used by JNI to load up field offsets which we won't be needing since our "native" code can access the fields directly.

dmlloyd:
Okay this should fix the remaining problems 😅 

Additional commits to implement GetAndAdd and be smarter about resolving unsafe field offsets.  Right now it snoops the program graph to determine if field offsets are valid, but in the future we should devise a better way to know the local value of a given program element if it is statically knowable.

plugins/intrinsics/src/main/java/org/qbicc/plugin/intrinsics/core/CoreIntrinsics.java
```diff
@@ -1596,5 +1626,151 @@ private static void registerJdkInternalMiscUnsafeIntrinsics(final CompilationCon
...
+                fieldName = ((StringLiteral) string).getValue();
+            } else if (string instanceof ObjectLiteral) {
+                VmObject vmObject = ((ObjectLiteral) string).getValue();
+                if (! (vmObject instanceof VmString)) {
+                    ctxt.error(builder.getLocation(), "objectFieldOffset string argument must be a literal string");
+                    return lf.literalOf(0L);
+                }
+                fieldName = ((VmString) vmObject).getContent();
+            } else {
+                ctxt.error(builder.getLocation(), "objectFieldOffset string argument must be a literal string");
```
ashu-mehra, line 1665: @dmlloyd Is it not possible for the field name to not be string literal?

dmlloyd: No, that will not be supported for Unsafe field accesses. It must be a string literal or traceable back to a string literal (for example a static final field whose value is definitely a string literal).  As a possible exception, we could allow the interpreter to compute a `VmString` as long as it is done at compile time.

ashu-mehra: In that case is it correct to say it is more like a limitation imposed by qbicc on the usage of this API,

dmlloyd: Yes, but it's not a good idea to use Unsafe generally in most cases, and it's really not a good idea to plug in arbitrary field offsets when doing so.  Most users are going to use `VarHandle` instead.  So, it is a limitation, but not a significant one.

PR #663, opened 2021-09-29
extend reachability analysis to model build time initialization

dgrove-oss:
Three main changes, all conditional on builtTimeInit being true:
1. Extend the interpreter to log the instantiated types with RTAInfo (buildTimeInit == true implied)
2. Process the logged types as reachable at the beginning of each RTA analysis phase
3. Do not enqueue the `<clinit>` method as reachable

I want to take a closer look in the morning, but this seems to be doing something reasonable at first glance at the -rta-debug logs.

### DIFF
--- a/interpreter/pom.xml
+++ b/interpreter/pom.xml
@@ -32,6 +32,11 @@
         </dependency>
 
         <dependency>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>qbicc-plugin-reachability</artifactId>
+        </dependency>
+
+        <dependency>
             <groupId>io.smallrye.common</groupId>
             <artifactId>smallrye-common-constraint</artifactId>
         </dependency>
--- a/interpreter/src/main/java/org/qbicc/interpreter/impl/Frame.java
+++ b/interpreter/src/main/java/org/qbicc/interpreter/impl/Frame.java
@@ -130,6 +130,7 @@ import org.qbicc.interpreter.VmObject;
 import org.qbicc.interpreter.VmThrowable;
 import org.qbicc.plugin.coreclasses.CoreClasses;
 import org.qbicc.plugin.layout.Layout;
+import org.qbicc.plugin.reachability.RTAInfo;
 import org.qbicc.type.ArrayObjectType;
 import org.qbicc.type.ArrayType;
 import org.qbicc.type.BooleanType;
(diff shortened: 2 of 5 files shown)


dgrove-oss:
On the HelloWorld program, it looks like this change makes a nice improvement.  

Executable size and RTA statistics after the Analyze phase for three configurations.

Main branch, no build time disabled (default). 35 MB exec
```
[D] (org.qbicc.plugin.reachability.rta) ANALYZE: RTA Reachability Statistics
[D] (org.qbicc.plugin.reachability.rta) ANALYZE:   Reachable interfaces:       173
[D] (org.qbicc.plugin.reachability.rta) ANALYZE:   Reachable classes:          1581
[D] (org.qbicc.plugin.reachability.rta) ANALYZE:   Instantiated classes:       1292
[D] (org.qbicc.plugin.reachability.rta) ANALYZE:   Initialized types:          1577
[D] (org.qbicc.plugin.reachability.rta) ANALYZE:   Invokable instance methods: 7142
[D] (org.qbicc.plugin.reachability.rta) ANALYZE:   Deferred instance methods:  179
```
Main branch, build time init enabled,  21.5 MB executable
```
[D] (org.qbicc.plugin.reachability.rta) ANALYZE: RTA Reachability Statistics
[D] (org.qbicc.plugin.reachability.rta) ANALYZE:   Reachable interfaces:       173
[D] (org.qbicc.plugin.reachability.rta) ANALYZE:   Reachable classes:          1581
[D] (org.qbicc.plugin.reachability.rta) ANALYZE:   Instantiated classes:       1292
[D] (org.qbicc.plugin.reachability.rta) ANALYZE:   Initialized types:          1577
[D] (org.qbicc.plugin.reachability.rta) ANALYZE:   Invokable instance methods: 7142
[D] (org.qbicc.plugin.reachability.rta) ANALYZE:   Deferred instance methods:  179

```

This PR, build time init,  10 MB executable
```
[D] (org.qbicc.plugin.reachability.rta) ANALYZE: RTA Reachability Statistics
[D] (org.qbicc.plugin.reachability.rta) ANALYZE:   Reachable interfaces:       116
[D] (org.qbicc.plugin.reachability.rta) ANALYZE:   Reachable classes:          926
[D] (org.qbicc.plugin.reachability.rta) ANALYZE:   Instantiated classes:       764
[D] (org.qbicc.plugin.reachability.rta) ANALYZE:   Initialized types:          926
[D] (org.qbicc.plugin.reachability.rta) ANALYZE:   Invokable instance methods: 3756
[D] (org.qbicc.plugin.reachability.rta) ANALYZE:   Deferred instance methods:  118
```

ashu-mehra:
@dgrove-oss rta stats for the first and second config look the same but still there is difference in executable size. Any idea why is that so?

dgrove-oss:
> @dgrove-oss rta stats for the first and second config look the same but still there is difference in executable size. Any idea why is that so?
No....there are other changes that are conditional on build-time-init being enabled, so my guess would be some combination of them.  For example, I think when we enable build-time-init, we don't include all of the support for runtime `<clinit>` in the executable (like class init checks)

ashu-mehra:
Just a general comment - If I understand correctly here we are assuming that when BTI is enabled the class initializer have already been run. If that's the case should we have an assert (or something similar) in `RTAInfo` that the initializer for the class has already been processed at the places where we used to enqueue initializer, so that we don't ever end up with a case where we determine the class to be initialized but its initializer has not run.

ashu-mehra (approved):

ashu-mehra:
LGTM

PR #666, opened 2021-10-01
Many more compiler and interpreter fixes and improvements

dmlloyd:
* Fix inference of `Unsafe` field accesses
* Implement several more `Unsafe` atomics and fences
* Implementation of `Class.forName0`
* POSIX `time.h` and `sched.h` support
* Consolidate empty `init` natives
* Simple support for `auto`
* Bug fixes: compound type member extraction, Extract* should be schedulable, problems with `$_native` binding, redundant initialization, `Map.of()` doesn't like `null`s
* Halt the interpreter if the compiler fails
* Implement missing atomic instruction in LLVM
* Introduce `Value#isConstant` to help constant registration be more aggressive
* More OS-specific information on OS class
* Early initialization system properties setup
* Getting closer to having usable stack traces from in-interpreter exceptions


dgrove-oss (approved):
Scanned through commit by commit;  looks plausible.

PR #667, opened 2021-10-01
Serialize interpreter heap

dgrove-oss:
Serialize the values that the interpreter has computed for static fields.
Fix several bugs in VMBuildTimeHeap exposed by actually serializing interesting object graphs.

### DIFF
--- a/interpreter/src/main/java/org/qbicc/interpreter/impl/VmClassImpl.java
+++ b/interpreter/src/main/java/org/qbicc/interpreter/impl/VmClassImpl.java
@@ -140,7 +140,7 @@ class VmClassImpl extends VmObjectImpl implements VmClass {
             FieldElement field = typeDefinition.getField(i);
             if (field.isStatic()) {
                 Literal initValue = field.getInitialValue();
-                if (initValue instanceof ZeroInitializerLiteral) {
+                if (initValue == null || initValue instanceof ZeroInitializerLiteral) {
                     // Nothing to do;  memory starts zeroed.
                     continue;
                 }
--- a/interpreter/src/main/java/org/qbicc/interpreter/impl/VmClassImpl.java
+++ b/interpreter/src/main/java/org/qbicc/interpreter/impl/VmClassImpl.java
@@ -11,22 +11,31 @@ import io.smallrye.common.constraint.Assert;
 import org.qbicc.context.ClassContext;
 import org.qbicc.context.CompilationContext;
 import org.qbicc.graph.MemoryAtomicityMode;
+import org.qbicc.graph.literal.FloatLiteral;
+import org.qbicc.graph.literal.IntegerLiteral;
+import org.qbicc.graph.literal.Literal;
+import org.qbicc.graph.literal.StringLiteral;
+import org.qbicc.graph.literal.ZeroInitializerLiteral;
 import org.qbicc.interpreter.Thrown;
 import org.qbicc.interpreter.VmClass;
 import org.qbicc.interpreter.VmClassLoader;
 import org.qbicc.interpreter.VmInvokable;
 import org.qbicc.interpreter.VmObject;
+import org.qbicc.interpreter.VmString;
 import org.qbicc.interpreter.VmThrowable;
 import org.qbicc.plugin.coreclasses.CoreClasses;
 import org.qbicc.plugin.layout.Layout;
 import org.qbicc.type.ClassObjectType;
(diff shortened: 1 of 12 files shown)


interpreter/src/main/java/org/qbicc/interpreter/impl/VmClassImpl.java
```diff
@@ -121,6 +131,49 @@
...
+                        staticMemory.store64(member.getOffset(), val.doubleValue(), MemoryAtomicityMode.UNORDERED);
+                    }
+                } else if (initValue instanceof StringLiteral) {
+                    if (vm.bootstrapComplete) {
+                        VmString sv = vm.intern(((StringLiteral) initValue).getValue());
+                        staticMemory.storeRef(member.getOffset(), sv, MemoryAtomicityMode.UNORDERED);
+                    }
+                } else {
+                    // CONSTANT_Class, CONSTANT_MethodHandle, CONSTANT_MethodType
+                    vm.getCompilationContext().warning("Did not properly initialize interpreter memory for constant static field "+field);
```
dgrove-oss, line 173: This generates a _ton_ of warnings right now.  Many of them associated with Enums.

plugins/lowering/src/main/java/org/qbicc/plugin/lowering/Lowering.java
```diff
@@ -61,7 +65,7 @@ public GlobalVariableElement getGlobalForField(FieldElement fieldElement) {
         builder.setEnclosingType(enclosingType);
         final String sectionName;
         if (fieldElement.getType() instanceof ReferenceType) {
-            sectionName = GLOBAL_REFERENCES;
+            sectionName = CompilationContext.IMPLICIT_SECTION_NAME; // TODO: GLOBAL_REFERENCES;
```
dgrove-oss, line 68: This is a hack around a mach os problem with the way sections are named. I think the section needs to be named DATA,QBICC_GLOBALS when targeting mach os.

interpreter/src/main/java/org/qbicc/interpreter/impl/VmClassImpl.java
dgrove-oss: Fixed; forgot that a null return from getInitialValue means that the field is should just be zero initialized.

interpreter/src/main/java/org/qbicc/interpreter/impl/VmClassImpl.java
```diff
@@ -121,6 +131,49 @@
...
+                    if (val.getType().getSize() == 4) {
+                        staticMemory.store32(member.getOffset(), val.floatValue(), MemoryAtomicityMode.UNORDERED);
+                    } else {
+                        staticMemory.store64(member.getOffset(), val.doubleValue(), MemoryAtomicityMode.UNORDERED);
+                    }
+                } else if (initValue instanceof StringLiteral) {
+                    if (vm.bootstrapComplete) {
+                        VmString sv = vm.intern(((StringLiteral) initValue).getValue());
+                        staticMemory.storeRef(member.getOffset(), sv, MemoryAtomicityMode.UNORDERED);
+                    }
```
DanHeidinga, line 170: Does there need to be a pass over the loaded `VmClass`es to fix up the static String fields when `vm.bootstrapComplete` is set to true?

interpreter/src/main/java/org/qbicc/interpreter/impl/VmImpl.java
```diff
@@ -266,6 +269,9 @@
             }
             return clazz;
         });
+
+        bootstrapComplete = true;
+        throwableClass.initializeConstantStaticFields(); // Has constant String fields that can't be initialized when we first process the class
```
DanHeidinga, line 274: Ok, so we've got the start of the loop over loaded classes to fix this up here.  We can do the explicit list if it's easy enough to detect the classes that need it.

DanHeidinga (approved):
Overall, this looks reasonable to me.  There's one question about how to handle bootstrapping string constants for other classes, but it shouldn't block this moving foward

interpreter/src/main/java/org/qbicc/interpreter/impl/VmImpl.java
dgrove-oss: Yeah, it was easy to detect.  I didn't leave the printf in initializeConstantStaticFields(), but that's how I determined that Throwable was the only one that needed it done.

PR #670, opened 2021-10-01
Improve reachability analysis of interpreter heap

dgrove-oss:
Replace the conservative assumption that all types that were instantiated in the interpreter's heap are reachable with the proper optimistic algorithm that instead traces the portions of the heap that can be reached from static fields as those static fields become reachable. We will inspect each object in the interpreter heap at most once during this analysis (once an object is reachable from one static, it doesn't matter if it is also reachable from other statics).

This PR builds on top of #667.   The results seem almost too good to be true, so I decided to not incorporate this change into #667 in case I messed something up and didn't see it at first...

### DIFF
--- a/interpreter/pom.xml
+++ b/interpreter/pom.xml
@@ -32,11 +32,6 @@
         </dependency>
 
         <dependency>
-            <groupId>${project.groupId}</groupId>
-            <artifactId>qbicc-plugin-reachability</artifactId>
-        </dependency>
-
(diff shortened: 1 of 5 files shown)


dgrove-oss:
To review this separately, just look at a2a9bf7.

dgrove-oss:
rebased after #667 was merged to main.

dmlloyd (approved):

PR #672, opened 2021-10-04
Do not commit hotspot crash files

dmlloyd:


DanHeidinga (approved):

DanHeidinga:
Rerunning the CI job as the failure doesn't make sense for this PR - likely a hiccup downloading the dependency

PR #675, opened 2021-10-04
replace "reallyFinal" getstatics with interpreter-computed values

dgrove-oss:
Tested locally before updated with #666 and it appears to be doing something reasonable.  Posting for review, but as main is broken with --init-build-time atm, we might want to fix that before merging more --init-build-time cleverness ;)

### DIFF
--- a/plugins/optimization/src/main/java/org/qbicc/plugin/opt/InitializedStaticFieldBasicBlockBuilder.java
+++ b/plugins/optimization/src/main/java/org/qbicc/plugin/opt/InitializedStaticFieldBasicBlockBuilder.java
@@ -1,0 +1,35 @@
+package org.qbicc.plugin.opt;
+
+import org.qbicc.context.CompilationContext;
+import org.qbicc.graph.BasicBlockBuilder;
+import org.qbicc.graph.DelegatingBasicBlockBuilder;
+import org.qbicc.graph.MemoryAtomicityMode;
+import org.qbicc.graph.StaticField;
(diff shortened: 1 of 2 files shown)


dmlloyd (approved):

plugins/optimization/src/main/java/org/qbicc/plugin/opt/InitializedStaticFieldBasicBlockBuilder.java
```diff
@@ -0,0 +1,35 @@
...
+        this.ctxt = ctxt;
+    }
+
+    @Override
+    public Value load(ValueHandle handle, MemoryAtomicityMode mode) {
+        if (handle instanceof StaticField) {
+            final FieldElement fieldElement = ((StaticField) handle).getVariableElement();
+            if (fieldElement.isReallyFinal()) {
+                Value contents = fieldElement.getEnclosingType().load().getInitialValue(fieldElement);
+                if (contents instanceof Literal) {
```
DanHeidinga, line 27: I believe this covers the object literal cases as well but would appreciate confirmation.  Is the check that `contents instanceof Literal` effectively an assertion?  What else would it be here for a static field?

DanHeidinga: That makes sense - thanks Dave

DanHeidinga (approved):

PR #678, opened 2021-10-05
Fix inconsistent GEP behavior for function pointers

dmlloyd:
This should cause pointers-to-functions (`AddressOf(Function*Handle)`) to be emitted as the global symbol corresponding to the function without incorrectly trying to assign it to a local first.


theresa-m (approved):
Thanks David. Looks good to me.

PR #683, opened 2021-10-06
Internal annotation is processed as a RuntimeVisibleAnnotation

theresa-m:


ashu-mehra:
LGTM

PR #684, opened 2021-10-06
 interpreter support for UnsafeHandle

dgrove-oss:


dgrove-oss (commented):
This is working around a problem that surfaced in java.util.concurrent.ConcurrentHashMap.TreeBin.balanceInsertion.  The class file contained two variableTableEntries for what was really only 1 live range.  See discussion on Zulip on the interpreter topic: https://qbicc.zulipchat.com/#narrow/stream/284393-qbicc-development/topic/interpreter/near/256346299

PR #690, opened 2021-10-07
Update main codebase to Java 17

### LINKED ISSUE #688: Update qbicc itself to Java 17 (make sure the run time libs are still compiled targeting 11)

dmlloyd:
Fixes #688

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -28,6 +28,12 @@ jobs:
         run: |
           sudo apt-get install -y libunwind-dev
 
+      - name: Install OpenJDK 17
+        uses: joschi/setup-jdk@v2
+        with:
+          java-version: '17'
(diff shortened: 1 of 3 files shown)


dmlloyd:
> The accepted class file version need to be updated as well: https://github.com/qbicc/qbicc/blob/main/compiler/src/main/java/org/qbicc/type/definition/classfile/ClassFileImpl.java#L104 It causes a failure when running hello/world/Main with jbang.

I guess this is due to compiling the hello world source with Java 17. I think if you set `jbang default 11` it would not exhibit this issue.

We will have to increase the accepted class file version to 61 before we move to 17 (of course) but I'm hesitant to do that here because there could be structural changes that the class file parser isn't equipped to deal with, and that could make the change a lot more complex when I was hoping to keep it incremental.

PR #713, opened 2021-10-13
misc fixes for build time init

dgrove-oss:


dgrove-oss:
Taking a break for lunch; letting CI check to see that I didn't break anything in the normal path with this.  Will be more fixes this afternoon...

PR #720, opened 2021-10-16
disable rol/ror intrinsics for now; LLVM backend doesn't understand them

dgrove-oss:
Not sure what the intent of these intrinsics was, but they aren't implemented in the LLVM backend and a little quick googling seemed to indicate that they aren't LLVM instructions, so it wasn't obvious what to do with them.  We actually hit these with --init-build-time now; one of the hashcode methods uses Integer.rol.


dgrove-oss:
Folded into #719 because we need it there...

Issue #695, opened 2021-10-07
Fixing pointer dereferencing problems


dmlloyd:
Pointer dereferencing has a few problems that must be addressed (no pun intended).  In order for the Java level constructs to all work properly, the various eager `load` ops need to be eliminated from C-style element/member access chains.

Here are the steps:

- [x] #712
- [x] #754
- [x] #752
- [x] Ensure that field accesses to pointer-to-structure types translate e.g. `foo.bar` to `AddressOf(MemberOf(PointerHandle(<foo>), <bar>))`
- [x] Ensure that array element accesses to pointer types translate e.g. `foo[123]` to `AddressOf(ElementOf(PointerHandle(<foo>), 123))`
- [x] Fix `CNative.ptr#get()` to work identically to array element accesses (i.e. return the address of the element rather than the element itself)
- [x] Eliminate redundant loads used by `addr_of` and other similar constructs (a useless load elimination pass could suffice, however if the field happens to be `volatile` or something then such a pass would not be able to eliminate the load unless it knew for sure that it was only used for this purpose)
- [x] Ensure that the existing optimization of `PointerHandle(AddressOf(X)) -> X` still holds for above, so that e.g. `foo[123].bar` becomes `AddressOf(MemberOf(ElementOf(PointerHandle(<foo>), 123), <bar>))`

** If automatic promotion doesn't work out (due to too many ambiguities or something like that) then we'll have to explore another strategy, like introducing a new `Deref(ValueHandle)` node which represents a *possible* load and go from there.

dmlloyd:
> @dmlloyd what did you mean by CNative.load? That doesn't currently exist.

Sorry that was an error, I was referring to `CNative.ptr#get(*)`.

theresa-m:
Update from https://github.com/qbicc/qbicc/pull/772#pullrequestreview-810132283:

> We might have to pivot to a completely different approach, like abandoning Deref and transforming either or both of:
> 
> Load(MemberOf(Load(MemberOf(x, m1), NONE), m2), NONE) -> Load(MemberOf(MemberOf(x, m1), m2), NONE)
> ExtractMember(Load(x, NONE), m1) -> Load(MemberOf(x, m1), NONE) (this one is slightly complicated by the fact that the new load has to be placed with the original load due to ordering, which requires coordination between two phases: one to discover the member extraction and one to generate the load)
> ...and then sweeping up the unused whole-structure loads using a new load elimination pass. (Also apply similar transformations to array value nodes.)

[...]

PR #737, opened 2021-10-21
Add stack walker and intrinsics to access method data

ashu-mehra:
Add stack walker and provided intrinsics to access method data.
I have addressed the review comments mentioned in https://github.com/qbicc/qbicc/pull/709 viz:

1. Moved stack walker to runtime main module
2. Updated stack walker to handle frames for inlined methods (although it is not tested)

### DIFF
--- a/driver/src/main/java/org/qbicc/driver/CompilationContextImpl.java
+++ b/driver/src/main/java/org/qbicc/driver/CompilationContextImpl.java
@@ -283,13 +283,16 @@ final class CompilationContextImpl implements CompilationContext {
         DefinedTypeDefinition dtd = bootstrapClassContext.findDefinedType("org/qbicc/runtime/main/ObjectModel");
         if (dtd == null) {
             error("Can't find runtime library class: " + "org/qbicc/runtime/main/ObjectModel");
+            return null;
         }
         LoadedTypeDefinition helpers = dtd.load();
         int idx = helpers.findMethodIndex(e -> name.equals(e.getName()));
         if (idx == -1) {
             error("Can't find the runtime helper method %s", name);
+            return null;
         }
         return helpers.getMethod(idx);
+
     }
 
     public void enqueue(final ExecutableElement element) {
--- a/plugins/core-classes/src/main/java/org/qbicc/plugin/coreclasses/RuntimeMethodFinder.java
+++ b/plugins/core-classes/src/main/java/org/qbicc/plugin/coreclasses/RuntimeMethodFinder.java
@@ -1,0 +1,62 @@
+package org.qbicc.plugin.coreclasses;
+
+import org.qbicc.context.AttachmentKey;
+import org.qbicc.context.ClassContext;
+import org.qbicc.context.CompilationContext;
+import org.qbicc.type.definition.DefinedTypeDefinition;
+import org.qbicc.type.definition.LoadedTypeDefinition;
+import org.qbicc.type.definition.element.ConstructorElement;
+import org.qbicc.type.definition.element.MethodElement;
+import org.qbicc.type.descriptor.MethodDescriptor;
+
+public class RuntimeMethodFinder {
+    private static final AttachmentKey<RuntimeMethodFinder> KEY = new AttachmentKey<>();
+    private final CompilationContext ctxt;
+
+    private RuntimeMethodFinder(CompilationContext ctxt) {
+        this.ctxt = ctxt;
+    }
+
+    public static RuntimeMethodFinder get(CompilationContext ctxt) {
+        RuntimeMethodFinder helpers = ctxt.getAttachment(KEY);
+        if (helpers == null) {
+            helpers = new RuntimeMethodFinder(ctxt);
+            RuntimeMethodFinder appearing = ctxt.putAttachmentIfAbsent(KEY, helpers);
+            if (appearing != null) {
+                helpers = appearing;
+            }
+        }
+        return helpers;
+    }
+
+    public MethodElement getMethod(String runtimeClass, String helperName) {
+        ClassContext context = ctxt.getBootstrapClassContext();
+        DefinedTypeDefinition dtd = context.findDefinedType(runtimeClass);
+        if (dtd == null) {
+            ctxt.error("Can't find runtime library class: " + runtimeClass);
+            return null;
+        }
+        LoadedTypeDefinition ltd = dtd.load();
+        int idx = ltd.findMethodIndex(e -> helperName.equals(e.getName()));
+        if (idx == -1) {
+            ctxt.error("Can't find the runtime helper method %s", helperName);
+            return null;
+        }
+        return ltd.getMethod(idx);
+    }
+
+    public ConstructorElement getConstructor(String runtimeClass, MethodDescriptor descriptor) {
+        ClassContext context = ctxt.getBootstrapClassContext();
+        DefinedTypeDefinition dtd = context.findDefinedType(runtimeClass);
+        if (dtd == null) {
+            ctxt.error("Can't find runtime library class: " + runtimeClass);
+            return null;
+        }
+        LoadedTypeDefinition ltd = dtd.load();
+        int idx = ltd.findConstructorIndex(descriptor);
+        if (idx == -1) {
+            ctxt.error("Can't find the constructor with descriptor %s for class %s", descriptor.toString(), runtimeClass);
+        }
+        return ltd.getConstructor(idx);
+    }
+}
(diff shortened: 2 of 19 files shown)


ashu-mehra:
Without BTI we are going into infinite recursion. At some point during handling of invokedynamic bytecode, a `Throwable` instance is created which results in call to native `Throwable#fillInStackTrace()`. Since the interpreter does not register intrinsic for `Throwable#fillInStackTrace()`  when BTI is disabled, the call to native `fillInStackTrace` results in raising `UnsatisfiedLinkError` which result in calling `Throwable#fillInStackTrace()` again, resulting in infinite recursion.

With build time init on, this should work fine as the call to native `Throwable#fillInStackTrace()` is converted to no-op. So once #760 is resolved, it should pass the CI.

ashu-mehra:
This is ready for review.

plugins/intrinsics/src/main/java/org/qbicc/plugin/intrinsics/core/CoreIntrinsics.java
```diff
@@ -656,22 +660,239 @@ public static void registerJavaLangThrowableIntrinsics(CompilationContext ctxt)
...
+        intrinsics.registerIntrinsic(Phase.LOWER, mdDesc, "getMethodDesc", intToStringDesc, getMethodDesc);
+
+        String methodDataClass = "org/qbicc/runtime/stackwalk/MethodData";
+        MethodElement getLineNumberElement = ctxt.getRuntimeHelperMethod(methodDataClass, "getLineNumber");
+        MethodElement getMethodInfoIndexElement = ctxt.getRuntimeHelperMethod(methodDataClass, "getMethodInfoIndex");
+        MethodElement getFileNameElement = ctxt.getRuntimeHelperMethod(methodDataClass, "getFileName");
+        MethodElement getClassNameElement = ctxt.getRuntimeHelperMethod(methodDataClass, "getClassName");
+        MethodElement getMethodNameElement = ctxt.getRuntimeHelperMethod(methodDataClass, "getMethodName");
+
+        StaticIntrinsic fillStackTraceElement = (builder, target, arguments) -> {
```
dgrove-oss, line 872: Could some of these intrinsics be written in Java as "native" methods?  Might be more maintainable.

dgrove-oss (commented):
Generally seems reasonable to me, but not an expert on how libunwind...

ashu-mehra:
I added `typeId` to the method data so that we can populate the `StackTraceElement#declaringClassObject` field.

plugins/intrinsics/src/main/java/org/qbicc/plugin/intrinsics/core/CoreIntrinsics.java
ashu-mehra: yes, I agree and that was my intention initially and hence created https://github.com/qbicc/qbicc/pull/611 as accessing the native arrays does not work properly. But it has been closed in favor of https://github.com/qbicc/qbicc/issues/695. So once that gets resolved, I can replace these intrinsics with the Java methods.

dmlloyd (changes requested):
I'd like the extra compilation context methods to be moved elsewhere before going forward.

compiler/src/main/java/org/qbicc/context/CompilationContext.java
```diff
@@ -52,10 +54,15 @@ default ClassContext getClassContextForLoader(VmClassLoader classLoaderObject) {
...
 
+    /* TODO: Add method descriptor as a parameter */
+    MethodElement getRuntimeHelperMethod(String runtimeClass, String helperName);
+
     MethodElement getVMHelperMethod(String helperName);
 
     MethodElement getOMHelperMethod(String helperName);
 
+    ConstructorElement getRuntimeClassConstructor(String runtimeClass, MethodDescriptor descriptor);
+
```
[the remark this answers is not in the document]
ashu-mehra: I have added `RuntimeMethodFinder` to core classes plugin. Changes can be seen [here](https://github.com/ashu-mehra/qbicc/commit/f739e0a3abf620c27ba41ca05a7d164564afb4ea). Is that in line with your suggestion?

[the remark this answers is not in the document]
dmlloyd: Yes, that would be a great improvement, thanks.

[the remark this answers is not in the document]
ashu-mehra: With `RuntimeMethodFinder` in place, I guess we can remove `getVMHelperMethod` and `getOMHelperMethod` as well, but probably do it in a follow-up PR.

ashu-mehra:
This is ready for next round of review.

compiler/src/main/java/org/qbicc/context/CompilationContext.java
```diff
@@ -24,10 +24,12 @@
 import org.qbicc.type.TypeSystem;
 import org.qbicc.type.definition.DefinedTypeDefinition;
 import org.qbicc.type.definition.NativeMethodConfigurator;
+import org.qbicc.type.definition.element.ConstructorElement;
 import org.qbicc.type.definition.element.ExecutableElement;
 import org.qbicc.type.definition.element.FieldElement;
 import org.qbicc.type.definition.element.MemberElement;
 import org.qbicc.type.definition.element.MethodElement;
+import org.qbicc.type.descriptor.MethodDescriptor;
```
dmlloyd, line 32: I believe these imports are now unused.

compiler/src/test/java/org/qbicc/type/generic/TestClassContext.java
```diff
@@ -75,10 +76,14 @@ public ClassContext constructClassContext(final VmClassLoader classLoaderObject)
             return null;
         }
 
+        public MethodElement getRuntimeHelperMethod(String runtimeClass, String helperName) { return null; }
+
```
dmlloyd, line 80: These methods are now superfluous.

driver/src/main/java/org/qbicc/driver/CompilationContextImpl.java
```diff
@@ -56,6 +56,7 @@
 import org.qbicc.type.definition.element.MemberElement;
 import org.qbicc.type.definition.element.MethodElement;
 import org.qbicc.type.descriptor.ClassTypeDescriptor;
+import org.qbicc.type.descriptor.MethodDescriptor;
```
dmlloyd, line 59: This import seems to be unused.

plugins/core-classes/src/main/java/org/qbicc/plugin/coreclasses/RuntimeMethodFinder.java
```diff
@@ -0,0 +1,62 @@
...
+import org.qbicc.type.definition.LoadedTypeDefinition;
+import org.qbicc.type.definition.element.ConstructorElement;
+import org.qbicc.type.definition.element.MethodElement;
+import org.qbicc.type.descriptor.MethodDescriptor;
+
+public class RuntimeMethodFinder {
+    private static final AttachmentKey<RuntimeMethodFinder> KEY = new AttachmentKey<>();
+    private final CompilationContext ctxt;
+
+    public RuntimeMethodFinder(CompilationContext ctxt) {
```
dmlloyd, line 16:
```suggestion
    private RuntimeMethodFinder(CompilationContext ctxt) {
```

dmlloyd (commented):
Looks good now, just needs a couple minor cleanups.

compiler/src/test/java/org/qbicc/type/generic/TestClassContext.java
```diff
@@ -40,6 +40,7 @@
 import org.qbicc.context.ClassContext;
 import org.qbicc.type.definition.DefinedTypeDefinition;
 import org.qbicc.type.definition.NativeMethodConfigurator;
+import org.qbicc.type.definition.element.ConstructorElement;
```
dmlloyd, line 43: Here is one more unused import.

dmlloyd (commented):
There is one more. I'd recommend two things:

* Squash or rearrange the commits so that the only commits in the PR are the commits that implement the new functionality (with no cleanup commits or commits that undo some or all of a previous commit for some other reason); this ensures that there are few or no commits which don't actually function correctly and generally keeps things tidy
* Look at a `git diff upstream/main..HEAD` before pushing it up for review just to make sure no other cleanups got missed

ashu-mehra:
Squashed the commits to remove the intermediate ones.

dmlloyd (approved):

ashu-mehra:
After rebasing CI is passing now. Merging it as it has already been approved.

PR #738, opened 2021-10-21
More invokedynamic progress, also add logging for init failures

dmlloyd:
We actually attempt to construct indy bootstrap methods now. Also, interpreter threads must always be attached even in run-time-init mode.


dgrove-oss:
> Note that bootstrap method creation at build time is doomed to failure without build time init being enabled. But, it's no worse than the status quo.

Maybe time to consider switching the default to build time init enabled?

PR #739, opened 2021-10-21
implement VmClass.getDeclaredFields

dgrove-oss:
in conjunction with https://github.com/qbicc/qbicc-class-library/pull/36 implements enough of field/offset lookup to get Random's clinit to complete


dgrove-oss:
I've reduced the PR to just adding the interpreter support for `Class.getDeclaredFields` without trying to support getOffset on the Field instances.

dmlloyd (approved):

PR #756, opened 2021-10-27
Introduce futex support on Linux and POSIX condition support

dmlloyd:
First step towards getting `park`/`unpark` going on Linux (and now Mac too).

### DIFF
--- a/runtime/linux/src/main/java/org/qbicc/runtime/linux/Futex.java
+++ b/runtime/linux/src/main/java/org/qbicc/runtime/linux/Futex.java
@@ -1,0 +1,75 @@
+package org.qbicc.runtime.linux;
+
+import static org.qbicc.runtime.CNative.*;
+import static org.qbicc.runtime.linux.SysSyscall.*;
+import static org.qbicc.runtime.stdc.Stdint.*;
+import static org.qbicc.runtime.stdc.Time.*;
+
+/**
+ * Support for {@code futex(2)} on Linux.  See the manpage for more information.
+ */
+@include("<linux/futex.h>")
+public class Futex {
+
+    private static c_long futex(uint32_t_ptr uaddr, c_int futex_op, uint32_t val, const_struct_timespec_ptr timeout, uint32_t_ptr uaddr2, uint32_t val3) {
+        // no glibc wrapper!
+        return syscall(SYS_futex, uaddr, futex_op, val, timeout, uaddr2, val3);
+    }
+
+    /**
+     * Wait until the value at given address is equal to {@code val}.  Returns when woken (may be spurious).  Spurious
+     * wakeups may return {@code false} and set {@code errno} to {@code EAGAIN}, or may return {@code true}.  If the
+     * value at the test memory location already does not equal {@code val}, a wakeup will result.
+     *
+     * @param uaddr the address of the memory to test
+     * @param val the value to observe
+     * @param timeout the relative amount of time to wait
+     * @return {@code true} on success, or {@code false} on error (in {@code errno})
+     */
+    public static boolean futex_wait(uint32_t_ptr uaddr, uint32_t val, const_struct_timespec_ptr timeout) {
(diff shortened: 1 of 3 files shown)


runtime/linux/src/main/java/org/qbicc/runtime/linux/SysSyscall.java
```diff
@@ -0,0 +1,14 @@
...
+
+/**
+ *
+ */
+@include("<sys/syscall.h>")
+public class SysSyscall {
+    public static native c_long syscall(c_long number, object... args);
+
+    public static final c_long SYS_futex = constant();
+    public static final c_long SYS_futex_time64 = constant();
```
ashu-mehra, line 13: why do we need `SYS_futex_time64` constant, doesn't seem to be used anywhere.

runtime/linux/src/main/java/org/qbicc/runtime/linux/Futex.java
```diff
@@ -0,0 +1,75 @@
...
+     * wakeups may return {@code false} and set {@code errno} to {@code EAGAIN}, or may return {@code true}.  If the
+     * value at the test memory location already does not equal {@code val}, a wakeup will result.
+     *
+     * @param uaddr the address of the memory to test
+     * @param val the value to observe
+     * @param timeout the relative amount of time to wait
+     * @return {@code true} on success, or {@code false} on error (in {@code errno})
+     */
+    public static boolean futex_wait(uint32_t_ptr uaddr, uint32_t val, const_struct_timespec_ptr timeout) {
+        return futex(uaddr, word(FUTEX_WAIT.intValue() | FUTEX_PRIVATE_FLAG.intValue() | FUTEX_CLOCK_REALTIME.intValue()), val, timeout, zero(), zero()).longValue() != -1;
```
ashu-mehra, line 30: I am wondering if we should use `CLOCK_REALTIME` here as it be affected by NTP adjustments. Wouldn't `CLOCK_MONOTONIC` be more appropriate here?

dmlloyd: My reading of the usages of absolute timeouts in the JDK is that they are generally derived from `System.currentTimeMillis()` which would use the real time clock, not the monotonic clock.  Do we wait for the duration that is originally calculated between now and the deadline, or do we wait for the deadline wallclock time to pass?  (It's an honest question, I'm not really sure.) If the former, we could implement this method in terms of the relative timeout one.

runtime/linux/src/main/java/org/qbicc/runtime/linux/SysSyscall.java
dmlloyd: I implemented these constants first, before I knew what I was going to need. I could remove it.

runtime/linux/src/main/java/org/qbicc/runtime/linux/Futex.java
ashu-mehra: I would expect it to be former but I guess it would also depend on the higher level construct being built using the futex. We would want to be consistent in the type of clock being used across all operations. In that case would it be better to accept clock id as the parameter and set the flag accordingly?

dmlloyd: My concern is that when waiting for a deadline the JDK generally states that the deadline has to elapse.  If we go off of a projected duration from the start time, it would be possible for the deadline to not have elapsed when the method returns.  See `LockSupport.parkUntil(long)` and `Condition.awaitUntil(Date)`.

ashu-mehra: So the `deadline` in both APIs specifies time from the Epoch which would correspond to `CLOCK_REALTIME`. If the futex support is added only for such APIs then yeah we should use `CLOCK_REALTIME`.

ashu-mehra (approved):

PR #758, opened 2021-10-28
Runtime enhancements: LLVM intrinsics, `<math.h>`, inline assembly, NoReturn/NoThrow

dmlloyd:


dmlloyd:
Updated to include `@NoReflect` annotation and to detect annotations on fields and constructors.

dmlloyd:
Added the appropriate annotations to the various helper classes and now the stack traces look a bit more "authentic".

dgrove-oss (approved):
Looks reasonable to me.   

I think this introduces `asm` support machinery, but doesn't use actually use it anywhere yet. Is that right, or did I miss the usage?

dmlloyd:
> Looks reasonable to me.
> 
> I think this introduces `asm` support machinery, but doesn't use actually use it anywhere yet. Is that right, or did I miss the usage?

That's correct. There are a couple of places where we'll need it (LLVM doesn't have a way to emit an x86_64 `PAUSE` instruction to implement `Thread.onSpinWait` for example) but we don't have a place to implement it yet.  But since I was implementing all of the LLVM intrinsics, and that item has been on my whiteboard forever, I thought I'd get it out of the way at the same time.  We'll probably need a working patcher as discussed in #764 before we'll have a chance to use it though.

PR #760, opened 2021-10-28
Make build time initialization always-on

dmlloyd:
Switch over to always using build time init.  The run time init code is preserved until we solidify our strategy on split initialization; it is my feeling that the easiest way forward will be to supplement classes with a run time init stub in some way. I'll come up with a more formal proposal for the discussion board.

May require #759 and #757 in order to compile correctly.

### DIFF
--- a/plugins/reachability/src/main/java/org/qbicc/plugin/reachability/ReachabilityInfo.java
+++ b/plugins/reachability/src/main/java/org/qbicc/plugin/reachability/ReachabilityInfo.java
@@ -107,10 +107,6 @@ public class ReachabilityInfo {
         forceCoreClassesReachable(ctxt, true);
     }
 
-    public static void forceCoreClassesReachableRunTimeInit(CompilationContext ctxt) {
-        forceCoreClassesReachable(ctxt, false);
-    }
-
     private static void forceCoreClassesReachable(CompilationContext ctxt, boolean buildTimeInit) {
         ReachabilityInfo info = get(ctxt);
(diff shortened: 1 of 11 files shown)


plugins/reachability/src/main/java/org/qbicc/plugin/reachability/ReachabilityInfo.java
```diff
@@ -107,10 +107,6 @@ public static void forceCoreClassesReachableBuildTimeInit(CompilationContext ctx
         forceCoreClassesReachable(ctxt, true);
     }
 
-    public static void forceCoreClassesReachableRunTimeInit(CompilationContext ctxt) {
-        forceCoreClassesReachable(ctxt, false);
-    }
-
     private static void forceCoreClassesReachable(CompilationContext ctxt, boolean buildTimeInit) {
```
dgrove-oss, line 110: We could "constant-propagate" `buildTimeInit` is `true` into the ReachabilityBasicBlock builder and enable some minor cleanups

dgrove-oss (approved):
LGTM;  I think we should make the flip...

plugins/reachability/src/main/java/org/qbicc/plugin/reachability/ReachabilityInfo.java
dmlloyd: I tried to minimize the cleanups because this PR had a potential to live for a long time, and it would get annoying to rebase if other refactors would happen in the meantime. I'm in favor of following up with cleanups though.

PR #765, opened 2021-10-29
Track threads started in initializers and make them available

dmlloyd:
Just an incremental step over what's already there. Next step would be finding a solution for starting them up at run time (see #764 for more discussion).


dgrove-oss (approved):
LGTM

Issue #754, opened 2021-10-27
Make `AddrOf` & `Deref` consistent; solve associated load/store issues


theresa-m:

dmlloyd:
This issue has evolved a bit.  Quoting from the current PR iteration:

> I think neither this nor the following transformation `reference.getValueHandle()` are correct: not every argument to `referenceHandle` is guaranteed to have a value handle in turn.
> 
> I'm thinking we probably want these transformations in this class (I'll also include the valid one we already have here):
> 
> * `instanceFieldOf(instance, field)` ⇒ `memberOf(instance, member)` when `instance` is has a type of `CompoundType` (this one is already implemented & valid)
> * `referenceHandle(Deref(x))` ⇒ `pointerHandle(x)` (this should work in all cases because `x` must be a pointer for `Deref` to work)
> 
> Transitively, these two should automatically yield `instanceFieldOf(ReferenceHandle(Deref(x)), field)` ⇒ `memberOf(PointerHandle(x), member)` when `x` is a pointer to a structure, without having to do any additional transformations.  This should partially mitigate the problem, but not fully.
> 
> Still missing is a general solution a value is an immediate structure value: `MemberOf(ReferenceHandle(immediateStructure),member)`.  We know that we can't pass things that are not references through to the last `referenceHandle(xxx)` or else we'll get a class cast exception on the value type.  So we have to transform this one here, somehow.
> 
> Note that the same problem applies for `ElementOf(ReferenceHandle(thingThatIsNotAReference),index)` in the exact same way for arrays that `MemberOf` applies for compound types.
> 
> The ideal transformation would work like this:
> 
> * `memberOf(ReferenceHandle*(sv),member)` ⇒ `extractMember(sv,member)` where `sv` is a value of type `CompoundType` and `*` indicates an invalid/impossible node
> * `elementOf(ReferenceHandle*(av),index)` ⇒ `extractElement(av,index)` where `av` is a value of type `ArrayType` and `*` indicates an invalid/impossible node
> 
> However we can't actually construct these incorrect nodes.  There are two possible solutions I can think of:
> 
[...]

PR #772, opened 2021-10-29
Deref results in load, eliminated if element is accessed

theresa-m:
Fixes: #754 


plugins/intrinsics/src/main/java/org/qbicc/plugin/intrinsics/core/CoreIntrinsics.java
```diff
@@ -878,6 +878,8 @@ static void registerOrgQbiccRuntimeCNativeIntrinsics(final CompilationContext ct
             if (value instanceof Load) {
                 Load load = (Load) value;
                 return builder.addressOf(load.getValueHandle());
+            } else if (value.getType() instanceof PointerType) {
+                return value;
```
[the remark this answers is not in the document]
theresa-m: I was thinking since `deref` is now just returning instance instead of loading the value that `addr_of` should do something similar since they are dereference & reference. That way `addr_of(ptr).deref()` for example returns the expected outcome. What am I missing?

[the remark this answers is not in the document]
dmlloyd: I'm out for the weekend but maybe it would be a good idea to have a call next week to figure it out, wdyt?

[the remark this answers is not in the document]
theresa-m: Sounds great. I'll set something up.

theresa-m:
@dmlloyd is this what you had in mind? deref now works properly for dereferencing a CompoundType directly, or accessing a member. addr_of returns the original value if a deref block is passed.

compiler/src/main/java/org/qbicc/graph/Deref.java
```diff
@@ -1,23 +1,40 @@
...
+    Deref(Node callSite, ExecutableElement element, int line, int bci, Node dependency, Value value) {
         super(callSite, element, line, bci);
         this.value = value;
+
+        PointerType pt = (PointerType)value.getType();
+        this.toType = pt.getPointeeType();
+
+        this.valueHandle = new PointerHandle(callSite, element, line, bci, value);
+        // compound types cannot be loaded atomically
+        this.load = new Load(callSite, element, line, bci, dependency, valueHandle,
```
dmlloyd, line 26: We do not want nodes to "breed" other nodes, especially `Load` given that `Deref` is not itself an `OrderedNode`.  If the `Deref` only points to a delegate `Value` then the rest of the logic should still work without the pointer handle or the load.

compiler/src/main/java/org/qbicc/graph/MemberOf.java
```diff
@@ -20,7 +20,7 @@
 
         if (structureHandle instanceof PointerHandle ph && ph.getPointerValue() instanceof Deref deref) {
             /* unwrap value since this dereference should not cause an error */
-            this.structureHandle = new PointerHandle(callSite, element, line, bci, deref.getInput());
+            this.structureHandle = deref.getValueHandle();
         } else {
             this.structureHandle = structureHandle;
         }
```
dmlloyd, line 26: This section should be eliminated at this point and reverted back to the original simple assignment to `structureHandle`.

compiler/src/main/java/org/qbicc/graph/ValueVisitor.java
```diff
@@ -384,10 +384,6 @@ default R visit(T param, ZeroInitializerLiteral node) {
         return visitUnknown(param, node);
     }
 
-    default R visit(T param, Deref node) {
-        return visitUnknown(param, node);
-    }
-
```
dmlloyd, line 390: Following up to the above note, the visitor and copy methods should be retained.

compiler/src/main/java/org/qbicc/graph/Deref.java
```diff
@@ -36,10 +53,14 @@ public boolean equals(final Deref other) {
...
+        return toType;
+    }
+
+    public Load getLoad() {
+        return load;
     }
 
     public <T, R> R accept(final ValueVisitor<T, R> visitor, final T param) {
-        return visitor.visit(param, this);
+        return visitor.visit(param, this.getLoad());
```
dmlloyd, line 64: This is not allowed. Either `Deref` is a node of its own, or it's not; by bypassing its visitor method, things which analyze the node graph will likely get confused.

runtime/api/src/main/java/org/qbicc/runtime/CNative.java
```diff
@@ -815,12 +815,10 @@ public final boolean isNotPositive() {
          * Dereference the pointer, returning what the pointer points to. This operation
          * does not necessarily directly translate to a physical memory operation.
          *
-         * Pointers to structure types cannot be dereferenced, but members can be accessed as structPtr.deref().member
-         * which would be represented in C as structPtr->member.
          *
          * @return the pointed-to value
          */
-        public native T deref();
+        public static native <T extends object, P extends ptr<T>> T deref(ptr<T> obj);
```
dmlloyd, line 821: This should stay as an instance method. The intrinsic should still work identically as an instance intrinsic, except that the instance contains the pointer instead of the first argument.

plugins/verification/src/main/java/org/qbicc/plugin/verification/MemberResolvingBasicBlockBuilder.java
```diff
@@ -206,13 +207,8 @@ public Value checkcast(final Value value, final TypeDescriptor desc) {
...
             return ctxt.getLiteralFactory().undefinedLiteralOfType(castType);
-        } else if (value.getType() instanceof PointerType pt && castType.equals(pt.getPointeeType())) {
-            /* Pointer<x> -> x */
-            if (castType instanceof CompoundType) {
-                return super.deref(value);
-            } else {
-                return super.load(super.pointerHandle(value), MemoryAtomicityMode.UNORDERED);
-            }
+        } else if (value instanceof Deref) {
+            return value;
```
dmlloyd, line 211: This should be OK, but once this is all functional we might want to do a more complex validation between `value.getType()` vs `castType` to ensure that the input is assignable to the output.

dmlloyd (changes requested):
I think that the basic change to wrap values with the simple `Deref` node when `deref` is called, and to unwrap `Deref` nodes in `addr_of`, might be enough to remove the necessity for the other changes that I've marked here, allowing the implementation to be more uniform.  If the change no longer functions properly after applying my comments here, then let's talk about what is failing.

plugins/intrinsics/src/main/java/org/qbicc/plugin/intrinsics/core/CoreIntrinsics.java
```diff
@@ -878,6 +879,8 @@ static void registerOrgQbiccRuntimeCNativeIntrinsics(final CompilationContext ct
             if (value instanceof Load) {
                 Load load = (Load) value;
                 return builder.addressOf(load.getValueHandle());
+            } else if (value instanceof Deref deref) {
+                return deref.getInput();
```
dmlloyd, line 883: 👍🏻

theresa-m:
Hey David, I've applied most of the changes locally but I'm not sure now about where the load fits into this. 
Thinking about initializing Deref with a load such as`builder.deref(builder.load(builder.pointerHandle(instance)))`, marking when a deref is used as a member access and choosing to process as a load or just the instance in a later phase.

dmlloyd:
If `Deref` is purely wrapping its pointer value then I think we shouldn't have to care about loads at all in terms of a special case.  So we'd end up with (for example) `Deref(Load(LocalVar))` which is just fine.  Then we don't have to care about doing anything with `Deref` until it is stored or used in some other way.  For the near term we could implement `BBB#deref` in `LOWER` to emit a `Load`, but I think we could have a more clever transformation in the future maybe.

theresa-m:
Okay this is what I have so far which is generating mostly right. The load is being duplicated somewhere.

plugins/native/src/main/java/org/qbicc/plugin/native_/StructMemberAccessBasicBlockBuilder.java
```diff
@@ -19,6 +20,9 @@ public StructMemberAccessBasicBlockBuilder(CompilationContext context, BasicBloc
 
     public ValueHandle referenceHandle(Value reference) {
         if (reference.getType() instanceof CompoundType) {
+            if (reference instanceof Deref d) {
+                d.excludeLoad();
+            }
```
dmlloyd, line 25: I think neither this nor the following transformation `reference.getValueHandle()` are correct: not every argument to `referenceHandle` is guaranteed to have a value handle in turn.

I'm thinking we probably want these transformations in this class (I'll also include the valid one we already have here):

* `instanceFieldOf(instance, field)` ⇒ `memberOf(instance, member)` when `instance` is has a type of `CompoundType` (this one is already implemented & valid)
* `referenceHandle(Deref(x))` ⇒ `pointerHandle(x)` (this should work in all cases because `x` must be a pointer for `Deref` to work)

Transitively, these two should automatically yield `instanceFieldOf(ReferenceHandle(Deref(x)), field)` ⇒ `memberOf(PointerHandle(x), member)` when `x` is a pointer to a structure, without having to do any additional transformations.  This should partially mitigate the problem, but not fully.

Still missing is a general solution a value is an immediate structure value: `MemberOf(ReferenceHandle(immediateStructure),member)`.  We know that we can't pass things that are not references through to the last `referenceHandle(xxx)` or else we'll get a class cast exception on the value type.  So we have to transform this one here, somehow.

Note that the same problem applies for `ElementOf(ReferenceHandle(thingThatIsNotAReference),index)` in the exact same way for arrays that `MemberOf` applies for compound types.

The ideal transformation would work like this:

* `memberOf(ReferenceHandle*(sv),member)` ⇒ `extractMember(sv,member)` where `sv` is a value of type `CompoundType` and `*` indicates an invalid/impossible node
* `elementOf(ReferenceHandle*(av),index)` ⇒ `extractElement(av,index)` where `av` is a value of type `ArrayType` and `*` indicates an invalid/impossible node

However we can't actually construct these incorrect nodes.  There are two possible solutions I can think of:

1. Add another node type - call it `ImmediateHandle` - which will throw an exception if you try to take its pointer, or lower it, or use it in any way, which acts as a replacement for `ReferenceHandle` whenever the value is an immediate non-reference/non-pointer value, and implement the transformations in terms of this element (e.g. `Load(MemberOf(ImmediateHandle(x),m))` ⇒ `ExtractMember(x,m)`; with this approach we could leak these handles into the graph though, and cause late errors to be raised which might be harder to debug
2. Modify `MethodParser` to evaluate values before passing them to `referenceHandle` - if the value is an immediate value, then use the erstwhile super-secret (i.e. unused) `ExtractField` instead for structures (and `ExtractElement` for arrays); I favor this approach because it does not rely on having an invalid intermediate graph state (note that while `MethodParser` is a huge class, there are only six sites that actually do loads, and the only ones we actually care about are the ones for `OP_GETFIELD` and the `OP_*ALOAD` opcodes)

So let's assume for discussion we go with solution 2, and modify `MethodParser` to use `ExtractField` instead of `Load(InstanceFieldOf(...))` when the instance is not a reference.  In that case we can now easily implement several more transformations:

* `extractField(x, fld)` ⇒ `extractMember(x, member)` where `x.getType()` is `CompoundType`; this would be similar to the way `instanceFieldOf` is already implemented to resolve fields to members below
* `extractMember(Deref(ptr), member)` ⇒ `deref(memberOf(pointerHandle(ptr), member)` (i.e. recursively factor `Deref` to the outermost place)
* `extractElement(Deref(ptr), index)` ⇒ `deref(elementOf(pointerHandle(ptr), index)` (same thing but for arrays - note that some array handling is in `PointerBasicBlockBuilder` for historical reasons and probably needs cleaning up there)

These transformations all together should combine to reduce the problem where we are passing around whole structures (but probably not eliminate it).  I suspect we will have to deal with `Load(var)` where `var` is a whole compound structure next (typically in a local variable).  With the aforementioned `MethodParser` changes we could try the following additional transformations:

* `extractMember(Load(x,NONE),member)` ⇒ `load(memberOf(x,member),NONE)` for any `x`, but only when atomicity mode is `NONE`
* `extractElement(Load(x,NONE),index)` ⇒ `load(elementOf(x,index),NONE)` with same restriction

These are slightly sketchy because `Load` is ordered (even though it has no atomicity constraint) and so we're bending the rules a bit here by transforming a non-ordered node into an ordered node, which could go south on us in subtle or bizarre ways later on.  So this leads me to think that loads with no atomicity mode might be better implemented as a new node `LoadPlain`, but I'm not decided on that yet; anyway that would have a high similarity to `Deref` if `Deref` operated on handles instead of pointer values (which is a change that we might want to try *after* this is all sorted out).  So this is something we'll definitely revisit one way or another after this PR is done.

dmlloyd (commented):
More comments, mainly towards simplifying the `Deref` node, some correctness issues, and adding several transformations aimed at simplifying things overall.

plugins/intrinsics/src/main/java/org/qbicc/plugin/intrinsics/core/CoreIntrinsics.java
```diff
@@ -948,9 +952,17 @@ static void registerOrgQbiccRuntimeCNativeIntrinsics(final CompilationContext ct
...
 
-        intrinsics.registerIntrinsic(ptrDesc, "deref", MethodDescriptor.synthesize(classContext, nObjDesc, List.of()), identity);
         intrinsics.registerIntrinsic(ptrDesc, "asArray", MethodDescriptor.synthesize(classContext, ArrayTypeDescriptor.of(classContext, nObjDesc), List.of()), identity);
 
+        InstanceIntrinsic deref = (builder, instance, target, arguments) -> {
+            MemoryAtomicityMode mode = MemoryAtomicityMode.UNORDERED;
+            if (instance.getType() instanceof PointerType pt && pt.getPointeeType() instanceof CompoundType) {
+                mode = MemoryAtomicityMode.NONE;
+            }
+            return builder.deref(builder.load(builder.pointerHandle(instance), mode));
```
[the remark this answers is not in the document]
theresa-m: I'm still not sure where this load comes into play as it would be needed for any case that is not a compound type accessing a field.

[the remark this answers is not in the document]
dmlloyd: Honestly I'm no longer sure that it is even possible to do this correctly in this way.  The problem is that loads must always be ordered, so they cannot migrate out from their original position (at least, until we have better dependency links between memory operations). At the same time, we want `Deref` to be side-effect free i.e. not ordered. These two requirements seem fundamentally incompatible; either `Deref` becomes a redundant variation of `Load` or else we have invalid ordering. 
Even if we generate a load to keep in our back pocket in the event that we need it to be transformed, then the load is in the graph already and the `Deref` is not really doing anything useful.

We might have to pivot to a completely different approach, like abandoning `Deref` and transforming either or both of:

* `Load(MemberOf(Load(MemberOf(x, m1), NONE), m2), NONE)` -> `Load(MemberOf(MemberOf(x, m1), m2), NONE)`
* `ExtractMember(Load(x, NONE), m1)` -> `Load(MemberOf(x, m1), NONE)` (this one is slightly complicated by the fact that the new load has to be placed with the original load due to ordering, which requires coordination between two phases: one to discover the member extraction and one to generate the load)

...and then sweeping up the unused whole-structure loads using a new load elimination pass.  (Also apply similar transformations to array value nodes.)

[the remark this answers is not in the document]
theresa-m: I think you're right. I've updated this pr then to remove deref and put back `load(pointervalue(instance))` for the deref intrinsic. Combined with your recent MethodParser changes to use ExtractMember this is working fine to access compound members.
Since the transformations you mentioned would be optimizations I can update the orignal issues for those in a separate pull request.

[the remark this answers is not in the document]
dmlloyd: Sounds good. We can revisit later if we need to optimize it more for any reason.

plugins/verification/src/main/java/org/qbicc/plugin/verification/MemberResolvingBasicBlockBuilder.java
```diff
@@ -237,6 +230,8 @@ public Value checkcast(final Value value, final TypeDescriptor desc) {
             // we can't just bitcast it really, in fact it's an error unless the actual value is zero
             if (value instanceof Literal && ((Literal) value).isZero()) {
                 return ctxt.getLiteralFactory().zeroInitializerLiteralOfType(castType);
+            } else if (value instanceof Load) {
+                return value;
```
dmlloyd, line 234: This wouldn't be correct either, since it's possible to insert a definitely-incorrect cast on the result of a load. I think what we probably want to do here is add a check to see if the `castType` is equal to the value type being cast (or is a generic cast type like `CNative.object`), and if so, pass the value through.  Would that be enough, or was there another error you've observed here?

plugins/intrinsics/src/main/java/org/qbicc/plugin/intrinsics/core/CoreIntrinsics.java
```diff
@@ -1210,9 +1210,18 @@ static void registerOrgQbiccRuntimeCNativeIntrinsics(final CompilationContext ct
...
 
+        InstanceIntrinsic deref = (builder, instance, target, arguments) -> {
+            MemoryAtomicityMode mode = MemoryAtomicityMode.UNORDERED;
+            if (instance.getType() instanceof PointerType pt && pt.getPointeeType() instanceof CompoundType) {
+                /* structures cannot be loaded atomically */
+                mode = MemoryAtomicityMode.NONE;
+            }
+            return builder.load(builder.pointerHandle(instance), mode);
+        };
+        intrinsics.registerIntrinsic(ptrDesc, "deref", MethodDescriptor.synthesize(classContext, nObjDesc, List.of()), deref);
```
dmlloyd, line 1223: As long as we're in here and making at least one more change (below), please go ahead and rename this method from `deref` to `loadUnshared` and make it always use `MemoryAtomicityMode.NONE`.  Over this week's break (US Thanksgiving holiday) I plan to do a fairly extensive writeup of the JMM and how it applies to us and to LLVM and I expect we'll be adding more `load*` variants here afterwards.

plugins/verification/src/main/java/org/qbicc/plugin/verification/MemberResolvingBasicBlockBuilder.java
theresa-m: Yes that works.

plugins/intrinsics/src/main/java/org/qbicc/plugin/intrinsics/core/CoreIntrinsics.java
theresa-m: Sounds good. What is the meaning of Unshared for this?

theresa-m: Also derefAssign -> loadUnsharedAssign? Not sure if there are plans to use derefAssign.

dmlloyd: `derefAssign` will become `storeUnshared`.  `Unshared` refers to the memory semantics of the `NONE` memory ordering mode.

I'm working on a much more detailed writeup, but the quick version is:

* `MemoryAtomicityMode.NONE` -> "Unshared"
* `MemoryAtomicityMode.UNORDERED` -> "Plain" (per JMM) or "Unordered" (per LLVM)
* `MemoryAtomicityMode.MONOTONIC` -> "Opaque" (per JMM) or "Relaxed" (per C/C++) or "Monotonic" (per LLVM)

The others don't really have other names, but I'm trying to recenter around the JMM terminology whenever possible.  We'll see at least the skeleton API for accessing many more modes via the pointer API soon.

theresa-m: ooh right. store should have been obvious. Thanks for the details. I've updated the naming then and also rebased to resolve the merge conflict.

dmlloyd (approved):
Looks good. Thanks for sticking with this one!

dmlloyd:
Will run CI tomorrow; it's likely failing because of the time delay in publishing Maven artifacts.  This should be a one-time problem.

dmlloyd:
I reran CI and it is still failing; the error message is:

```

**Comment by theresa-m** [comment]
I'm not quite sure yet either. It happens when deref is StackWalker.java is renamed to loadUnshared. Other renames don't seem to have an issue.

**Comment by theresa-m** [comment]
I ended up adding the missing `MethodData` methods to VMHelpersSetupHook.java and that seems to have fixed it locally at least.

**Comment by dmlloyd** [comment]
Please also squash this PR down into one commit when you get a chance.

**Comment by theresa-m** [comment]
Still not quite sure about this error... after getting passed the last it looks like "deref" is lowered to
```
  %L1D = load i64, i64* %L1C, align 8, !dbg !43 ; StackWalker.java:22 bci@93
```
but renaming the method to `loadUnshared` lowers to:
```
  %L1C = load i64*, i64** %L3, align 8, !dbg !42 ; StackWalker.java:22 bci@92
  %L1D = call i8 addrspace(1)* (i8 addrspace(1)*, i64, i32) @exact.org.qbicc.runtime.gc.nogc.NoGcHelpers.allocate.ref.1.class.java-lang-Object.2.s64.s32(i8 addrspace(1)* %thr0, i64 64, i32 8) "statepoint-id"="12301", !dbg !42 ; StackWalker.java:22 bci@93
    ; StackWalker.java:22 bci@93
    ; StackWalker.java:22 bci@93
  call void (i8 addrspace(1)*, i8 addrspace(1)*, i64) @exact.org.qbicc.runtime.gc.nogc.NoGcHelpers.clear.v.2.ref.1.class.java-lang-Object.s64(i8 addrspace(1)* %thr0, i8 addrspace(1)* %L1D, i64 64) "statepoint-id"="12303", !dbg !42 ; StackWalker.java:22 bci@93
  %L1F = bitcast i8 addrspace(1)* %L1D to %class.java.lang.NoSuchMethodError addrspace(1)*, !dbg !42 ; StackWalker.java:22 bci@93
  %L20 = getelementptr %class.java.lang.NoSuchMethodError, %class.java.lang.NoSuchMethodError addrspace(1)* %L1F, i32 0, i32 2, !dbg !42 ; member typeId
  store atomic i32 795, i32 addrspace(1)* %L20 unordered, align 4, !dbg !42 ; StackWalker.java:22 bci@93
  %L21 = getelementptr %class.java.lang.NoSuchMethodError, %class.java.lang.NoSuchMethodError addrspace(1)* %L1F, i32 0, i32 3, !dbg !42 ; member nativeObjectMonitor
  store i64 0, i64 addrspace(1)* %L21, align 8, !dbg !42 ; StackWalker.java:22 bci@93
  %L22 = getelementptr %class.java.lang.NoSuchMethodError, %class.java.lang.NoSuchMethodError addrspace(1)* %L1F, i32 0, i32 2, !dbg !42 ; member typeId
  store atomic i32 795, i32 addrspace(1)* %L22 unordered, align 4, !dbg !42 ; StackWalker.java:22 bci@93
  %L23 = getelementptr %class.java.lang.NoSuchMethodError, %class.java.lang.NoSuchMethodError addrspace(1)* %L1F, i32 0, i32 3, !dbg !42 ; member nativeObjectMonitor
  store i64 0, i64 addrspace(1)* %L23, align 8, !dbg !42 ; StackWalker.java:22 bci@93
  call void (i8 addrspace(1)*, i8 addrspace(1)*) @init.java.lang.NoSuchMethodError.0(i8 addrspace(1)* %thr0, i8 addrspace(1)* %L1D) "statepoint-id"="12305", !dbg !42 ; StackWalker.java:22 bci@93
  %L26 = getelementptr %class.java.lang.Thread, %class.java.lang.Thread addrspace(1)* %L25, i32 0, i32 25, !dbg !42 ; member thrown
  store i8 addrspace(1)* %L1D, i8 addrspace(1)* addrspace(1)* %L26, align 8, !dbg !42 ; StackWalker.java:22 bci@93
  call i32 (%struct._Unwind_Exception*) @_Unwind_RaiseException(%struct._Unwind_Exception* %L2A) noreturn "statepoint-id"="12309", !dbg !42 ; StackWalker.java:22 bci@93
```

dmlloyd:
That looks like the call to `loadUnshared` got changed to `NoSuchMethodError`... maybe there's a problem with the intrinsic, like a descriptor mismatch or something?

theresa-m:
Based on the recent [zulip convo](https://qbicc.zulipchat.com/#narrow/stream/284393-qbicc-development/topic/progress/near/263163257) I'm removing my changes to VMHelpersSetupHook. I deprecated `deref` and added loadUnshared as a rename. I'm not sure I understand what this issue could have to do with class libraries though.

dmlloyd (approved):

dmlloyd:
> Based on the recent [zulip convo](https://qbicc.zulipchat.com/#narrow/stream/284393-qbicc-development/topic/progress/near/263163257) I'm removing my changes to VMHelpersSetupHook. I deprecated `deref` and added loadUnshared as a rename. I'm not sure I understand what this issue could have to do with class libraries though.

It's not obvious. The class libraries depend on the release version of Qbicc, including its run time modules... meaning that integration tests are currently testing the *current* Qbicc with the *last release* of Qbicc's run time modules rather than the ones just built.  This is not ideal, but I haven't come up with a good way to override those dependencies yet so that it uses the run time modules of the just-built Qbicc instead.  I think it might be possible to override the resolver somehow (maybe by overriding the version system property), but I haven't gotten to experiment with that yet.

PR #776, opened 2021-11-01
Refactor j.l.Class intrinsics to invoke ObjectModel helpers

michihirohorie:
Refactoring to remove code duplication with `ObjectModel` helpers


dmlloyd (approved):
Seems reasonable. Using the first builder allows the intrinsics to be recursively expanded.

PR #778, opened 2021-11-01
post #760 cleanup of Reachability

dgrove-oss:
Simplify reachability analysis by removing conditional logic for
processing <clinit> methods at either build time or runtime.


ashu-mehra (approved):
Looks good.

PR #780, opened 2021-11-02
Avoid one potential deadlock during method compilation

dmlloyd:


dmlloyd:
The original code was mainly meant to avoid redundant work, but deadlocks are decidedly worse than redundant work :)

PR #781, opened 2021-11-02
interpreter implementation of getClassAccessFields

dgrove-oss:


dgrove-oss:
Needs https://github.com/qbicc/qbicc-class-library/pull/41 (so we don't bail out too early on the native method)

interpreter/src/main/java/org/qbicc/interpreter/impl/VmImpl.java
```diff
@@ -524,6 +524,10 @@ public void initialize() {
                 VmClassImpl clazz = cl.loadClass(def.getInternalName());
                 return clazz;
             });
+            reflectClass.registerInvokable("getClassAccessFlags", (thread, target, args) -> {
+                VmClassImpl cls = (VmClassImpl)args.get(0);
+                return cls.getTypeDefinition().getModifiers();
```
dmlloyd, line 529: This needs to be masked with `0x1FFF` as per the spec in `jdk.internal.reflect.Reflection#getClassAccessFlags` in order to prevent our extra flags from leaking out.

dmlloyd (changes requested):

dmlloyd:
The change is OK but we do need to mask those bits.

interpreter/src/main/java/org/qbicc/interpreter/impl/VmImpl.java
dgrove-oss: fixed.

dmlloyd (approved):

PR #782, opened 2021-11-03
Eliminate long class path arguments to main

dmlloyd:
Replace the boot class path mechanism with one that allows Maven artifact coordinates and automatically installs the required run time libraries.

This change eliminates all hard-coded artifact version strings from docs, examples, and tests. It also unifies test logging, because the Maven resolving libraries won't log correctly otherwise.


dmlloyd:
This change requires https://github.com/qbicc/qbicc-class-library/pull/42 to be merged before it will succeed.

dgrove-oss:
I merged qbicc/qbicc-class-library#42 and kicked this CI job to run again.

dgrove-oss (approved):

PR #785, opened 2021-11-04
Add VMHelpers methods that raise an exception as entry points

ashu-mehra:
Some of the methods like `VMHelpers#raiseUnsatisfiedLinkError`
may be referenced post lower phase only. Therefore if we rely on manually
enqueuing such methods, then it has to be done in each phase.
Instead its better if we register them as entry points so that they get
queued in each phase.


DanHeidinga:
> Some of the methods like `VMHelpers#raiseUnsatisfiedLinkError`
> may be referenced post lower phase only. Therefore if we rely on manually
> enqueuing such methods, then it has to be done in each phase.

We already add the methods in both the `ADD` and `ANALYZE` phases.  Are they being dead code eliminated as there are no callers until the Lower phase?  ie: this is only showing up in the itable / vtable slots for unimplemented natives?

I'm OK with the approach you're taking here as it will be a very minor amount of always present runtime code but would like to see the comment describing why these helpers are different added to the code - too easy to lose context otherwise.

dgrove-oss:
Approving because trunk is broken without this change.  Can clean up comment later.

dgrove-oss (approved):

PR #786, opened 2021-11-04
Complete constant folding support

dmlloyd:
The earlier rework to reachability plus the interpreter have largely eliminated wrong-platform class compilation, but finishing up constant-folded method calls here puts the last nail in.


dgrove-oss (approved):
LGTM

compiler/src/main/java/org/qbicc/type/definition/classfile/ClassFile.java
```diff
@@ -117,6 +117,10 @@
      * For executable members which never throw - not even {@link StackOverflowError} or {@link OutOfMemoryError}.
      */
     int I_ACC_NO_THROW = 1 << 26;
+    /**
+     * For executable members which should be evaluated during compilation.
+     */
+    int I_ACC_FOLD = 1 << 27;
```
dgrove-oss, line 123: Almost out of bits :(

PR #787, opened 2021-11-04
Correctly support JAR class path elements

dmlloyd:
This should fix the problem of compiling a main class in a JAR.


dgrove-oss:
Tested locally.  It does allow a jar to be added as a file element.

dgrove-oss (approved):

PR #788, opened 2021-11-04
Don't add -lunwind when targeting macos

dgrove-oss:
The unwind functions we want are in the system libs; adding an
extra flag just confuses the linker because there is no libunwind.


dmlloyd (approved):
LGTM but @ashu-mehra should verify.

dmlloyd:
Nice to see `unless/when` being used!

ashu-mehra (approved):
Looks fine

PR #790, opened 2021-11-04
Exclude Hidden methods from stack trace

ashu-mehra:
~Do not generate method data for methods marked as `Hidden` so that
they are not included in stack trace.~
Store method modifiers in method data so that the frames for methods
marked as hidden can be skipped when creating stack trace.
Annotate following methods with `Hidden` annotation:
- `org.qbicc.runtime.Main#main`
- `org.qbicc.runtime.stackwalk.JavaStackWalker#getFrameCount`
- `org.qbicc.runtime.stackwalk.JavaStackWalker#walkStack`
    
Also exclude exception constructors and `Throwable.fillInStackTrace`
method from stack trace

### DIFF
--- a/plugins/methodinfo/src/main/java/org/qbicc/plugin/methodinfo/MethodDataEmitter.java
+++ b/plugins/methodinfo/src/main/java/org/qbicc/plugin/methodinfo/MethodDataEmitter.java
@@ -67,7 +67,7 @@ public class MethodDataEmitter implements Consumer<CompilationContext> {
             methodName = ((FunctionElement)element).getName();
         }
         String fileName = element.getSourceFileName();
-        String className = element.getEnclosingType().getInternalName();
+        String className = element.getEnclosingType().getInternalName().replace('/', '.');
         String methodDesc = element.getDescriptor().toString();
         int typeId = element.getEnclosingType().load().getTypeId();
 
@@ -85,7 +85,7 @@ public class MethodDataEmitter implements Consumer<CompilationContext> {
         SymbolLiteral mdLiteral = btHeap.getSerializedVmObject(vm.intern(methodDesc));
         Assert.assertNotNull(mdLiteral);
 
-        return methodData.add(new MethodInfo(fnLiteral, cnLiteral, mnLiteral, mdLiteral, typeId));
+        return methodData.add(new MethodInfo(fnLiteral, cnLiteral, mnLiteral, mdLiteral, typeId, element.getModifiers()));
     }
 
     private int createSourceCodeInfo(CompilationContext ctxt, MethodData methodData, ExecutableElement element, int lineNumber, int bcIndex, int inlinedAtIndex) {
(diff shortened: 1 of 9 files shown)


plugins/methodinfo/src/main/java/org/qbicc/plugin/methodinfo/MethodDataEmitter.java
```diff
@@ -97,9 +98,15 @@ private int createSourceCodeInfo(CompilationContext ctxt, MethodData methodData,
         int sourceCodeIndex;
         ExecutableElement element = node.getElement();
         if (node.getCallSite() == null) {
+            if (element.hasAllModifiersOf(ClassFile.I_ACC_HIDDEN)) {
```
ashu-mehra, line 101: Another way to skip methods annotated as `Hidden` would be to store the `modifiers` value in method data and do the check for `ClassFile.I_ACC_HIDDEN` in `JavaStackWalker#skipFrame`. This would result in extra 4 bytes for every entry in `qbicc_method_info_table`.

runtime/main/src/main/java/org/qbicc/runtime/stackwalk/JavaStackWalker.java
```diff
@@ -1,36 +1,59 @@
...
+        StackFrameVisitor javaStackWalker = new JavaStackWalker(exceptionObject, visitor);
         StackWalker.walkStack(javaStackWalker);
     }
 
+    private boolean skipFrame(int scIndex) {
+        int minfoIndex = MethodData.getMethodInfoIndex(scIndex);
+        String methodName = MethodData.getMethodName(minfoIndex);
+        String className = MethodData.getClassName(minfoIndex);
+        if (exceptionObject.getClass().getName().equals(className)) {
+            // skip frames for excepption constructor or fillInStackTrace methods
```
dmlloyd, line 34: The exception constructor may be on the stack more than one time; I think what we want to do here is to skip the topmost frame exception frame(s) instead.

plugins/methodinfo/src/main/java/org/qbicc/plugin/methodinfo/MethodDataEmitter.java
ashu-mehra: I checked up `qbicc_method_info` structure and it appears there is already a 4 byte of padding at the end to align the struct to 64-bits. I think storing method `modifiers` in method data would be a cleaner solution. The overhead would be the method data for hidden methods which would not be needed for stack trace. Not sure if it would be useful for any other purpose. In any case I expect that overhead in memory should be very low and hence acceptable.

dmlloyd: I like that solution; it could be useful for other purposes as well.

runtime/main/src/main/java/org/qbicc/runtime/stackwalk/JavaStackWalker.java
ashu-mehra: Fixed it.

plugins/methodinfo/src/main/java/org/qbicc/plugin/methodinfo/MethodDataEmitter.java
ashu-mehra: Updated the commit to follow this.

dmlloyd (approved):

PR #792, opened 2021-11-04
No longer need qbicc-runtime-api dep in hello.world example

dgrove-oss:
We don't need the dependency, and having it actually is a problem for doing
```
 --boot-path-append-file $(jbang info classpath examples/helloworld/hello/world/Main.java)
```
 because `--boot-path-append-file` doesn't support currently support adding a `:` separated list of files.


dmlloyd (approved):
Looks good. I'm thinking maybe the path arguments should accept `:`/`;`-separated lists in addition to the name change we talked about...

PR #797, opened 2021-11-05
Replace SymbolLiteral with ProgramObjectLiteral

### LINKED ISSUE #577: Replace last usages of SymbolLiteral with data handles
We should introduce `ValueHandle` implementations for `Data` and `DataDeclaration`.  These would mirror the ones for `Function` and `FunctionDeclaration`, and generally allow us to get rid of `SymbolLiteral`.

dmlloyd:
* Remove all occurrences of `SymbolLiteral`
* Add mechanism to pre-establish an exact `FunctionElement` for a `MethodElement` specifically to make it easier to find the personality function
* Clearly differentiate between value and symbol types in program objects
* Introduce `ProgramObjectLiteral` to replace `SymbolLiteral`
* Cache all program object literals by their corresponding declaration
* Eagerly "stuff" `Data` and `Function` with their declaration when late-replacing a declaration with a definition
* Provide a general method to get the declaration for any program object (returns self by default)
* Prevent address space from being changed after the pointer type of a program object is obtained
* Add efficient convenience methods to produce declarations from existing declarations and definitions; reuse the original declaration object in every program module section that needs that declaration
* Get rid of the global `getCurrentThreadLocalSymbolLiteral()` method from `CompilationContext`

As a follow up, we should consider eliminating `FunctionHandle`, `FunctionDeclarationHandle`, `DataHandle`, and `DataDeclarationHandle` in favor of `PointerHandle(ProgramObjectLiteral)`.  Both forms presently occur in the program graph but they're fundamentally equivalent.

Fixes #577.


dgrove-oss (approved):
Nice improvement!

PR #808, opened 2021-11-09
Delete files if they're too big

### LINKED ISSUE #806: Missing } on .dot for FormatSpecifier.print method
`add.dot` for `java/util/Formatter$FormatSpecifier/methods/print.id46` lacks a closing `}`

```
subgraph cluster_b96 {
label = "b96";
n499 [shape="rectangle",style="diagonals, filled",label="goto",fixedsize="shape"]
n500 [label="get-and-set"]

```

Is it a missing terminator, or some other issue? Depth is capped at 500...

galderz:
Closes #806

* This avoids confusion when files are partially written,
and you think the dot node visitor has a bug instead.
* Also fixed some formatting on local variables.
* Too big graphs take too long to convert to svg,
so discarding option to add a system property to expand
the configured counts in dot node visitor.

### DIFF
--- a/plugins/dot/src/main/java/org/qbicc/plugin/dot/DotGenerator.java
+++ b/plugins/dot/src/main/java/org/qbicc/plugin/dot/DotGenerator.java
@@ -120,24 +120,35 @@ public class DotGenerator implements ElementVisitor<CompilationContext, Void>, C
             return;
         }
         Path path = dir.resolve(phase.toString() + ".dot");
-        try (BufferedWriter bw = Files.newBufferedWriter(path, StandardCharsets.UTF_8, StandardOpenOption.CREATE, StandardOpenOption.TRUNCATE_EXISTING)) {
-            bw.write("digraph {");
-            bw.newLine();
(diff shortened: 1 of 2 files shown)


plugins/dot/src/main/java/org/qbicc/plugin/dot/DotGenerator.java
```diff
@@ -138,6 +138,11 @@ private void process(final MemberElement element, MethodBody methodBody) {
             failedToWrite(ctxt, path, cause);
         } catch (TooBigException e) {
             log.debugf("Element \"%s\" is too big to graph", element);
+            try {
+                Files.delete(path);
```
dmlloyd, line 142: Some operating systems will not let you delete a file while it is open. I would suggest putting this catch block on a try block that encloses this one to avoid that possibility.

dmlloyd (commented):
Only one comment, otherwise looks good.

galderz:
@dmlloyd Updated PR

dmlloyd (approved):

PR #817, opened 2021-11-15
Add offset property to pointer handles

dmlloyd:
This fixes various problems with the supposition that `ElementOf(PointerHandle(x), i)` modifies the pointer instead of what is pointed to.


dgrove-oss (approved):
LGTM - nice simplification!

PR #818, opened 2021-11-15
Multiple optimization and verification improvements

dmlloyd:
Introduce optimizations which:

* Convert bitcasts to the same type as the first element or member of an array or structure to a GEP construct
* Allow `Neg` to be canonical for integer negation
* Transform `If`->`Goto`x2->`Phi` diamonds to `Select`
* Improve the quality and accuracy of simple transformations in the simple optimizer
* Remove an invalid conversion
* Introduce some new simple optimizations
* Remove the old unused pointer arithmetic constructs

Introduce static verification BBB which prevents invalid or buggy constructs from being emitted.  Also, remove some dead code.


dgrove-oss:
Looks like ArithmeticCompare test is unhappy.

dgrove-oss (approved):

PR #824, opened 2021-11-16
Add VMHelpers to perform type tests purely with type_ids

dmlloyd:
This PR lays the groundwork for supporting `Class.isInstance`, `Class.cast`, and `Class.isAssignableFrom`.  This change was achieved largely, but not *completely*, through safe non-functional IDE refactors, so it does require a second set of eyes.

### DIFF
--- a/runtime/main/src/main/java/org/qbicc/runtime/main/VMHelpers.java
+++ b/runtime/main/src/main/java/org/qbicc/runtime/main/VMHelpers.java
@@ -76,35 +76,32 @@ public final class VMHelpers {
     // Invariant: value is not null
     @NoSideEffects
     @Hidden
-    private static boolean isAssignableTo(Object value, type_id toTypeId, uint8_t toDimensions) {
-        type_id valueTypeId = ObjectModel.type_id_of(value);
-        int intDim = toDimensions.intValue();
-        if (intDim == 0) {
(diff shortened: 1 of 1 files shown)


dmlloyd:
Depends on #820.

dgrove-oss (commented):
I think this change is ok as long as we are setting up the implements array for the `[ref` pseudo-type so that `ObjectModel.does_implement([ref, ...)` will return true for serializable and cloneable.   At least at some point in the past we weren't doing this, so we have special case code to handle that case.  This case is tested in the integration test, so if that is passing, then it is probably fine.

dmlloyd:
> I think this change is ok as long as we are setting up the implements array for the `[ref` pseudo-type so that `ObjectModel.does_implement([ref, ...)` will return true for serializable and cloneable. At least at some point in the past we weren't doing this, so we have special case code to handle that case. This case is tested in the integration test, so if that is passing, then it is probably fine.

I'm in the process of verifying this now.  We don't want to treat the ref array class itself any differently from any other class to the maximal extent possible, and the same goes for the primitive array classes.  The generated classes do implement those interfaces so it should be straightforward but I'm still pulling on the thread.

dmlloyd:
This has been verified to work with array-of-array assignability per the `Arrays` integration test.

dgrove-oss (approved):

PR #830, opened 2021-11-17
Add delegating element builder interfaces

dmlloyd:


DanHeidinga (approved):
Code looks fine.  Not sure I follow the use case (yet) though.  Is there an issue / description for why this new interface is needed?

dmlloyd:
> Code looks fine. Not sure I follow the use case (yet) though. Is there an issue / description for why this new interface is needed?

The patcher is going to work by intercepting type and element builders and manipulating them in various ways.  Having the delegating interfaces makes this much simpler, in the same vein as having a delegating BBB or visitor.

PR #832, opened 2021-11-17
Introduce LLVM opaque pointer support

### LINKED ISSUE #586: LLVM 13: Opaque pointers
When LLVM 13 is released, it will (likely) introduce opaque (typeless) pointers.  At that time we can move our reference and pointer emission to use these pointer types instead of typed pointers, which will reduce the amount of bitcasting we'll need to do.

Note that this does not affect debug information, which uses DWARF information for typing.

dmlloyd:
Note that this doesn't actually work with LLVM 13; there are some bugs that are fixed in LLVM trunk relating to literals with pointer members and function types.  So I'm leaving this in draft for now.

Fixes #586.


dmlloyd:
Just tried this again with LLVM 13.0.1 and it's still broken. The release of LLVM 14 is imminent though so I'll give it another try then.

dmlloyd:
This one will have to wait for 15. It looks like 14 will never work: https://godbolt.org/z/dcrxWdW1x

evacchi:
FWIW it looks like it won't crash if you leave out `--rewrite-statepoints-for-gc` so that must be the faulty pass...

dgrove-oss (approved):

PR #833, opened 2021-11-17
Fix word comparison methods

dmlloyd:
This was a very bad cut & paste snafu :-(


dgrove-oss (approved):
Apparently the reviewer of #820 was also asleep at the wheel.  😞

PR #834, opened 2021-11-17
Allow field elements to declare a run time initializer

dmlloyd:


compiler/src/main/java/org/qbicc/type/definition/element/FieldElement.java
```diff
@@ -11,10 +11,12 @@
 public final class FieldElement extends VariableElement implements MemberElement {
     public static final FieldElement[] NO_FIELDS = new FieldElement[0];
     private final Literal initialValue;
+    private final InitializerElement runTimeInitializer;
 
     FieldElement(BuilderImpl builder) {
         super(builder);
         this.initialValue = builder.initialValue;
+        this.runTimeInitializer = builder.runTimeInitializer;
```
DanHeidinga, line 19: Should we assert that `runTimeInitializer`'s are only added for static fields?

dmlloyd: Possibly, but not here.  The modifiers of a field element can be changed after the fact, so we'd want to do this later as some kind of verification, if anywhere.

DanHeidinga: > The modifiers of a field element can be changed after the fact

We change static to non-static?  Or vice versa?  I'm kind of surprised by the assertion the modifiers can change as we should be reading them from the `.class` file

dmlloyd: Modifiers can change because sometimes we add properties to the field, or otherwise finish defining it, after the builders are complete.  For example we change `System.out` and friends to be not-really-final, and add various discovered flags to methods in response to annotations we've read from those methods.  While we don't routinely toggle the `static` flag, it is nevertheless a modifier and thus subject to the same behavioral contract.  In any event, validation is not presently a significant concern since run time initializers are only going to be added by one point in the code for now (the patcher), and once the patcher is in, we could conceivably replace all the points where we toggle modifier flags after construction with it (or with type builder interception, generally) thus allowing us to tighten the contract.

DanHeidinga: Flipping the static-ness of a field invalidates the bytecodes that access the field (it either doesn't need a receiver or it does).

But sure, we can change it later.

DanHeidinga (approved):

PR #836, opened 2021-11-17
Improve nullability detection

dmlloyd:


dgrove-oss (approved):
I'm missing the big picture of exactly what the semantics of `getValueIfTrue` and `getValueIfFalse` are and how they are used but otherwise it makes sense.

compiler/src/main/java/org/qbicc/graph/IsNe.java
```diff
@@ -15,14 +15,14 @@
...
-        if (input.getType() instanceof ReferenceType) {
+        if (input.getType() instanceof NullableType) {
             if (input.equals(getLeftInput()) && getRightInput() instanceof ZeroInitializerLiteral ||
                 input.equals(getRightInput()) && getLeftInput() instanceof ZeroInitializerLiteral) {
                 // todo: maybe require a BBB input instead
                 return new NotNull(input.getCallSite(), input.getElement(), input.getSourceLine(), input.getBytecodeIndex(), input);
             }
         }
-        return input;
+        return super.getValueIfTrue(input);
```
DanHeidinga, line 25: Is the super call here correct?  Won't it call `AbstractBooleanCompare`'s implementation which is checking `equals(input)` when we actually want the negation of that for `IsNe`?

dmlloyd: It is checking whether `input` is equal to the condition itself. The method `getValueIfFalse` is called to determine what the actual value of `input` is when the condition is `false`. If `input` is referring to the condition itself, then the result is `false` reflexively.

dmlloyd:
> I'm missing the big picture of exactly what the semantics of `getValueIfTrue` and `getValueIfFalse` are and how they are used but otherwise it makes sense.

I will add some documentation tomorrow.

dmlloyd:
I've added documentation to the `BooleanValue` interface which explains the functionality.  I will also file a follow-up issue regarding expanding constraints beyond just `NotNull`.

compiler/src/main/java/org/qbicc/graph/IsNe.java
DanHeidinga: Ok.  This is cannonicalizing the condition to either a true / false if possible and then we'll check if the condition holds in https://github.com/qbicc/qbicc/blob/4d333d08cfeaed9bf7414d616643142a5aa55958/compiler/src/main/java/org/qbicc/graph/If.java#L26-L38

Got it now.

DanHeidinga (approved):

PR #839, opened 2021-11-18
Commutative equals/hashCode for commutative binary values

dmlloyd:
Ideally we'd do this in a more object-oriented manner, however the commutative binary operations do not always have a superclass slot available, and it's not really possible to correctly implement equals/hashCode as interface default methods.  This gets the job done and keeps the mess localized.


dgrove-oss (approved):
LGTM

PR #841, opened 2021-11-19
A number of improvements to boolean values & a unit test fwk

dmlloyd:
* Introduce a general unit test utility
* Introduce `Comp` ones-complement node, use it as canonical form for boolean inversion
* Introduce several transformations to optimize and/or canonicalize boolean forms
* Allow any value to have a value-if-true/false
* Front-load `InstanceOf` value-if-true (allows optimizations to simplify it)
* Add the start of unit tests for individual optimization transformations using the unit test utility

### DIFF
--- a/compiler/src/main/java/org/qbicc/graph/InstanceOf.java
+++ b/compiler/src/main/java/org/qbicc/graph/InstanceOf.java
@@ -4,7 +4,6 @@ import java.util.Objects;
 
 import org.qbicc.type.BooleanType;
 import org.qbicc.type.ObjectType;
-import org.qbicc.type.ReferenceType;
 import org.qbicc.type.definition.element.ExecutableElement;
 
 /**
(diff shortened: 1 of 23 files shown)


dmlloyd:
Looks like it broke in rebase. Will fix after lunch

dmlloyd:
Fixed

compiler/src/main/java/org/qbicc/graph/InstanceOf.java
```diff
@@ -13,15 +12,17 @@
 public final class InstanceOf extends AbstractValue implements InstanceOperation, OrderedNode, BooleanValue {
     private final Node dependency;
     private final Value input;
+    private final Value valueIfTrue;
```
dgrove-oss, line 15: I expected that the addition of `valueIfTrue` would have also meant that InstanceOf had two value dependencies.  Are we going to get into trouble because `valueIfTrue` is "invisible" to some fairly generic part of the compiler that uses the valueCount/getValue to traverse the input values?

dmlloyd: It does not, for a couple of reasons.

* The value-if-true is always regenerated whenever the node is created so if it is never copied, there is no functional difference to this node whether or not the value-if-true is ever copied.
* A node dependency of any kind imposes an ordering constraint.  Since this node does not depend on its value-if-true in an ordering sense, it is not needed for that purpose either.

In the event that the value-if-true *is* actually copied, then effect is that there may be two different instances of the same node around (adding a deduplication cache to `SimpleBBB` has been on my whiteboard for a while now but I haven't gotten around to it, nor to transferring that to an actual issue for someone else... maybe it's time for a whiteboard purge 🙂)

dgrove-oss: Got it.  Thanks

dgrove-oss (approved):

PR #845, opened 2021-11-19
Change current thread read to a value handle

### LINKED ISSUE #844: Replace `CurrentThreadRead` with a handle
### Original description

`CurrentThreadRead` - this one was ordered from the start which is good. I'd drop this node in favor of a `CurrentThreadHandle` node plus `Load`. This one can't be eliminated until lowering, because that's the point where we know where the current thread is stashed - either the load gets replaced with a straight value, or the handle gets translated to a native thread-local global variable.  By changing this to a handle, we can also potentially bind the current thread by storing to the handle.

### Additional Notes

This handle should always be accessed with an atomicity node of `NONE` since it is always thread local.  The handle should be considered read-only (i.e. permanently cacheable) except in `FunctionElement` where it is writable.  The core plugin should map loads directly to the parameter value unless the original element is a `FunctionElement`, in which case it should be read each time from the associated thread-local global variable.

The type of the handle value should be reference-to-`java.lang.Thread`, not the thread itself.

dmlloyd:
Fixes #844

### DIFF
--- a/compiler/src/main/java/org/qbicc/graph/BasicBlockBuilder.java
+++ b/compiler/src/main/java/org/qbicc/graph/BasicBlockBuilder.java
@@ -151,8 +151,6 @@ public interface BasicBlockBuilder extends Locatable {
 
     ParameterValue parameter(ValueType type, String label, int index);
 
-    Value currentThread();
-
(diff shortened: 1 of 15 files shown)


dgrove-oss (approved):
LGTM

PR #847, opened 2021-11-23
Add name and description to all qbicc pom.xmls

### LINKED ISSUE #773: POM names and descriptions
This is somewhat more minor, but we can't deploy to Maven Central unless it is fixed.

Each POM in the project should have a `name` element, e.g. `Qbicc plugin: Try/Catch` and a `description` e.g. `Plugin which handles try/catch constructs for Qbicc`.

dgrove-oss:
Fixes #773.

### DIFF
--- a/compiler/pom.xml
+++ b/compiler/pom.xml
@@ -11,6 +11,8 @@
     </parent>
 
     <artifactId>qbicc-compiler</artifactId>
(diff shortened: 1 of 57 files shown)


dgrove-oss:
redundant with larger PR to be able to publish to maven central

PR #862, opened 2021-11-29
Accessor support

dmlloyd:


dgrove-oss (approved):
I see that this is providing the compiler implementation for the `Accessor` and `AccessWith` annotations from the patcher api.  The mechanics look plausible, but I'm missing the big picture of the use case for these two annotations (ie, why we need them as distinct operations as opposed to injecting an accessor with with a `Add`.  Is the use case to allow us to remove a field from a JDK class by replacing it with a method that computes the value instead of storing it?  To allow us to use a field of a different type to store a value, while still allowing client code to access it as-if there was a field of the expected type?

dmlloyd:
> I see that this is providing the compiler implementation for the `Accessor` and `AccessWith` annotations from the patcher api. The mechanics look plausible, but I'm missing the big picture of the use case for these two annotations (ie, why we need them as distinct operations as opposed to injecting an accessor with with a `Add`.) Is the use case to allow us to remove a field from a JDK class by replacing it with a method that computes the value instead of storing it?

Yes.

> To allow us to use a field of a different type to store a value, while still allowing client code to access it as-if there was a field of the expected type?

Also yes.  The motivating currently relevant example is `org.qbicc.runtime.stdc.Errno` though.

PR #863, opened 2021-11-29
Evaluate conditional annotations lazily to avoid VM bootstrap issues

dmlloyd:


plugins/patcher/src/main/java/org/qbicc/plugin/patcher/MethodDeleteInfo.java
```diff
@@ -0,0 +1,37 @@
...
+        return descriptor;
+    }
+
+    Annotation getAnnotation() {
+        return annotation;
+    }
+
+    @Override
+    public Location getLocation() {
+        return ClassContextPatchInfo.getMethodLocation(internalName, "<init>");
```
dgrove-oss, line 35:
```suggestion
        return ClassContextPatchInfo.getMethodLocation(internalName, name);
```

dgrove-oss: Looks like this is a copy & paste mistake from ConstructorDeleteInfo??

dgrove-oss (approved):
Looks good modulo what I think is a cut&paste error

plugins/patcher/src/main/java/org/qbicc/plugin/patcher/MethodDeleteInfo.java
dmlloyd: Good catch...

PR #868, opened 2021-11-29
Remove unreachable if branch

dmlloyd:


dgrove-oss:
CI agrees with you that this branch isn't needed, but after poking around for a few minutes I'm not finding where we do the rewrite from TypeType to integral types...... how is this working?

dmlloyd:
`TypeType` is a `WordType` so that earlier similar branch already handles the case.

dgrove-oss (approved):
Got it; thanks!

PR #870, opened 2021-11-29
Relax CAS type check for null values

dmlloyd:


dgrove-oss (approved):
Had to think a little bit about what this boolean expression was doing.  Maybe it's just me, but the nested combination of `!` `&&` and `||` means that the changed code, although correct, is pretty hard to read.

dmlloyd:
I've updated the PR with clearer boolean expressions.

PR #872, opened 2021-11-29
Patcher fixes

dmlloyd:


plugins/patcher/src/main/java/org/qbicc/plugin/patcher/ClassContextPatchInfo.java
```diff
@@ -103,7 +106,6 @@ void processClasses(ClassContext classContext, Iterator<String> iterator) {
     private static final int K_REPLACE = 3;
 
     void processClass(final ClassContext classContext, final String className) {
-        ConditionEvaluation ce = ConditionEvaluation.get(classContext.getCompilationContext());
```
dgrove-oss, line 106: This is a missing part to the previous change to defer conditional evaluation to avoid bootstrapping problems?

dgrove-oss (approved):

plugins/patcher/src/main/java/org/qbicc/plugin/patcher/ClassContextPatchInfo.java
dmlloyd: Yes I think you're right. It must have gotten lost in the shuffle.

PR #873, opened 2021-11-29
Begin migration to per-initializer InitCheck

dmlloyd:


plugins/instanceof-checkcast/src/main/java/org/qbicc/plugin/instanceofcheckcast/LowerClassInitCheckBlockBuilder.java
```diff
@@ -31,26 +32,29 @@ public LowerClassInitCheckBlockBuilder(final CompilationContext ctxt, final Basi
...
-    public Node initCheck(final ObjectType objectType) {
-        doInitCheck(objectType);
-        return nop();
-    }
+    public Node initCheck(InitializerElement initializer) {
+        if (initializer.hasAllModifiersOf(ClassFile.I_ACC_RUN_TIME)) {
+            // generate a run time init check
+
+            // and call the initializer
+            call(initializerOf(initializer), List.of());
```
DanHeidinga, line 40: The approach looks good.  Is the part that limits the initializer to only being executed once with this approach still TBD?

dgrove-oss (approved):
LGTM.
Went back and read https://github.com/qbicc/qbicc/discussions/764 to remind myself what we are doing :). Nice to see the checks only applying to static field access now as planned.

PR #874, opened 2021-11-29
Use `LockSupport`-based monitors with full API support

dmlloyd:
This implementation allows thread interruption and conditions to work properly, in theory.

There may be a dependency on some other PRs but I wanted to get this one up here for review sooner rather than later. #866 doesn't do much without this (because those methods are native and we can't update the class libs without this API).


dmlloyd:
This looks like it's going to depend on class library changes.

PR #875, opened 2021-11-30
Introduce once one-time action class

dmlloyd:


runtime/main/src/main/java/org/qbicc/runtime/main/Once.java
```diff
@@ -0,0 +1,60 @@
...
+                // we're already running it
+                return;
+            }
+            synchronized (lock) {
+                done = this.done;
+                if (! done) try {
+                    Runnable target = this.target;
+                    if (target != null) {
+                        target.run();
+                    }
```
DanHeidinga, line 45: Do we want to enforce that `target` is never null?

runtime/main/src/main/java/org/qbicc/runtime/main/Once.java
```diff
@@ -0,0 +1,60 @@
...
+                    // volatile write publishes thrown
+                    this.done = true;
+                    throw t;
+                }
+            }
+        }
+        Throwable thrown = this.thrown;
+        if (thrown != null) {
+            throw thrown;
+        }
```
DanHeidinga, line 58: Can we move this code into the synchronized block?  The common case will be that the `Once` operation has already run and succeeded so coding this so the fast path only does one check would be better.

If we don't set `this.done = true` in the catch block, and moved this check to the start of the `synchronized` block, failed cases would be slower but overall runtime would be better

runtime/main/src/main/java/org/qbicc/runtime/main/Once.java
dmlloyd: I wasn't sure, but I guess it's a reasonable default which can be revisited later.

DanHeidinga (approved):

dgrove-oss (approved):
LGTM

PR #885, opened 2021-12-01
Remove dependency from run time to compiler

dmlloyd:


dmlloyd:
This causes a number of extra JARs to be added to the bootstrap class path. On the upside, it did help me find a bug in the Maven dependency resolver.

DanHeidinga (approved):
I hate duplicating constants like this due to potential inconsistencies later but that doesn't seem like a good reason to block this.

approved

PR #886, opened 2021-12-01
Add initializer replacement capability to patcher

dmlloyd:


plugins/patcher/src/main/java/org/qbicc/plugin/patcher/ClassPatchInfo.java
```diff
@@ -70,6 +72,10 @@ MethodDeleteInfo getDeletedMethodInfo(final String name, final MethodDescriptor
         return deletedMethods.getOrDefault(name, Map.of()).get(descriptor);
     }
 
+    boolean isDeletedInitializer() {
```
DanHeidinga, line 75: Doesn't this also require the lock?  The setter asserts the lock is held so the getter should as well:
```
assert Thread.holdsLock(this);
```

DanHeidinga (approved):
otherwise this looks fine to me

plugins/patcher/src/main/java/org/qbicc/plugin/patcher/ClassPatchInfo.java
dmlloyd: Yep, I missed this. Fixed.

PR #888, opened 2021-12-01
Simplify lowering of memory accesses

dmlloyd:
Simplify lowering of memory accesses by allowing handles to be transformed directly. This was not previously possible due to how memory-based boolean accesses were translated. By moving that translation to the node copier, we can go back to a more straightforward memory access scheme.

The boolean access problem could be more elegantly solved in the future by allowing `BooleanType` to be a word of any size.  Most in-memory booleans would become `BooleanType[8]` while in-register booleans would be `BooleanType[1]`.

The reason I didn't immediately go with this approach is that we use booleans in a comparatively large number of places, so it would require a considerable amount of auditing and would be a fairly far-reaching change.  But I think it might be worth a revisit at some point.


dmlloyd:
Another feature of this PR is the introduction of per-reference-array compound types. This allows another simplification relating to how references to arrays are cast and used - by moving the bit cast to the reference transformation, we eliminate the need to bitcast the elements as we pull them out.

This change helps code gen right now, but it might become obsolete once LLVM moves to opaque pointers by default.  Once we reach that point, we can determine whether to pull out this feature.

PR #898, opened 2021-12-02
Fix method parsing problem when attributes appear in unusual order

dmlloyd:


DanHeidinga (approved):
Change looks straightforward.  Not sure I understand the bug though =)

dmlloyd:
The attributes aren't actually parsed at this time, so the buffer position has to be moved past the end of the attribute. It just happened to work before because of the order that the attributes appear (the position was only ever invalid in the last attribute so it didn't matter), but this breaks as soon as `javac` decides to do something else - or as soon as you have something other than `javac` producing your class files.

PR #899, opened 2021-12-02
Introduce pure-Java Monitor implementation

dmlloyd:
Replaces #874.  Only introduce the new API; wait for release cadence before doing any more.


dgrove-oss (approved):
LGTM

PR #904, opened 2021-12-03
compiler support for Array.newArray

dgrove-oss:


dgrove-oss:
Put it in draft mode because when I looked at the LLVM IR for the intrinsic, I noticed that all stores to initialize object headers are being emitted twice.  I think this is an existing bug in main (because we run the BasicInitializationBasicBlockBuilder twice: once in ADD and once in ANALYZE), but I need to confirm that this wasn't caused by anything I did on this branch...

dmlloyd:
Yeah I think this is an existing bug.  It probably needs a bit of a rethink unfortunately.

dgrove-oss:
I thought about it this morning; I think I see how to fix the double initialization.

dmlloyd (approved):

PR #905, opened 2021-12-03
New `AccessMode` API

dmlloyd:
Our current usage of access modes are heavily based on what is provided by LLVM and follow a rather haphazard mix of Java and LLVM memory model semantics.  This is an attempt to clean up and organize the access modes, as a prelude to not only the corresponding cleanup in the LLVM-compatible BBB, but also as a prelude to correctly implementing the various `Unsafe`/`VarHandle`/`Atomic*` classes by way of new methods coming on to `ptr` that @theresa-m's PR #772 alludes to.

A more detailed writeup is forthcoming in a discussion topic but I wanted to get the starter code posted in a draft PR to start to socialize the concepts a bit.  The primary feature of this code change is using Java's type system to enforce usage of specific  categories of access modes in specific situations; for example, a `load` operation would require a mode that implements at least `ReadAccessMode`.

As a final note, right now, `CmpAndSwap` is specified in terms of a "success" mode and "failure" mode, and `GetAndSet*` are specified in terms of a single mode that implicitly applies (in different ways) to both the read and write operation. However, the JMM specification (at least the parts defined in the JDK) generally specify CAS and RMW operations in terms of the read and (conditional in the case of CAS) write side.  So the expectation is that we'd do the same here, and (in the RMW case) interpret the combination of read and write operation mode to emit the correct equivalent-or-next-strongest LLVM mode.  In the CAS case this generally amounts to translating the combination of `XxxAcquire` and `XxxRelease` to `XxxAcqRel`.

### DIFF
--- a/compiler/src/main/java/org/qbicc/graph/MemoryAtomicityMode.java
+++ b/compiler/src/main/java/org/qbicc/graph/MemoryAtomicityMode.java
@@ -1,43 +1,56 @@
 package org.qbicc.graph;
 
+import static org.qbicc.graph.atomic.AccessModes.*;
+
+import org.qbicc.graph.atomic.AccessMode;
+import org.qbicc.graph.atomic.AccessModes;
+
 /**
  * The mode for a memory access operation.
+ *
+ * @deprecated Use the modes in {@link AccessModes} instead.
  */
+@Deprecated
 public enum MemoryAtomicityMode {
     /**
      * Non-atomic access (cannot be given for fence or read-modify-write operations).
      */
-    NONE,
+    NONE(SingleUnshared),
     /**
      * Plain access (cannot be given for fence or read-modify-write operations); any value that is read was written by
      * some other operation.
      */
-    UNORDERED,
+    UNORDERED(SinglePlain),
     /**
      * Single total order for a given memory address (cannot be given for fence operations).
      */
-    MONOTONIC,
+    MONOTONIC(SingleOpaque),
     /**
      * Like {@link #MONOTONIC} but also forms a synchronization edge with a corresponding {@link #RELEASE} operation.
      */
-    ACQUIRE,
+    ACQUIRE(SingleAcquire),
     /**
      * Like {@link #MONOTONIC} but also forms a synchronization edge with a corresponding {@link #ACQUIRE} operation.
      */
-    RELEASE,
+    RELEASE(SingleRelease),
     /**
      * Acts as both {@link #ACQUIRE} and {@link #RELEASE} on one operation.
      */
-    ACQUIRE_RELEASE,
+    ACQUIRE_RELEASE(SingleAcqRel),
     /**
      * Like {@link #ACQUIRE_RELEASE} but applying a global total order.
      */
-    SEQUENTIALLY_CONSISTENT,
+    SEQUENTIALLY_CONSISTENT(SingleSeqCst),
     /**
      * Java {@code VOLATILE} access.
      */
-    VOLATILE
-    ;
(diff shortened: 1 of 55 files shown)


compiler/src/main/java/org/qbicc/graph/atomic/AccessModes.java
```diff
@@ -0,0 +1,82 @@
...
+     * any <em>acquire</em> operation on the same memory address.
+     * This mode may only be used for write operations.
+     */
+    public static final WriteAccessMode SingleRelease = WriteModes.SingleRelease;
+    /**
+     * The <em>acquire-release</em> single access mode.
+     * This mode combines {@link #SingleAcquire} with {@link #SingleRelease}.  It is not usable by itself
+     * but is useful for compatibility with the previous scheme.
+     */
+    public static final AccessMode SingleAcqRel = PureModes.SingleAcqRel;
```
dmlloyd, line 74: This mode is also potentially useful for lowering `CmpAndSwap` to LLVM.

compiler/src/main/java/org/qbicc/graph/atomic/AtomicTraits.java
```diff
@@ -0,0 +1,25 @@
+package org.qbicc.graph.atomic;
+
+/**
+ * A set of bits used to represent the set of capabilities that an atomic operation may have, in order to determine which
+ * operations include which other operations.
+ */
```
dmlloyd, line 6: I implemented this all in one place so that it's easier to analyze and debug problems.

compiler/src/main/java/org/qbicc/graph/MemoryAtomicityMode.java
```diff
@@ -47,6 +60,15 @@ public static MemoryAtomicityMode min(MemoryAtomicityMode a, MemoryAtomicityMode
...
     }
 
+    /**
+     * Get the closest corresponding access mode, which may have to be transformed further before use.
+     *
+     * @return the access mode (not {@code null})
+     */
+    public AccessMode getAccessMode() {
+        return  accessMode;
+    }
```
dmlloyd, line 70: This method should hopefully allow us to more easily bridge over to the new API.

compiler/src/main/java/org/qbicc/graph/atomic/AccessModes.java
```diff
@@ -0,0 +1,190 @@
...
+ * <p>
+ */
+public final class AccessModes {
+    private AccessModes() {}
+
+    /**
+     * The <em>load-load</em> global access mode.  When used as a fence, this mode prevents any
+     * loads which succeed the fence from being moved before any global loads which precede the fence in the original program order:
+     * {@code AnyLoad ≺ GlobalLoadLoad < AnyGlobalLoad}.
+     */
```
dmlloyd, line 101: I think I may have these definitions inverted or am missing the nuance of preventing preceding operations from moving past the fence to a succeeding position.

dmlloyd: Fixed... I think!

dmlloyd:
I've added examples of how `Fence` and `Load` can be adapted to the new API without requiring an immediate and complete rewrite of all the callers of `fence` and `load` on BBB.

dmlloyd:
I've added `CmpAndSwap` as a more complex example.  Also, I've added an `AccessMode.combineWith()` method which simplifies bridging from JMM semantics to LLVM semantics.

dgrove-oss (approved):
LGTM.  Nice to get to a JMM-centric API.

PR #906, opened 2021-12-03
restructure header initialization to avoid doing it twice

dgrove-oss:
A fairly minimal change that fixes the double-initialization problem (https://github.com/qbicc/qbicc/pull/904#issuecomment-985192707).

I'm tempted to attempt a cleanup where we push `manuallyInitialize` inside of the implementation of VmClass.newInstance to DRY things up... any a priori reason that won't work?

### DIFF
--- a/plugins/core-classes/src/main/java/org/qbicc/plugin/coreclasses/ArrayLengthBasicBlockBuilder.java
+++ b/plugins/core-classes/src/main/java/org/qbicc/plugin/coreclasses/ArrayLengthBasicBlockBuilder.java
@@ -1,0 +1,26 @@
+package org.qbicc.plugin.coreclasses;
+
+import org.qbicc.context.CompilationContext;
+import org.qbicc.graph.BasicBlockBuilder;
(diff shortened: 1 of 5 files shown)


plugins/core-classes/src/main/java/org/qbicc/plugin/coreclasses/ArrayLengthBasicBlockBuilder.java
```diff
@@ -0,0 +1,26 @@
...
+public class ArrayLengthBasicBlockBuilder extends DelegatingBasicBlockBuilder {
+    private final CompilationContext ctxt;
+
+    public ArrayLengthBasicBlockBuilder(CompilationContext ctxt, BasicBlockBuilder delegate) {
+        super(delegate);
+        this.ctxt = ctxt;
+    }
+
+    @Override
+    public ValueHandle lengthOf(ValueHandle array) {
```
dgrove-oss, line 19: There is exactly one use of `lengthOf`...from MethodParser because it can't depend on CoreClasses.

dmlloyd (approved):

PR #911, opened 2021-12-03
use camelCase names for runtime helper methods and intrinsics

dgrove-oss:
There should be no semantic change here; just lots of methods being moved and/or renamed.

### DIFF
--- a/interpreter/src/main/java/org/qbicc/interpreter/impl/VmImpl.java
+++ b/interpreter/src/main/java/org/qbicc/interpreter/impl/VmImpl.java
@@ -444,7 +444,7 @@ public final class VmImpl implements Vm {
             // VMHelpers
             VmClassImpl vmHelpersClass = bootstrapClassLoader.loadClass("org/qbicc/runtime/main/VMHelpers");
 
-            vmHelpersClass.registerInvokable("getClass", (thread, target, args) -> ((VmObjectImpl) args.get(0)).getVmClass());
+            vmHelpersClass.registerInvokable("getClass", 1, (thread, target, args) -> ((VmObjectImpl) args.get(0)).getVmClass());
             vmHelpersClass.registerInvokable("classForName", (thread, target, args) -> {
                 VmClassLoaderImpl classLoader = (VmClassLoaderImpl) args.get(2);
                 if (classLoader == null) {
--- a/interpreter/src/main/java/org/qbicc/interpreter/impl/VmImpl.java
+++ b/interpreter/src/main/java/org/qbicc/interpreter/impl/VmImpl.java
@@ -444,7 +444,7 @@ public final class VmImpl implements Vm {
             // VMHelpers
             VmClassImpl vmHelpersClass = bootstrapClassLoader.loadClass("org/qbicc/runtime/main/VMHelpers");
 
-            vmHelpersClass.registerInvokable("get_class", (thread, target, args) -> ((VmObjectImpl) args.get(0)).getVmClass());
+            vmHelpersClass.registerInvokable("getClass", (thread, target, args) -> ((VmObjectImpl) args.get(0)).getVmClass());
             vmHelpersClass.registerInvokable("classForName", (thread, target, args) -> {
                 VmClassLoaderImpl classLoader = (VmClassLoaderImpl) args.get(2);
                 if (classLoader == null) {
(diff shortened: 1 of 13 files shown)


runtime/main/src/main/java/org/qbicc/runtime/main/VMHelpers.java
```diff
@@ -632,4 +618,12 @@ public static void JLT_start0(void_ptr_unaryoperator_function_ptr runFuncPtr, vo
             throw new InternalError("pthread error code: " + result);
         }
     }
+
+    // TODO: Class library compatibility kludge; remove once qbicc 0.3.0 is released
```
dgrove-oss, line 622: Poking at this a bit more to see if I can avoid this or reduce it....not 100% sure why this is needed, but I needed at least some bits of it to get Thread.start to compile correctly.

dgrove-oss: we need to leave a forwarding method for threadWrapper. Reduced to minimal kludge and left a comment about what to update post qbicc-rt 0.3.0

interpreter/src/main/java/org/qbicc/interpreter/impl/VmImpl.java
```diff
@@ -444,7 +444,7 @@ public void initialize() {
             // VMHelpers
             VmClassImpl vmHelpersClass = bootstrapClassLoader.loadClass("org/qbicc/runtime/main/VMHelpers");
 
-            vmHelpersClass.registerInvokable("get_class", (thread, target, args) -> ((VmObjectImpl) args.get(0)).getVmClass());
+            vmHelpersClass.registerInvokable("getClass", (thread, target, args) -> ((VmObjectImpl) args.get(0)).getVmClass());
```
dmlloyd, line 447: I think this will likely conflict with `Object.getClass()` - if you provide the parameter count argument though, it should be enough to disambiguate.

dmlloyd (commented):
Otherwise looks good

interpreter/src/main/java/org/qbicc/interpreter/impl/VmImpl.java
dgrove-oss: Done.

plugins/intrinsics/src/main/java/org/qbicc/plugin/intrinsics/core/CoreIntrinsics.java
```diff
@@ -313,7 +313,7 @@ public static void registerJavaLangClassIntrinsics(CompilationContext ctxt) {
         intrinsics.registerIntrinsic(jlcDesc, "desiredAssertionStatus", emptyToBool, desiredAssertionStatus);
         intrinsics.registerIntrinsic(jlcDesc, "initClassName", emptyToString, initClassName);
         intrinsics.registerIntrinsic(jlcDesc, "getPrimitiveClass", stringToClass, getPrimitiveClass);
-        intrinsics.registerIntrinsic(Phase.LOWER, jlcDesc, "getSuperclass", emptyToClass, getSuperclass);
+        intrinsics.registerIntrinsic(Phase.LOWER, jlcDesc, "getSuperClass", emptyToClass, getSuperclass);
```
dmlloyd, line 316: This one is a JDK method and can't be renamed.  I guess it was a find/replace error.

plugins/lowering/src/main/java/org/qbicc/plugin/lowering/VMHelpersSetupHook.java
```diff
@@ -17,13 +17,13 @@ public void accept(final CompilationContext ctxt) {
...
-        ctxt.enqueue(methodFinder.getMethod("instanceof_class"));
-        ctxt.enqueue(methodFinder.getMethod("instanceof_typeId"));
-        ctxt.enqueue(methodFinder.getMethod("get_class"));
-        ctxt.enqueue(methodFinder.getMethod("classof_from_typeid"));
-        ctxt.enqueue(methodFinder.getMethod("get_superclass"));
+        ctxt.enqueue(methodFinder.getMethod("checkcastClass"));
+        ctxt.enqueue(methodFinder.getMethod("checkcastTypeId"));
+        ctxt.enqueue(methodFinder.getMethod("instanceofClass"));
+        ctxt.enqueue(methodFinder.getMethod("instanceofTypeId"));
+        ctxt.enqueue(methodFinder.getMethod("getClass"));
```
dmlloyd, line 24: This might return `Object.getClass`. Maybe we should just rename this one to `getClassForObject`?

dgrove-oss: yep.  good catch. fixed

plugins/intrinsics/src/main/java/org/qbicc/plugin/intrinsics/core/CoreIntrinsics.java
dgrove-oss: yep.  good catch.  fixed

dmlloyd (approved):

PR #913, opened 2021-12-06
Use Patcher instead of injectField

dmlloyd:
This replaces all but one usage of `injectField` with the patcher. The main difference here is that the injected fields will always be on the target class, instead of being added late, thus the interpreter doesn't need to have special hooks to access them.

### DIFF
--- a/plugins/core-classes/src/main/java/org/qbicc/plugin/coreclasses/CoreClasses.java
+++ b/plugins/core-classes/src/main/java/org/qbicc/plugin/coreclasses/CoreClasses.java
@@ -45,6 +45,8 @@ public final class CoreClasses {
     };
     private static final String OBJECT_INT_NAME = "java/lang/Object";
     private static final String CLASS_INT_NAME = "java/lang/Class";
+    private static final String THREAD_INT_NAME = "java/lang/Thread";
+    private static final String THROWABLE_INT_NAME = "java/lang/Throwable";
 
     private final CompilationContext ctxt;
 
@@ -54,6 +56,8 @@ public final class CoreClasses {
     private final FieldElement classTypeIdField;
     private final FieldElement classDimensionField;
     private final FieldElement arrayClassField;
+
+    private final FieldElement thrownField;
 
     private final FieldElement arrayLengthField;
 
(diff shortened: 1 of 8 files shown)


plugins/core-classes/src/main/java/org/qbicc/plugin/coreclasses/CoreClasses.java
```diff
@@ -41,6 +43,10 @@
         builder.setEnclosingType(enclosing);
         return builder.build();
     };
+    private static final String OBJECT_INT_NAME = "java/lang/Object";
```
dgrove-oss, line 46: Instead of making these private static final, should we have a central list of constants for the internal name of well-known classes?  We've got class names as strings all over the place...

plugins/core-classes/src/main/java/org/qbicc/plugin/coreclasses/CoreClasses.java
```diff
@@ -248,6 +208,100 @@ private static FieldElement makeContentField(final int index, final DefinedTypeD
         return fieldBuilder.build();
     }
 
+    /**
+     * Initialize all of our fields ahead of time using patcher.
+     *
+     * @param ctxt the compilation context (must not be {@code null})
+     */
+    public static void init(CompilationContext ctxt) {
```
dgrove-oss, line 216: Nice!  I hadn't thought about doing this programatically.  I was envisioning having to have a Object$_patch.java in the class library and dealing with all the complexity of keeping the class library and compiler in synch...

dgrove-oss (approved):

plugins/core-classes/src/main/java/org/qbicc/plugin/coreclasses/CoreClasses.java
dmlloyd: Yeah, that's probably a good idea, but I'll save that for a follow up because I'm weary of conflicts :)

plugins/core-classes/src/main/java/org/qbicc/plugin/coreclasses/CoreClasses.java
dmlloyd: The programmatic API was almost an afterthought TBH... but I think it might turn out to be very useful if we can fix the object layout *before* the interpreter runs.

PR #914, opened 2021-12-06
Build-time-complete layout

dmlloyd:
This PR introduces changes which:

* Make layout (nearly) 100% consistent between build and run time
* Opens up the `VmClass.registerInvokable` API
* Lowers static fields to structures again, in order to facilitate `Unsafe.staticFieldBase`
* Removes obsolete `Class.cast` intrinsic
* Add `Class.isInstance` support to the VM (not to run time though)
* Fix a couple of problems and missing features in reflective field and method enumeration
* Change support for `objectFieldOffset` and friends to yield an exact offset, but be build-time only - fields whose offsets are observed are marked as `I_ACC_PINNED`, to allow future `LOWER`-time field deletion and reshuffling

### DIFF
--- a/interpreter/src/main/java/org/qbicc/interpreter/impl/VmClassImpl.java
+++ b/interpreter/src/main/java/org/qbicc/interpreter/impl/VmClassImpl.java
@@ -109,8 +109,8 @@ class VmClassImpl extends VmObjectImpl implements VmClass {
         ClassContext classContext = typeDefinition.getContext();
         classLoader = (VmClassLoaderImpl) classContext.getClassLoader();
         CompilationContext ctxt = classContext.getCompilationContext();
-        layoutInfo = typeDefinition.isInterface() ? null : Layout.getForInterpreter(ctxt).getInstanceLayoutInfo(typeDefinition);
-        staticLayoutInfo = Layout.getForInterpreter(ctxt).getInterpreterStaticLayoutInfo(typeDefinition);
+        layoutInfo = typeDefinition.isInterface() ? null : Layout.get(ctxt).getInstanceLayoutInfo(typeDefinition);
+        staticLayoutInfo = Layout.get(ctxt).getInterpreterStaticLayoutInfo(typeDefinition);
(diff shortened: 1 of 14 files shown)


interpreter/src/main/java/org/qbicc/interpreter/impl/VmClassImpl.java
```diff
@@ -610,6 +632,29 @@ void registerInvokable(final ExecutableElement element, final VmInvokable invoka
...
+                registerInvokable(typeDefinition.getInitializer(), invokable);
+            } else {
+                throw new IllegalArgumentException("Invalid initializer descriptor " + descriptor + " on " + this);
+            }
+        }
+        if (name.equals("<init>")) {
+            int idx = typeDefinition.findConstructorIndex(descriptor);
+            if (idx == -1) {
+                throw new IllegalArgumentException("No constructor found with descriptor " + descriptor + " on " + this);
+            }
```
dgrove-oss, line 648:
```suggestion
            }
            registerInvokable(typeDefinition.getConstructor(idx), invokable);
```

interpreter/src/main/java/org/qbicc/interpreter/impl/VmClassImpl.java
```diff
@@ -610,6 +632,29 @@ void registerInvokable(final ExecutableElement element, final VmInvokable invoka
...
+        } else {
+            int idx = typeDefinition.findMethodIndex(name, descriptor);
+            if (idx == -1) {
+                throw new IllegalArgumentException("No method found with name \"" + name + "\" and descriptor " + descriptor + " on " + this);
+            }
+            registerInvokable(typeDefinition.getConstructor(idx), invokable);
```
dgrove-oss, line 654:
```suggestion
            registerInvokable(typeDefinition.getMethod(idx), invokable);
```

dgrove-oss (approved):
I think there's a likely cut & paste error in `registerInvokable` (see suggested changes), but otherwise makes sense

PR #915, opened 2021-12-06
reshape New to enable CompilerIntrinsics.emitNew

dgrove-oss:


dgrove-oss:
Nicely enabled by #914 🚀

compiler/src/main/java/org/qbicc/graph/Node.java
```diff
@@ -752,7 +752,7 @@ public Value visit(final Copier param, final Neg node) {
 
             public Value visit(final Copier param, final New node) {
                 param.copyNode(node.getDependency());
-                return param.getBlockBuilder().new_(node.getClassObjectType());
+                return param.getBlockBuilder().new_(node.getClassObjectType(), node.getTypeId(), node.getSize(), node.getAlign());
```
dmlloyd, line 755: These values should be copied in case they are dynamic e.g. `param.copyValue(node.getTypeId())` etc.

dmlloyd (commented):
Looks okay except for one minor issue.

compiler/src/main/java/org/qbicc/graph/Node.java
dgrove-oss: Thanks. fixed.

dmlloyd (approved):

PR #926, opened 2021-12-08
fix bug in main method finder

dgrove-oss:
Motivated by the test program below, where `qbicc` failed compilation because it was convinced that `printIt` was the main method.

```public class NewArray {
    static Object alloc(int size) {
        return Array.newInstance(String.class, size);
    }

    static void printIt(String[] s) {
        System.out.println(s[0]+" "+s[1]);
    }
    
        Object x = alloc(5);
        String [] y = (String[])x;
        y[0] = "Hello";
        y[1] = "World";
        printIt(y);
    }
}
```


dmlloyd (approved):

dmlloyd:
Good catch 😅

PR #928, opened 2021-12-09
change emitNew and emitNewReferenceArray to take Class

dgrove-oss:


plugins/intrinsics/src/main/java/org/qbicc/plugin/intrinsics/core/CoreIntrinsics.java
```diff
@@ -1067,22 +1067,25 @@ static void registerOrgQbiccCompilerIntrinsics(final CompilationContext ctxt) {
         CoreClasses coreClasses = CoreClasses.get(ctxt);
 
         ClassTypeDescriptor ciDesc = ClassTypeDescriptor.synthesize(classContext, "org/qbicc/runtime/main/CompilerIntrinsics");
-        ClassTypeDescriptor typeIdDesc = ClassTypeDescriptor.synthesize(classContext, "org/qbicc/runtime/CNative$type_id");
-        ClassTypeDescriptor uint8Desc = ClassTypeDescriptor.synthesize(classContext, "org/qbicc/runtime/stdc/Stdint$uint8_t");
         ClassTypeDescriptor objDesc = ClassTypeDescriptor.synthesize(classContext, "java/lang/Object");
         ClassTypeDescriptor clsDesc = ClassTypeDescriptor.synthesize(classContext, "java/lang/Class");
 
-        MethodDescriptor newRefArrayDesc =  MethodDescriptor.synthesize(classContext, objDesc, List.of(typeIdDesc, uint8Desc, BaseTypeDescriptor.I));
+        MethodDescriptor newRefArrayDesc =  MethodDescriptor.synthesize(classContext, objDesc, List.of(clsDesc, BaseTypeDescriptor.I, BaseTypeDescriptor.I));
```
DanHeidinga, line 1073: Why the change to take an `int` here?  Code looks correct for the change but not sure I follow the reason for it

dgrove-oss: It avoids a `CNative.word(dim)` in the class library.  We could go either way... the code snippet in Array.newArray is:
```
            int dimensions = 1;
            while (componentType.isArray()) {
                dimensions += 1;
                componentType = componentType.getComponentType();
            }
            if (dimensions > 255) {
                throw new IllegalArgumentException();
            }
            type_id typeId = CompilerIntrinsics.getTypeIdFromClass(componentType);
            return CompilerIntrinsics.emitNewReferenceArray(componetType, dimensions, length);
```
or we can take a u_int8t and write:
```
            return CompilerIntrinsics.emitNewReferenceArray(componentType, CNative.word(dimensions), length);
```

DanHeidinga: Avoiding the `CNative.word(dim)` makes sense - thanks for clarifying!

dmlloyd: Just want to add a quick note that `word()` does not introduce any run-time overhead; this is just for source brevity.

PR #930, opened 2021-12-09
Add reflection plugin with annotation and constant pool support

dmlloyd:
This makes the reflection run time believe that each class has only one constant pool, even though our classes are stitched together from multiple class files. One advantage is that we only add constants to our simulated constant pool when they're actually needed.

Introduces a new dependency on the Eclipse collections library.


dgrove-oss (approved):
LGTM

PR #943, opened 2021-12-17
remove Clone node and use class lib implementation

### LINKED ISSUE #674: Implement `Object.clone()`
One of the Big Two sources of compilation warnings right now is because `Object.clone()` is not fully implemented.

In order to implement this method, I propose something like this:

* Add a field to `Class` which will be populated with the run time size of the corresponding object instance
* Create a `Object$_native` implementation for `Object.clone()` whose implementation allocates an object of the size from the corresponding class, and uses the LLVM `memcpy` intrinsic to populate that object (be sure to exclude the object header so it doesn't get overwritten), ending with a release fence
* Create overrides of this method on each of the generated array core classes which adds the allocated array size to the size of the corresponding element type, but otherwise uses the same process as above to populate & publish the object
* Remove the `Clone` node and corresponding intrinsics completely from the codebase

The nodes needed to implement this method - specifically, reading an injected field from class and calling the `memcpy` intrinsic are not particularly easy to do on the Java side yet.  For reading the injected field, we could define an intrinsic which returns the size given the corresponding `Class` object.  For the `memcpy` intrinsic, I think we need to consider adding an LLVM runtime library module which defines a few intrinsic methods, starting with `memcpy`. Alternatively the standard C `memcpy` could be used.  The release fence is accomplished by calling `VarHandle.releaseFence()`.

Update: We do not presently have a good way to define native method bindings which apply to the array classes.  We'll need to add a (public) override of `clone` onto the base array class at a minimum, or else overrides on to each subclass, when the core classes are created.  The method bodies could be created by hand in `CoreClasses`, or, if someone wants to go through the trouble, the could ensure that e.g. binding `internal_array$_native` works correctly and implement the overrides there.

### LINKED ISSUE #744: Interpreter clone() should check if receiver implements Cloneable
This is good for bootstrapping but we'll need to eventually check that the cloned object implements `Cloneable` and otherwise throw the `CloneNotSupportedException`

_Originally posted by @DanHeidinga in https://github.com/qbicc/qbicc/pull/741#discussion_r735602141_

dgrove-oss:
Fixes #744.
Fixes #674.


dgrove-oss:
Needs various fixes to Array, Class, and Object to get into a classlib release before we can merge this.   Unfortunately, one of our CI test requires a partially-working clone to pass, so we can't break the existing clone support...

dmlloyd (approved):

Issue #950, opened 2021-12-20
Catalog initialization steps by OS


dmlloyd:
In order to work out a cogent JVM initialization and re-initialization process, we have to analyze the steps performed by OpenJDK today.

These steps must be assembled from multiple places. They could be organized into two broad categories: cross-platform and platform-specific.

The majority of the cross-platform steps can be found by reading `System.initPhase1`,`initPhase2`, and `initPhase3`, with the "glue" in between these steps being found in `openjdk/src/hotspot/share/runtime/thread.cpp`.  In between these steps, platform-specific initialization occurs.  This can be understood by reading the `Threads::create_vm` method in `thread.cpp`, and following the OS-specific calls (e.g. `os::init()`, `os::init_before_ergo()`, `os::pause()`, etc.).

These steps should be charted out into a table.  For each table row, the applicable OS(es) should be denoted, as well as at least a guess at whether the step (or an equivalent step) should be performed at build time, run time, or both.

dmlloyd:
Example format:

<table>
<tr><th>BT?<th>RT?<th>Description<th>Phase<th>Found in
<tr><td>Y<td><td>`setJavaLangAccess`<td>1<td>`System.initPhase1()`
</table>

dgrove-oss:
I've made an initial cut in the wiki: https://github.com/qbicc/qbicc/wiki/Qbicc-Initialization-Steps

dmlloyd:
Initialization is fully functional now.

PR #983, opened 2022-01-05
Add hook to allow objects to be modified during serialization.

### LINKED ISSUE #940: Object replacement on serialize/deserialize
We will soon need a solution for replacement of serialized heap objects.  I believe the instigating use case will be `FileDescriptor`.  We might be able to get by with a solution here which replaces all `FileDescriptor` instances with one single closed (i.e. `fd == -1`, `closed == true`, no `Closeable` values) instance.

dgrove-oss:
Fixes #940.

### DIFF
--- a/plugins/serialization/src/main/java/org/qbicc/plugin/serialization/InstanceInterceptor.java
+++ b/plugins/serialization/src/main/java/org/qbicc/plugin/serialization/InstanceInterceptor.java
@@ -1,0 +1,49 @@
+package org.qbicc.plugin.serialization;
+
+import static org.qbicc.graph.atomic.AccessModes.SingleUnshared;
+
+import org.qbicc.context.CompilationContext;
+import org.qbicc.interpreter.VmObject;
+import org.qbicc.type.definition.LoadedTypeDefinition;
+import org.qbicc.type.descriptor.ClassTypeDescriptor;
+
+/**
+ * This class enables VmObject instances to be modified
+ * immediately before they are serialized.
+ */
+class InstanceInterceptor {
+    private final CompilationContext ctxt;
+
+    InstanceInterceptor(CompilationContext ctxt) {
+        this.ctxt = ctxt;
+    }
+
+     VmObject process(VmObject original) {
+        ClassTypeDescriptor otd = original.getVmClass().getTypeDefinition().getDescriptor();
+        if (otd.packageAndClassNameEquals("java/io", "FileDescriptor")) {
+            return interceptFileDescriptor(original);
+        }
+
+        return original;
+    }
+
+
+    private VmObject interceptFileDescriptor(VmObject original) {
+        LoadedTypeDefinition fdType = original.getVmClass().getTypeDefinition();
+
+        // Do not change already closed descriptors and the special descriptors stdin, stdout, stderr.
+        if (original.getMemory().load32(original.indexOf(fdType.findField("fd")), SingleUnshared) < 3) {
+            return original;
+        }
+
+        // For all other descriptors, update their backing Memory as-if they have been closed.
+        original.getMemory().store32(original.indexOf(fdType.findField("fd")), -1, SingleUnshared);
+        original.getMemory().store64(original.indexOf(fdType.findField("handle")), -1, SingleUnshared);
+        original.getMemory().store8(original.indexOf(fdType.findField("closed")), 1, SingleUnshared);
+        original.getMemory().storeRef(original.indexOf(fdType.findField("cleanup")), null, SingleUnshared);
+        original.getMemory().storeRef(original.indexOf(fdType.findField("parent")), null, SingleUnshared);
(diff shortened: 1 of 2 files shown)


plugins/serialization/src/main/java/org/qbicc/plugin/serialization/InstanceInterceptor.java
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+
+    private VmObject interceptFileDescriptor(VmObject original) {
+        LoadedTypeDefinition fdType = original.getVmClass().getTypeDefinition();
+
+        // Do not change already closed descriptors and the special descriptors stdin, stdout, stderr.
+        if (original.getMemory().load32(original.indexOf(fdType.findField("fd")), SingleUnshared) < 3) {
+            return original;
+        }
```
dmlloyd, line 37: I don't think we can or should preserve the illusion that FDs 0..2 are preserved between build time and run time.  We will definitely need to re-open them at run time, probably using a run time initializer.  So I think it would be best to just close them all.

plugins/serialization/src/main/java/org/qbicc/plugin/serialization/InstanceInterceptor.java
```diff
@@ -0,0 +1,49 @@
...
+ * immediately before they are serialized.
+ */
+class InstanceInterceptor {
+    private final CompilationContext ctxt;
+
+    InstanceInterceptor(CompilationContext ctxt) {
+        this.ctxt = ctxt;
+    }
+
+     VmObject process(VmObject original) {
```
dgrove-oss, line 21: I originally designed this API to actually support a more general substitution of one object with another (which is why it returns a `VmObject`), but we don't need that yet and it is subtle to get right.   In particular, this code is running in `LOWER`, not in `ADD`, so we have to be very careful to not make additional methods or types reachable via interpretation of code paths we didn't execute while constructing the build time heap.

plugins/serialization/src/main/java/org/qbicc/plugin/serialization/InstanceInterceptor.java
```diff
@@ -0,0 +1,49 @@
...
+
+import org.qbicc.context.CompilationContext;
+import org.qbicc.interpreter.VmObject;
+import org.qbicc.type.definition.LoadedTypeDefinition;
+import org.qbicc.type.descriptor.ClassTypeDescriptor;
+
+/**
+ * This class enables VmObject instances to be modified
+ * immediately before they are serialized.
+ */
```
dmlloyd, line 13: What I had in mind was more along the lines of letting users declare some action in code, which we'd run via the interpreter somehow.  This does get us past the immediate issue, but I don't think we can expect to have compiler plugins handle every case in every library in the long term.

plugins/serialization/src/main/java/org/qbicc/plugin/serialization/InstanceInterceptor.java
dgrove-oss: I can try it, but I suspect that will actually break `System.out.println` until we get the runtime initialization working.

plugins/serialization/src/main/java/org/qbicc/plugin/serialization/InstanceInterceptor.java
dgrove-oss: It's pretty touchy to interpret code in LOWER.  In fact, I had originally written this code to do a `Vm.newInstance` using the default no-arg `FileDescriptor` constructor, but that didn't work.  (We happily tried to interpret it, but since it apparently hadn't been enqueued before, there was no schedule, and stepping through it in the debugger it happily interpreted the block-terminating `return` and never interpreted the store instructions that initialize `fd` and `handle` to `-1`).

plugins/serialization/src/main/java/org/qbicc/plugin/serialization/InstanceInterceptor.java
dmlloyd: Yeah I think that would be expected. We'd need a run time aspect for `FileDescriptor` and then a resolution for #950 which includes a run time init sequence that resets `System.in/out/err` (which should be reset to `null` by a run time aspect initializer for those fields, and which in turn should prevent those fields' values from being serialized in the first place).

plugins/serialization/src/main/java/org/qbicc/plugin/serialization/InstanceInterceptor.java
dmlloyd: Yeah it would have to all happen in ADD, somehow.  Maybe we'd allow an instance method to be called which returns the value of the field to use on serialize; we'd call that right after instantiating the object and then store the value away somewhere. Then after ADD, we'd substitute all the values.

dgrove-oss:
Closing as a dead end;  will work on more general framework instead.

PR #989, opened 2022-01-07
fixes in patcher and reachability for runtime initialized fields

dgrove-oss:


plugins/patcher/src/main/java/org/qbicc/plugin/patcher/PatchedTypeBuilder.java
```diff
@@ -83,14 +84,12 @@ public void addField(FieldResolver resolver, int index, String name, TypeDescrip
...
-                patchInfo = classPatchInfo.getReplacementFieldInfo(name, descriptor);
+                patchInfo = classPatchInfo.getModifiedFieldInfo(name, descriptor);
             }
             if (patchInfo == null || !ce.evaluateConditions(classContext, patchInfo, patchInfo.getAnnotation())) {
                 getDelegate().addField(resolver, index, name, descriptor);
-            } else if (patchInfo.getAdditionalModifiers() == 0) {
-                getDelegate().addField(patchInfo.getFieldResolver(), patchInfo.getIndex(), name, descriptor);
             } else {
-                getDelegate().addField(new PatcherFieldResolver(patchInfo), patchInfo.getIndex(), name, descriptor);
+                getDelegate().addField(new PatcherFieldResolver(patchInfo, resolver), patchInfo.getIndex(), name, descriptor);
```
dmlloyd, line 92: The original `index` is lost here. It should be passed in alongside the `resolver`. I think that we end up calling `resolver` with `patchInfo.getIndex()`, which, if my Monday brain is working correctly, I believe will give the index of the patched field with the resolver from the original, causing the wrong field to potentially be resolved.

dgrove-oss: Sigh.  You are correct.  I think it will be less fragile if represent a patch converting an existing static field to having a runtime initializer as a first-class concept instead of trying to jam it into the same patch that we use to add or replace an entire field.

dgrove-oss: Reworked to use a distinct patching operation to encode runtime field initializers.

dmlloyd (approved):

PR #993, opened 2022-01-11
progress on runtime support for <rtinit>

dgrove-oss:
* enhance reachability analysis for ObjectLiteral
  * generate rtinit check sequence
  * use Once instances to implement rtinit operation

Still not working: actually invoking the rtinit from the thunk
(unimplemented intrinsic).

### DIFF
--- a/runtime/main/src/main/java/org/qbicc/runtime/main/RuntimeInitializerRunner.java
+++ b/runtime/main/src/main/java/org/qbicc/runtime/main/RuntimeInitializerRunner.java
@@ -1,0 +1,28 @@
+package org.qbicc.runtime.main;
+
+/**
+ * Effectively a lambda that will execute a specific runtime initializer.
+ * <p>
+ * Runtime init checks are constructed by compile-time instantiation of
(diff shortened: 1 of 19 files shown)


runtime/main/src/main/java/org/qbicc/runtime/main/RuntimeInitializerRunner.java
```diff
@@ -0,0 +1,28 @@
...
+    }
+
+    @SuppressWarnings("unused") // invoked at buildtime
+    public static Once allocateThunk(int initID) {
+        return new Once(new RuntimeInitializerRunner(initID));
+    }
+
+    @Override
+    public void run() {
+        CompilerIntrinsics.callRuntimeInitializer(initID);
```
dgrove-oss, line 26: This intrinsic (and the gloabal[] of rtinit function pointers) is what isn't implemented yet.

dmlloyd (approved):

dmlloyd:
This all makes sense to me.

PR #995, opened 2022-01-12
Use Guice for Maven Resolver

### LINKED ISSUE #974: Use DI for Maven Resolver
Maven Resolver makes incompatible changes in minor or micro releases, because they assume that users of that library will be using a dependency injection (DI) library in order to wire it up, instead of the legacy & deprecated `ServiceLocator` class that we rely on.

Add Guice or something similar to the class path and use it to acquire the initial `RepositorySystem` in `QbiccMavenResolver` via `@Inject`, instead of using `QbiccServiceLocator` which has to be fixed almost every time `maven-resolver` is updated.

evacchi:
fixes #974

---

lots of trial and error with this one :) it required more explicit bindings than I expected.


dmlloyd (commented):
Having explicit bindings seems like we haven't gained anything. Have you looked at SISU? It uses class path scanning to locate bindings.

evacchi:
I did try this, but it didn't really work

```java
        this.system = Guice.createInjector(
                new WireModule(new SpaceModule(new URLClassSpace(QbiccMavenConfigurationModule.class.getClassLoader()))))
            .getInstance(RepositorySystem.class);
```

```
com.google.inject.ConfigurationException: Guice configuration errors:

1) [Guice/MissingImplementation]: No implementation for RepositorySystem was bound.

Learn more:
  https://github.com/google/guice/wiki/MISSING_IMPLEMENTATION

1 error

======================
Full classname legend:
======================
RepositorySystem: "org.eclipse.aether.RepositorySystem"
========================
End of classname legend:
========================
```

so it's not fully automated. I may try to delve deeper, but there is not a lot of docs around... there is an alternative approach (still broken)

https://github.com/snyk/aether-demo/blob/master/aether-demo-snippets/src/main/java/org/eclipse/aether/examples/sisu/SisuRepositorySystemFactory.java

```
        this.system = SisuRepositorySystemFactory.newRepositorySystem();
```

```
com.google.inject.CreationException: Unable to create injector, see the following errors:

1) An exception was caught and reported. Message: This Provider cannot be used until the Injector has been created.
  at [unknown source]
```

🤔

dmlloyd:
Interesting. SISU is supposed to scan the class path for files called `META-INF/sisu/javax.inject.Named`, and then read those values to find the `@Named` classes, and then wire up from there.  At least that's the theory of operation.  I don't know why it wouldn't be working in our case: we're not doing anything weird or fancy with the class path...

evacchi:
yeah I suspect that there is some magic incantation that I am not doing. The only example I could find is 6-years old, and the 5-minute introduction seems outdated as well.

evacchi:
I investigated further and found an up-to-date usage demo, so I lifted it *pretty much* verbatim (using wagon instead of plain http/file resolver) from [maven-resolver-demos/maven-resolver-demo-snippets](https://github.com/apache/maven-resolver/blob/1aba84cebf146c5dce9dc4209a0581b88d5a78e1/maven-resolver-demos/maven-resolver-demo-snippets/src/main/java/org/apache/maven/resolver/examples/guice/DemoResolverModule.java#L57)

the [SISU demo there](https://github.com/apache/maven-resolver/blob/1aba84cebf146c5dce9dc4209a0581b88d5a78e1/maven-resolver-demos/maven-resolver-demo-snippets/src/main/java/org/apache/maven/resolver/examples/sisu/SisuRepositorySystemFactory.java#L49) is actually broken and [fails to run](https://github.com/apache/maven-resolver/blob/95ad3ca4b6f2380999ca6b7e312f4f31aff45c83/maven-resolver-demos/maven-resolver-demo-snippets/src/test/java/org/apache/maven/resolver/examples/AllResolverDemosTest.java#L40-L43)

PR #1006, opened 2022-01-13
improve reachability analysis of runtime functions and dispatch tables

### LINKED ISSUE #1005: Reachability of runtime support methods and function pointers
During each compilation phase, qbicc starts with the assumption that some root set of methods (the entrypoints) could be invoked and then iteratively expands the set of reachable methods that must be compiled until it reaches a fixed point. 

We start fresh in each phase because we expect the reachable program to decrease as we apply optimizations.  The largest drop is expected to be between ADD and ANALYZE (due to build time initialization during ADD).

There are a couple of areas where we have some adhoc mechanisms that should be revisited and improved.
1. The compilation context has a `registerEntrypoint` API.  This is intended to be used to identify the methods (such as the application `main` method) that could be invoked externally. We are currently overloading this as a way of making all sorts of methods reachable. We should not do this (eventually entrypoints may need specialized calling conventions since they could be invoked externally). 
2. There is a collection of runtime support methods that become reachable during the `LOWER` phase as we expand high-level operations into lower-level ones.  We need to ensure that all of these methods that are needed are enqueued during the ADD and ANALYZE phases (when we do not have actual `Call` nodes in the graph that invoke them yet).  As the reachability analysis has become more precise, we need to do more involved registration of these methods (especially those that are instance methods).  
3. After the ANALYZE phase, we lower away the program-level constructs such as `New` and `InvokeVirtual` that we rely on during reachability analysis.  We also generate many dispatching tables with function pointers. We currently end up registering a very large number of methods as entrypoints during this transition.  It would be better to have a distinct mechanism for tracking executable elements that are reachable because we have injected them into dispatching tables.

dgrove-oss:
Fixes #1005.

### DIFF
--- a/compiler/src/main/java/org/qbicc/context/CompilationContext.java
+++ b/compiler/src/main/java/org/qbicc/context/CompilationContext.java
@@ -80,6 +80,20 @@ public interface CompilationContext extends DiagnosticContext {
      */
     void registerEntryPoint(ExecutableElement method);
 
+    /**
+     * AutoQueued elements are non-entrypoint executable elements that the
+     * compiler has determined must be included in the final image.
+     * Typically this is because they have been entered into a
+     * dispatching table or they are a runtime function that is guaranteed to
+     * be invoked by expansions in the LOWER phase.
+     * These capability should be used sparingly, since AutoQueued elements
+     * (and all elements and types that are indirectly reachable from them)
+     * will be included in the final executable.
+     *
+     * @param element the element to register as an auto-queued element.
+     */
+    void registerAutoQueuedElement(ExecutableElement element);
+
     Path getOutputDirectory();
 
     Path getOutputFile(DefinedTypeDefinition type, String suffix);
--- a/plugins/reachability/src/main/java/org/qbicc/plugin/reachability/ReachabilityInfo.java
+++ b/plugins/reachability/src/main/java/org/qbicc/plugin/reachability/ReachabilityInfo.java
@@ -11,6 +11,7 @@ import org.qbicc.context.AttachmentKey;
 import org.qbicc.context.CompilationContext;
 import org.qbicc.plugin.coreclasses.CoreClasses;
 import org.qbicc.type.definition.LoadedTypeDefinition;
+import org.qbicc.type.definition.element.ExecutableElement;
 import org.qbicc.type.definition.element.MethodElement;
 
 /**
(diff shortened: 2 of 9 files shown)


compiler/src/main/java/org/qbicc/context/CompilationContext.java
```diff
@@ -80,6 +80,25 @@ default ClassContext getClassContextForLoader(VmClassLoader classLoaderObject) {
...
+     * be included in the final image because they have either been
+     * entered into a dispatching table and thus may be invoked via
+     * an indirect call through a function pointer.
+     * These elements are not necessarily EntryPoints, but the program
+     * points that invoke them have been lowered to the point where it
+     * is no longer possible to determine they are reachable from the graph.
+     *
+     * @param element the element to register as invokable.
+     */
+    void registerInvokableElement(ExecutableElement element);
```
dmlloyd, line 94: I would maybe go with something more general here, e.g. `registerAutoQueuedElement` whose description is something like "Automatically enqueue the given element for all remaining phases", with an admonition "The element can never be eliminated so use sparingly" or something like that.

compiler/src/main/java/org/qbicc/context/CompilationContext.java
```diff
@@ -80,6 +80,25 @@ default ClassContext getClassContextForLoader(VmClassLoader classLoaderObject) {
...
+
+    default void registerRuntimeFunction(ExecutableElement element) {
+        registerInvokableElement(element);
+    }
```
dmlloyd, line 98: What is this one for?  Is there a possibility that this would have different semantics from the other one?

plugins/reachability/src/main/java/org/qbicc/plugin/reachability/ReachabilityInfo.java
```diff
@@ -133,10 +134,19 @@ public static void forceCoreClassesReachable(CompilationContext ctxt) {
...
+            info.analysis.processClassInitialization(ltd);
+            if (elem instanceof MethodElement me) {
+                if (me.isStatic()) {
+                    info.analysis.processReachableStaticInvoke(me, null);
+                } else {
+                    info.analysis.processReachableInstanceMethodInvoke(me, null);
+                }
+            }
+        });
+        LOGGER.debugf("Completed processing of runtime functions");
```
dmlloyd, line 149: With the above change, I would have the Driver do this work to pre-seed the queue for each phase.

dgrove-oss: That makes sense.  May need to add a method to ReachabilityInfo that the Driver can invoke, but that's cleaner than exposing a foreachRuntimeFunction on the CompilationContext.

dgrove-oss: Done.  Didn't know about ElementHandlers before.  Nice.

dgrove-oss:
revised according to comments and squashed into single commit.

dmlloyd (approved):

PR #1009, opened 2022-01-14
Initial implementation of mmap-based heap allocation

dmlloyd:
Some things are still unimplemented, but this code ensures that the heap is present when any entry point method is called, and allocates all new objects out of a single contiguous memory area, throwing OOME when no more space is available.

TODO:

- ~~More than one kind of pre-allocated OOME with a descriptive error~~ Going to skip this one for now
- [x] Defer errors to a "checkErrors" method which is run at any entry point before binding a thread (so that error messages can be properly printed)
- [x] Relay pre-configured values (such as the configured default minimum and maximum heap sizes and heap alignment, which are presently hard-coded)

### DIFF
--- a/runtime/gc/nogc/src/main/java/org/qbicc/runtime/gc/nogc/NoGcHelpers.java
+++ b/runtime/gc/nogc/src/main/java/org/qbicc/runtime/gc/nogc/NoGcHelpers.java
@@ -1,14 +1,12 @@
 package org.qbicc.runtime.gc.nogc;
 
 import static org.qbicc.runtime.CNative.*;
-import static org.qbicc.runtime.posix.Stdlib.*;
-import static org.qbicc.runtime.stdc.Stddef.*;
-import static org.qbicc.runtime.stdc.Stdlib.*;
+import static org.qbicc.runtime.stdc.Stdint.*;
 import static org.qbicc.runtime.stdc.String.*;
 
 import org.qbicc.runtime.AutoQueued;
-import org.qbicc.runtime.Build;
 import org.qbicc.runtime.Hidden;
+import org.qbicc.runtime.gc.heap.Heap;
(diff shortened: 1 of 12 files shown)


dmlloyd:
This is ready now.  This gives us `-Xms` and `-Xmx` support and bump-pointer allocation.  Still to do is to set the heap break (`sbrk`) beyond the heap and to line up the executable data sections such that they are contiguous with the heap to the maximum extent possible.

runtime/gc/nogc/src/main/java/org/qbicc/runtime/gc/nogc/NoGcHelpers.java
```diff
@@ -1,44 +1,55 @@
...
  */
 public final class NoGcHelpers {
     private NoGcHelpers() {}
 
+    // our allocation position
+    private static long pos;
+
     @Hidden
     @AutoQueued
     public static Object allocate(long size, int align) {
```
dgrove-oss, line 22: I was initially confused because if the current thread succeeds in calling `Heap.allocateRegion` but the `compareAndSetRelease` on `oldPos` fails we just restart the loop.  After reading more carefully, I convinced myself this is ok because the entire heap is being managed through a single bump pointer (`NoGc.pos`).  Therefore we didn't lose the region when we failed to commit the change from `oldPos` to `oldPos+size`....the heap is just growing linearly and when we try again that allocated region will be reflected in `Heap.getCurrentHeapOffset`, so it is not lost.

dgrove-oss (approved):

dgrove-oss:
Should we move `NoGcMultiNewArrayBasicBlockBuilder` to the gc-common plugin?  I don't think that expansion strategy is really gated on the GC strategy.

PR #1011, opened 2022-01-14
Many improvements and fixes for VarHandles

dmlloyd:
Includes changes to fix indexes of members, wrong types on sig-poly call sites, some missing method implementations relating to method handles, and moving a bunch more logic to the new reflection plugin


dgrove-oss (approved):

dgrove-oss (approved):
Read through it and the logic makes sense.  Have to admit I skimmed the 900ish lines of the new Reflection hub though...

PR #1027, opened 2022-01-20
Annotations for serialization replacement

dgrove-oss:
Cover two basic cases of replacing a field with a zero-initializer and replacing an integral field with a constant value (eg -1 for FileDescriptor.fd).

Not that attached to the actual names of the annotations, so happy to adjust if there are suggestions.

Thought about supporting Float, Double, and Boolean constants as well but don't know if we have a motivating example yet.

### DIFF
--- a/compiler/src/main/java/org/qbicc/type/definition/element/FieldElement.java
+++ b/compiler/src/main/java/org/qbicc/type/definition/element/FieldElement.java
@@ -3,8 +3,16 @@ package org.qbicc.type.definition.element;
 import java.util.function.Function;
 
 import org.qbicc.context.ClassContext;
+import org.qbicc.context.CompilationContext;
 import org.qbicc.graph.literal.Literal;
+import org.qbicc.type.BooleanType;
+import org.qbicc.type.FloatType;
+import org.qbicc.type.IntegerType;
 import org.qbicc.type.ValueType;
+import org.qbicc.type.annotation.Annotation;
+import org.qbicc.type.annotation.BooleanAnnotationValue;
+import org.qbicc.type.annotation.DoubleAnnotationValue;
+import org.qbicc.type.annotation.LongAnnotationValue;
 import org.qbicc.type.definition.classfile.ClassFile;
 import org.qbicc.type.descriptor.ClassTypeDescriptor;
 import org.qbicc.type.descriptor.TypeDescriptor;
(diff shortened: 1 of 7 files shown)


runtime/api/src/main/java/org/qbicc/runtime/SerializedAsPrimitiveConstant.java
```diff
@@ -0,0 +1,22 @@
...
+
+/**
+ * An annotation to indicate that when an instance or static field is
+ * serialized to the qbicc initial heap that whatever value was stored
+ * in the field in the build time heap should be replaced by the
+ * zero-initializer appropriate to the type of the field in the runtime heap.
+ */
+@Retention(RetentionPolicy.CLASS)
+@Target(ElementType.FIELD)
+public @interface SerializedAsPrimitiveConstant {
```
dmlloyd, line 16: Maybe `@SerializedAsLong`?  Or maybe just a `@SerializedAs` with different properties `longValue`, `intValue`, etc. where we allow only exactly one to be specified?

runtime/api/src/main/java/org/qbicc/runtime/SerializedAsPrimitiveConstant.java
```diff
@@ -0,0 +1,22 @@
...
+    /**
+     * The bits in the long will be cast to the appropriate primitive type
+     * and used as the replacement value for the field.
+     */
+    long bits();
```
dmlloyd, line 21: If this is the only property for the annotation, it should be called `Value` so you can have `@SerializedAsWhatever(10L)`.

dgrove-oss: Went with a single property `value` of type `long` for now.

dgrove-oss:
Revised based on comments and squashed/force-pushed.

runtime/api/src/main/java/org/qbicc/runtime/SerializedAs.java
```diff
@@ -0,0 +1,20 @@
...
+ * in the field in the build time heap should be replaced by the
+ * value provided by the annotation in the initial runtime heap.
+ * The replacement value will be that supplied by the value method
+ * that matches the type of the field.
+ */
+@Retention(RetentionPolicy.CLASS)
+@Target(ElementType.FIELD)
+public @interface SerializedAs {
+    long value();
+}
```
dmlloyd, line 20: Not trying to be difficult (honest!)... but this should either be `SerializedAsLong` with a property named `value` (which we may eventually supplement with additional annotations like `SerializedAsString` or whatever) or `SerializedAs` with a property named `longValue` (which we may eventually supplement with defaulted properties named `intValue`, `stringValue`, etc.`). I think the former option (additional annotations in the future) is probably a bit more laborious but it might give us a bit more user-friendly flexibility (for example we could eventually have an annotation that gives a factory class and method name which are called to produce the value).

dgrove-oss: No worries... the current implementation is probably overly cute and I didn't write the comment to explain it.  Worth the time iterating on this, because it is going to be part of the user experience. 

The current single annotation takes the 64 bit long `value` and based on the type of the annotated field interprets those bits as a constant of that type.  I think this is pretty sensible for boolean, byte, char, short, int, and long.  It has a well-defined meaning for float and double, but is of questionable usability (as you can't actually do `Double.toLongBits(1.0)` for example in an annotation value).  It doesn't support String (or Class).

If we don't mind having 9 separate annotations, it wouldn't be that much work to just define them for each primitive type + String (`SerializeAsInt`, `SerializeAsChar`, etc).  I think that is probably more usable than a single annotation with 9 properties.

We could also do an intermediate design where we had four SerializeAs annotations: Boolean, FloatingPoint, Integral, and String.

dgrove-oss: Revised to separate Boolean, Integral, and FloatingPoint types.  Held off on String because it seemed like that would be better to handle under a general replacement mechanism for Objects (which I don't know how to do yet or if we will really need it) as opposed to a special case.

dmlloyd (approved):

PR #1033, opened 2022-01-21
Optimize away redundant field zeroing

dmlloyd:
This one I'm not quite 100% sure about. We might want to go back and add a check to ensure that nothing earlier has written a non-zero value to the field.


dgrove-oss:
Yeah, this seems a bit dodgy to me without some kind of analysis to detect intervening stores (including those that may have happened if `this` escaped to a method outside of the constructor).  Maybe if we walk back the dependency chain and we can get all the way to the initializing `memclear` operation it is ok to do this?

dmlloyd:
Yeah that'd likely be better than trusting Java `final` rules.

PR #1037, opened 2022-01-21
move C main function to class lib

dgrove-oss:
Relocate main function to qbicc-classlib project to make it more convenient to interleave runtime and JDK initialization.

Depends on https://github.com/qbicc/qbicc-class-library/pull/95


dgrove-oss:
This PR is passing CI, but is reliably failing with hundreds of errors in the interpreter for me locally.  Something funky is going on...

dgrove-oss:
Turned out to be something simple...`sigaction.sa_restorer` isn't really part of Posix (and is not defined on MacOS).  Switching the main method to use the one from the class lib means that we actually compile rtinitPhase1, which since earlier today means we pull in constants from posix.Signals into Signal$_native and the C compilation of sigaction was failing on macOS.

dmlloyd (approved):

PR #1045, opened 2022-01-25
`String.intern()` fixes and improvements

dmlloyd:


dgrove-oss (approved):
Maybe depends on the class lib change to implement String.intern?

dmlloyd:
Yes and no: nothing at run time (that we test) is calling `String.intern()` yet so it's basically harmless to revert this to a ULE - at least until qbicc/qbicc-class-library#97 is merged and released.

PR #1046, opened 2022-01-25
Fix isAssignableFrom invokable for primitives

dmlloyd:


plugins/intrinsics/src/main/java/org/qbicc/plugin/intrinsics/core/CoreIntrinsics.java
```diff
@@ -232,10 +232,10 @@ public static void registerJavaLangClassIntrinsics(CompilationContext ctxt) {
 
         // todo: this probably belongs in the class libraries rather than here
         StaticIntrinsic desiredAssertionStatus0 = (builder, target, arguments) ->
-            classContext.getLiteralFactory().literalOf(false);
+            classContext.getLiteralFactory().literalOf(true);
 
         InstanceIntrinsic desiredAssertionStatus =  (builder, instance, target, arguments) ->
-            classContext.getLiteralFactory().literalOf(false);
+            classContext.getLiteralFactory().literalOf(true);
```
dmlloyd, line 238: I didn't mean to include this change here. But I'm happy to keep it for now.

dgrove-oss: Will be "interesting" to see if it causes any new bugs to surface in CI.

dgrove-oss (approved):

PR #1050, opened 2022-01-26
disable JIT candidate message

dgrove-oss:
Cuts down some of the noise (flags 180 candidates instead of 680).


interpreter/src/main/java/org/qbicc/interpreter/impl/VmInvokableImpl.java
```diff
@@ -132,7 +132,7 @@ public Object invokeAny(VmThread thread, VmObject target, List<Object> args) {
 
     Object run(VmThreadImpl thread, VmObject target, List<Object> args) {
         long invCnt = ((long) countHandle.getAndAdd(this, 1)) + 1;
-        if (invCnt == 100) {
+        if (invCnt == 1000) {
```
dmlloyd, line 135: Honestly you could convert this to an `if (false)`. It has taught us what we needed to know: we need a JIT. :-)

dgrove-oss: Done...disabled it.

dmlloyd (approved):

PR #1052, opened 2022-01-26
todo to remove Signal intrinsics

dgrove-oss:


dgrove-oss:
Vacillated between a PR with the TODOs and a PR that removed them but couldn't be merged until after the next classlib release. Could go either way...

dmlloyd (approved):

PR #1053, opened 2022-01-27
Use system ThreadGroup from buildtime heap to initialize main thread

dgrove-oss:


dgrove-oss:
The test failures are coming from the main thread group having a Thread instance in it (we refuse to serialize the pthread_mutex field).  Once we have a class lib release, so we are using https://github.com/qbicc/qbicc-class-library/pull/100, this problem will be avoided.

dmlloyd (approved):

PR #1054, opened 2022-01-27
teach Reachability that Thread is always instantiated/initialized

dgrove-oss:


dmlloyd (approved):
Makes sense. I think I want to make this mechanism a bit more user-spacey though, something like this:

```java
Thread main = new Thread("main" ...);
attachAndStart(main);
try {
    ...do stuff..
} finally {
    detach(main);
}
terminate(main); // also terminates the program if it's the last thread...
```

But, that can come later...

dgrove-oss:
Makes sense.  We can use the patcher to add a Thread constructor that does the a cutdown initialization that maybe can fold in the addition of the thread to the system thread group that is done in rtinitPhase1.  Would be nice to move as much of this as possible to Java.

PR #1055, opened 2022-01-27
Escape Analysis v2

galderz:
This PR replaces #932, to which the following changes have been applied:

1. If the compiler graph contains node types not yet handled by EA, any new allocations in the data flow graph from the unhandled nodes have their escape value switched to global escape. As mentioned in #932, this is done to err on the side of safety. This should handle all feedback regarding unhandled graph node types.
2. `DotNodeVisitor` can now be hooked with delegating node visitors to further decorate nodes. EA uses this mechanism to colour `New` nodes according to their escape value. Also, given EA's two-phase nature (intra method analysis and inter method analysis), I've enhanced the dot generator to produce two dot files for the analysis phase: `analysis-intra.dot` and `analysis-inter.dot`. This way, we can inspect differences, if any, in escape values between the two phases.
3. New allocations that are assigned to object instance fields via `this` are assumed to escape as arguments. Before this change, allocations that were assigned to instance fields in constructors were being assumed to no escape, hence they would be stack allocated. This obviously led to seg fault (null pointer) when trying to use the instance fields.
3. I've incorporated javadoc comments to different data structures based on the feedback from last PR.

Please see the [Escape Analysis design discussion](https://github.com/qbicc/qbicc/discussions/931) for an overview of the implementation.

I've been using [this Escape Analysis sample project](https://github.com/galderz/qoccido/tree/main/ea-samples) to verify things work as expected. It uses byteman to [intercept stack allocations](https://github.com/galderz/qoccido/blob/main/ea-samples/src/main/resources/ea/verify.btm#L29-L47) and compare them with [expectations](https://github.com/galderz/qoccido/blob/main/ea-samples/src/main/resources/ea/verify.btm#L8-L24). A summary of the samples can be found [here](https://github.com/galderz/qoccido/tree/main/ea-samples#overview). 

With the changes to dot node visitor, I will likely deprecate and remove connection graph dot visitor and related classes in the future. I will decide on that once I've checked whether I can represent further EA supporting metadata in the existing dot files. Related this, one thing I'm considering is whether I need to track all the metadata arounds points-to and deferred edges via the different BBB graph node implementations. Now that the code walks the graph exhaustively upon finish to discovered unhandled nodes, I could maybe calculate this info there and then. Maybe I don't need a BBB at all and a visitor is enough. I'll explore all of this once the first EA version is in main.

### DIFF
--- a/plugins/optimization/src/main/java/org/qbicc/plugin/opt/ea/EscapeAnalysisState.java
+++ b/plugins/optimization/src/main/java/org/qbicc/plugin/opt/ea/EscapeAnalysisState.java
@@ -1,0 +1,69 @@
+package org.qbicc.plugin.opt.ea;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
+import org.qbicc.context.AttachmentKey;
+import org.qbicc.context.CompilationContext;
+import org.qbicc.context.PhaseAttachmentKey;
+import org.qbicc.graph.Call;
+import org.qbicc.graph.New;
+import org.qbicc.type.definition.element.ExecutableElement;
+
+final class EscapeAnalysisState {
+    private static final PhaseAttachmentKey<EscapeAnalysisState> KEY = new PhaseAttachmentKey<>();
+
+    private final Map<ExecutableElement, List<Call>> callGraph = new ConcurrentHashMap<>();
+    private final Map<ExecutableElement, ConnectionGraph> connectionGraphs = new ConcurrentHashMap<>();
+
+    ConnectionGraph getConnectionGraph(ExecutableElement element) {
+        return connectionGraphs.get(element);
+    }
+
+    // TODO Collection<MethodElement> instead?
+    Collection<ExecutableElement> getMethodsVisited() {
+        return connectionGraphs.keySet();
+    }
+
+    /**
+     * Returns the list of callees called from the given method.
+     * If method not found in the call graph, an empty list is returned.
+     */
+    List<Call> getCallees(ExecutableElement element) {
+        final List<Call> callees = callGraph.get(element);
+        return callees != null ? callees : Collections.emptyList();
+    }
+
+    void trackMethod(ExecutableElement element, ConnectionGraph connectionGraph) {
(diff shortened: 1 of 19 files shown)


galderz:
The algorithm used to search through the data flow graph for unhandled node types is based on the phi populator, but has an important difference: there could be multiple data paths to a `New` node, so if any of those paths contains an unsupported noded, we skip the fact its dependencies might already have been visited by some other path, if they've been marked as supported. Then, we continue all the way to end of the graph propagating the unsupported status. This way, if a `New` node has a data dependency path with an unhandled node, no matter what the order in which those paths are visited, the `New` node will end up being pessimistically being set to global escape.

plugins/optimization/src/main/java/org/qbicc/plugin/opt/ea/EscapeAnalysisState.java
```diff
@@ -0,0 +1,63 @@
...
+import java.util.concurrent.ConcurrentHashMap;
+
+import org.qbicc.context.AttachmentKey;
+import org.qbicc.context.CompilationContext;
+import org.qbicc.graph.Call;
+import org.qbicc.graph.New;
+import org.qbicc.type.definition.element.ExecutableElement;
+
+final class EscapeAnalysisState {
+    private static final AttachmentKey<EscapeAnalysisState> KEY = new AttachmentKey<>();
```
dmlloyd, line 17: I think we really want this to be a `PhaseAttachmentKey` so that the state is updated on every phase.  As the program gets smaller, the environment around values which previously escaped might be transformed so that the values no longer escape.  In general we should be ensuring that *every analysis* is repeatable between phases.

This would allow us, for example, to use the escape information from the *previous* phase within peephole optimizations in an unrelated BBB.

dmlloyd (commented):
This is shaping up really nicely @galderz! I only have one comment.

plugins/optimization/src/main/java/org/qbicc/plugin/opt/ea/EscapeAnalysisState.java
galderz: I've switched that now. Since `EscapeAnalysisOptimizeVisitor` runs on `LOWER`, it now asks for the previews phase escape state and uses that.

dmlloyd (approved):

PR #1059, opened 2022-01-27
Introduce Facts API, track field reachability

dmlloyd:
This is the ~~first~~ ~~second~~ final cut at tracking global field reachability.

* Tracks reads and writes of fields independently
* **New:** Introduce and use a prototype facts API as per #1109 
* New field reachability info for every phase, with the previous phase result available, by way of facts API
* ~~A first pass at deleting accesses of dead fields~~ Dead field removal omitted; this belongs in an optimization BBB after we've accounted for such things as on-heap objects, constant values, and fields whose usages are introduced in later phases
* Demonstration of how per-phase attachments are meant to be used, by way of facts API

### DIFF
--- a/compiler/src/main/java/org/qbicc/context/CompilationContext.java
+++ b/compiler/src/main/java/org/qbicc/context/CompilationContext.java
@@ -6,6 +6,8 @@ import java.util.function.BiConsumer;
 import java.util.function.BiFunction;
 import java.util.function.Consumer;
 
+import org.qbicc.facts.Facts;
+import org.qbicc.facts.core.ExecutableReachabilityFacts;
 import org.qbicc.graph.BasicBlock;
 import org.qbicc.graph.Node;
 import org.qbicc.graph.NodeVisitor;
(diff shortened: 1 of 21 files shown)


dgrove-oss:
> * Fields which are only accessed within intrinsics which register in `ANALYZE` or later phases get deleted

One possibility would be when we are defining the intrinsics, we could register the FieldElements that are being accessed.  Then we could seed the analysis with this read/write set.   It would be overly conservative (since the simple implementation would just assume that every intrinsic was reachable).  We'd also have to do something additional for any "reflectish" intrinsics like the Unsafe accessors where the accessed field isn't fixed.  Maybe these can all be expanded in ADD (so they will be visible to the analysis).

dmlloyd:
These intrinsics are one of the few things that break the "rule" of only adding "new" code during `ADD`.  Expanding them early might be a good option - it might also obviate the need for manual/auto-queueing in a few places.  Off the top of my head though I can imagine it being a problem for cases where we have e.g. constants whose value won't be known until `LOWER` - in cases like this we are using intrinsics to defer them, but maybe we should have another approach for this.  A deferred constant node perhaps.

dmlloyd:
I think this is ready to go in. The facts API itself is functional, and we can flesh out the individual facts and/or switch over to them as a later step.

Having this in would mean that we can execute arbitrary tasks in the main task queue, which I think will be useful.

dgrove-oss (approved):

PR #1060, opened 2022-01-27
Add some more Unsafe intrinsics

dmlloyd:
This should solve ULEs relating to `Unsafe#getShort` and a few others.


dgrove-oss (approved):
LGTM

PR #1076, opened 2022-01-29
implement the @Annotate patcher annotation

dgrove-oss:


dmlloyd (approved):
I definitely remembered that I had created an `@Annotate` annotation! 😬

Looks good.

PR #1079, opened 2022-01-31
Add ability to emit source files for debugging

dmlloyd:
This PR gathers any resolved source JARs from the Maven resolution path and creates an output directory which contains all of the reachable source files.  You would then point the debugger to that directory with a single command, e.g. `(gdb) directory target/src`, and then you'd have complete source available.

The one problem is that we are not presently fetching source JARs during Maven artifact resolution, so I'm putting this into draft until there is occasion to fix that.

A future enhancement would be to also modify the non-Maven class path item specification function to support source directories, perhaps using some kind of magical syntax e.g. an `--app-path-source-file` which would use classpath syntax to add source paths for the application class path.

Another (much later) future enhancement would be to add a facility so that class loaders which are created within the application during build time have some way to register their corresponding source directories.

### DIFF
--- a/compiler/src/main/java/org/qbicc/interpreter/Vm.java
+++ b/compiler/src/main/java/org/qbicc/interpreter/Vm.java
@@ -387,4 +387,18 @@ public interface Vm {
      * @return the memory (not {@code null})
      */
     Memory getGlobal(GlobalVariableElement variableElement);
+
+    /**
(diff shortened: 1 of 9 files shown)


dmlloyd:
We are now fetching source JARs so I was able to get this working!

dgrove-oss (approved):
LGTM

PR #1095, opened 2022-02-03
Further method handle work

dmlloyd:
Method handles are still dispatching on null pointers, but this brings us closer. Almost all method handles are being produced now, with only a couple of exceptions.


dgrove-oss (approved):
Nice!

dgrove-oss:
A missing `git add` on new files?

PR #1099, opened 2022-02-04
pass method/field name to NoSuch{Method,Field}Error constructor

dgrove-oss:


dgrove-oss:
We could easily add the descriptor to the error string as well.   Not sure if that is desirable or not..

dmlloyd (approved):

PR #1101, opened 2022-02-07
Specify pointee type on pointer accesses

dmlloyd:
At present, casting from one pointer type of type `ptr<X>` to another is essentially meaningless. Too much generic type information is lost, and while it might be barely possible in some cases to infer the type of certain casts, in most cases it seems like this is not going to be possible.  At present we work around this problem by defining reified subclasses of every conceivably useful pointer type. But this measure was intended to be temporary, and it imposes an unreasonable burden especially as we ramp up JDK native implementation, not only of creating the pointer classes themselves but also in the extra casting required to move values between what should be compatible types (e.g. `free(ptr.cast())` and so on).  In some cases Java forbids what should be a valid cast so the target has to be cast through an intermediate or raw type (e.g. adding const-ness to a pointee).

This change proposes an alternative, whereby the type of the pointer can be determined, changed, or affirmed at the use site of the pointer. This approach is similar to how LLVM opaque pointers work.  LLVM assumes that all memory is untyped, so any interactions with memory must specify the pointee type at the use site.  Opaque pointers formalize this fact and will become the default in a near-future LLVM version.

The generic type of the pointer `ptr<X>` type would still be retained, which can still add warnings at the Java level in some cases if something is obviously amiss, and also serves to document the pointer's intent.

This approach is similar in some ways to the `castPtr` approach previously tried. However, that approach still relied on re-casting the result type to the desired type since parameterized pointer types would still cast incorrectly.

If this approach works out, it can be applied to the other memory access methods (stores, atomics), pointer arithmetic operations, the array-type view, and the `sel()` operator.

The downside of this approach is that the pointee type itself cannot really be generic. In order to load a value from a `ptr<?> ptr` into an `int`where the pointer points to an int pointer, one would have to do: `ptr.loadPlain(ptr.class).loadPlain(int32_t.class).intValue()`.  The types used in the call chain in this expression would have to be manually checked.  On the other hand, this can be done as a one-liner with 100% confidence in the type inferencing involved regardless of the generic type or erasure status of the source pointer.

For now this is in draft as an RFC, implemented just on loads to start with, until I can try it out in a few different situations and maybe fine-tune things a little. Also the porting effort may be nontrivial due to the number of methods we already have specified using the specific pointer types.  I would love to hear comments on this approach.


dgrove-oss:
This seems plausible, in that it matches where LLVM is going and makes things explicit and much less magical (albeit maybe more verbose than currently). 

I agree that the explosion of reified subclasses isn't going to be viable.

PR #1110, opened 2022-02-10
optimize loads of reallyFinal fields from ObjectLiterals

### LINKED ISSUE #1096: Constant load optimization
Any loads from an object literal's instance field which is `final` (*really* `final`) can be replaced with a literal of the field's value.  This could result in the elimination of boot time objects. Also, it could help with inlining of method handles.

dgrove-oss:
Fixes #1096.


dgrove-oss:
Optimizes 315 instance field loads when compiling HelloWorld.  Of these 299 are loads of the length field of arrays, 12 are loads of the elementType field of refArray, and the remaining 3 are true instance fields:
```
info: Replacing getfield of java/nio/charset.CoderResult$Cache.malformed in java/nio/charset.CoderResult.malformedForLength(I)Ljava/nio/charset/CoderResult;
info: Replacing getfield of java/nio/charset.CoderResult$Cache.unmappable in java/nio/charset.CoderResult.unmappableForLength(I)Ljava/nio/charset/CoderResult;
info: Replacing getfield of java/math.BigInteger.mag in java/math.BigInteger(Ljava/lang/String;I)V
```

dmlloyd (approved):

PR #1117, opened 2022-02-12
Introduce host-side pointer value classes

dmlloyd:


dmlloyd:
Had to fix up a couple of things. Added static field support. Removed `NullPointer` (we can just use `null` like we do for objects), and added pointer caching for static fields and methods. Added support for pointers in `Memory`.

dgrove-oss:
Generally makes sense and looks good, but I'm not seeing how `IntegerAsPointer` is going to work.  Once we lose track of the actual `Memory` object associated with the pointer, I think we're in trouble.

dmlloyd:
> Generally makes sense and looks good, but I'm not seeing how `IntegerAsPointer` is going to work. Once we lose track of the actual `Memory` object associated with the pointer, I think we're in trouble.

This would be the case where we have a `long` field which we've turned into a pointer, yet Java code is still storing a `long` value there sometimes, for whatever reason.  We obviously wouldn't be able to do anything with such a pointer during build (if a user tried to dereference it, we'd have to throw a `NullPointerException` probably, or something along those lines), but we still need to store the value into the same slot so that when it is read out again (as a `long`), it has a value.  If it comes to pass that we would serialize a heap with that value in it, it's no problem to serialize an `intToPtr` of that value (at least, no problem for *us* - if the user foolishly dereferenced such a value using `Unsafe`, a segfault would be the *best* case outcome).

PR #1121, opened 2022-02-14
Improvements to interpreter array-backed memory

dmlloyd:
This lays more (indirect) groundwork for handling pointers in the interpreter and also typed memory. It also eases debugging when you can directly access the array backing a `VmArray` and inspect it in the debugger.


dgrove-oss (approved):
Nice improvement: simple, less-error prone, and should be faster too

PR #1125, opened 2022-02-15
fixes for allocateMemory0

dgrove-oss:


interpreter/src/main/java/org/qbicc/interpreter/impl/Frame.java
```diff
@@ -919,6 +919,11 @@ public Object visit(VmThreadImpl thread, IsNe node) {
         } else if (isFloat64(inputType)) {
             return Boolean.valueOf(unboxDouble(left) != unboxDouble(right));
         } else if (isInt64(inputType)) {
+            // Allow "null check" idiom of comparing a MemoryPointer to 0
+            Object leftRaw = require(left);
+            if (leftRaw instanceof MemoryPointer && unboxLong(right) == 0) {
+                return Boolean.valueOf(true);
+            }
```
dgrove-oss, line 926: This is a very specific idiom used to null check native pointers whose nominal Java type is long. If we end up needing to add more than a couple specific hacks like this, we should probably figure out a more general solution.

dmlloyd (approved):

PR #1139, opened 2022-02-17
interpreter support for unsafe memcpy

dgrove-oss:
Enough support for native memory idioms to allow us to interpret `NativeBuffers.copyCStringToNativeBuffer` successfully.

### DIFF
--- a/interpreter/src/main/java/org/qbicc/interpreter/impl/Frame.java
+++ b/interpreter/src/main/java/org/qbicc/interpreter/impl/Frame.java
@@ -314,8 +314,14 @@ final strictfp class Frame implements ActionVisitor<VmThreadImpl, Void>, ValueVi
         ValueType inputType = node.getLeftInput().getType();
         assertSameTypes(node);
         if (isInt64(inputType)) {
-            // long math
-            return box(unboxLong(node.getLeftInput()) + unboxLong(node.getRightInput()), node.getType());
+            Object raw = require(node.getLeftInput());
+            if (raw instanceof Pointer p) {
+                // pointer arithmetic
+                return p.offsetInBytes(unboxLong(node.getRightInput()), true);
+            } else {
+                // long math
+                return box(unboxLong(node.getLeftInput()) + unboxLong(node.getRightInput()), node.getType());
+            }
         } else if (isInteger(inputType)) {
             // truncated integer math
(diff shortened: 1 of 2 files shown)


interpreter/src/main/java/org/qbicc/interpreter/impl/Frame.java
```diff
@@ -314,8 +315,14 @@ public Object visit(VmThreadImpl thread, Add node) {
         ValueType inputType = node.getLeftInput().getType();
         assertSameTypes(node);
         if (isInt64(inputType)) {
-            // long math
-            return box(unboxLong(node.getLeftInput()) + unboxLong(node.getRightInput()), node.getType());
+            Object raw = require(node.getLeftInput());
+            if (raw instanceof MemoryPointer mp) {
+                // pointer arithmetic
+                return new MemoryPointer(mp.getType(), MemoryFactory.offset(mp.getMemory(), unboxLong(node.getRightInput())));
```
dmlloyd, line 321: I think this doesn't need to be special-cased for `MemoryPointer` - given `Pointer p` you should be able to do `p.offsetInBytes(offset, true)`.

dmlloyd (commented):
A couple points of feedback, I agree with the general idea though.

interpreter/src/main/java/org/qbicc/interpreter/impl/Frame.java
dgrove-oss: ok, that made this lot simpler of a patch :) 

Revised and squashed to a single commit since the digression of a specialized backing memory isn't needed.

interpreter/src/main/java/org/qbicc/interpreter/impl/VmImpl.java
```diff
@@ -846,6 +856,45 @@ public void initialize() {
...
+            Memory srcMem = mp.getMemory();
+            Memory dstMem = db.getMemory();
+            for (long i = 0; i<bytes; i++) {
+                dstMem.store8(dOff.longValue() + i, srcMem.load8(0 + i, SinglePlain), SinglePlain);
+            }
+            return;
+        }
+
+        // copy native memory to native memory
+        if (srcBase == null && srcOffset instanceof MemoryPointer src && dstBase == null && dstOffset instanceof MemoryPointer dst) {
```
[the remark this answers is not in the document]
dgrove-oss: done

dmlloyd (approved):

PR #1147, opened 2022-02-25
Allow patching of method bodies

dmlloyd:


dgrove-oss (approved):
Seems locally fine, but not sure whether this is intended as a replacement for the @Replace annotation in methods, or if it is for a compiler-internal use case related to method handles where we are generating a new method body we want to inject, but don't want to bother generating the rest of the method element.

dmlloyd:
It's meant as a complement, when we want to replace the method body without changing any of the method's other attributes like its modifiers, parameter names, or annotations.  We might want to consider supplementing the annotation-driven API so that patch classes can take advantage of the distinction.

PR #1149, opened 2022-02-25
Implement invoker for Unsafe.allocateInstance

dmlloyd:


dgrove-oss (approved):
Is this a pure performance optimization, or is there a reason why just letting this flow through to the definition in Unsafe$_native doesn't work?
```
    public Object allocateInstance(Class<?> clazz) throws InstantiationException {
        return CompilerIntrinsics.emitNew(clazz);
    }
```

dmlloyd:
`emitNew` doesn't work during build time because the type IDs aren't assigned yet. This causes the type ID to be 0, leading to bizarre side effects.

dgrove-oss:
Got it. That's a nasty one.

PR #1161, opened 2022-02-28
(Re-)Introduce method types and common base for invokable types

dmlloyd:
This is a prerequisite for method handles. Originally I reverted this idea but I found a way to make it work cleanly. Using method type pointers helps resolve a problem where we don't really know what kind of thing a function pointer is actually pointing to.


dmlloyd:
As an aside, this is the result of manually peeling and reassembling a commit history for method handles, so it might take one or two tries to get this going through CI. :-)

dgrove-oss (approved):

dmlloyd:
Closed in favor of #1162.

PR #1164, opened 2022-03-02
Reachability improvement for static fields

dgrove-oss:
Only consider static fields that are actually accessed in reachable code
as roots for tracing the build time heap to identify instantiated types.
This replace a more conservative approach in which all static fields
of initialized types were considered as heap roots.


dgrove-oss:
I'm not 100% sure this change is totally correct in the presence of Unsafe access to static fields.  It's basically assuming that the Unsafe field accesses will go through a StaticFieldPointer that shows up in reachable code, and I suspect that isn't actually true.  We may need to accumulate an additional set of static fields that have had their offsets taken and conservatively assume they are also accessed for the purposes of reachability.

dmlloyd:
Ideally I would like to infer the pointer target of every pointer possible during build time, using the same or similar methodology compared to how `Pointer#offsetInBytes` is working, in the hopes that we can eventually change all unsafe methods to use this mechanism.  So I think the optimism is OK (we can revisit if necessary).

dmlloyd (approved):

PR #1170, opened 2022-03-03
Better queue support methods

dmlloyd:


dgrove-oss (approved):
Good catch... that's something I screwed up in #1160 when I separated dispatchable from invokable.

dgrove-oss (approved):

PR #1182, opened 2022-03-08
Add ability to set executable name

dmlloyd:


dgrove-oss (approved):
Fancy :)

dmlloyd:
Changed the wording to remove the word "executable", since we can (at least theoretically) output shared libraries as well.

PR #1191, opened 2022-03-10
improvements in RTA / Reachability Analysis

dgrove-oss:
+ Fully split the concepts of a dispatchable method and an invokable method
+ Improvements in RTA to only track classes that are actually instantiated
  (stop propagating instantiation up the class hierarchy).
+ Improvements in RTA to consider actual method lookup semantics when
  deciding if a method is reachable (don't force a superclass method
  that is not actually invokable to be invokable just because a subclass
  that also defines an overriding method is instantiated).
+ Reduce vtable size by not allocating slots for private methods;
  they will always be invoked via an invokespecial and therefore
  should not need to be allocated vtable slots.

### DIFF
--- a/plugins/dispatch/src/main/java/org/qbicc/plugin/dispatch/DispatchTables.java
+++ b/plugins/dispatch/src/main/java/org/qbicc/plugin/dispatch/DispatchTables.java
@@ -87,12 +87,17 @@ public class DispatchTables {
 
     void buildFilteredVTable(LoadedTypeDefinition cls) {
         tlog.debugf("Building VTable for %s", cls.getDescriptor());
+        ReachabilityInfo reachabilityInfo = ReachabilityInfo.get(ctxt);
 
         ArrayList<MethodElement> vtableVector = new ArrayList<>();
         for (MethodElement m: cls.getInstanceMethods()) {
-            if (ctxt.wasEnqueued(m)) {
-                tlog.debugf("\tadding reachable method %s%s", m.getName(), m.getDescriptor().toString());
-                ctxt.registerAutoQueuedElement(m);
+            if (!m.isPrivate() && reachabilityInfo.isDispatchableMethod(m)) {
+                if (reachabilityInfo.isInvokableInstanceMethod(m)) {
+                    tlog.debugf("\tadding dispatchable and invokable method %s%s", m.getName(), m.getDescriptor().toString());
+                    ctxt.registerAutoQueuedElement(m);
+                } else {
+                    tlog.debugf("\tadding dispatchable but not invokable method %s%s", m.getName(), m.getDescriptor().toString());
+                }
                 vtableVector.add(m);
             }
         }
@@ -114,10 +119,11 @@ public class DispatchTables {
 
     void buildFilteredITableForInterface(LoadedTypeDefinition cls) {
         tlog.debugf("Building ITable for %s", cls.getDescriptor());
+        ReachabilityInfo reachabilityInfo = ReachabilityInfo.get(ctxt);
 
         ArrayList<MethodElement> itableVector = new ArrayList<>();
         for (MethodElement m: cls.getInstanceMethods()) {
-            if (ctxt.wasEnqueued(m) && ! m.isPrivate()) {
+            if (reachabilityInfo.isDispatchableMethod(m) && ! m.isPrivate()) {
                 tlog.debugf("\tadding invokable signature %s%s", m.getName(), m.getDescriptor().toString());
                 itableVector.add(m);
             }
@@ -184,6 +190,7 @@ public class DispatchTables {
             return;
         }
         RuntimeMethodFinder methodFinder = RuntimeMethodFinder.get(ctxt);
+        ReachabilityInfo reachabilityInfo = ReachabilityInfo.get(ctxt);
         VTableInfo info = getVTableInfo(cls);
         MethodElement[] vtable = info.getVtable();
         Section section = ctxt.getImplicitSection(cls);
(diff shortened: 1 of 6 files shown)


dgrove-oss:
Got some reduction in reachable functions (633 on HelloWorld), I think mainly from the incorporation of actual method lookup semantics in RTA, which avoids making a superclass method reachable that is always overridden in all instantiated subclasses.
Before
```
[D] (org.qbicc.plugin.reachability) ANALYZE: Reachability Statistics
[D] (org.qbicc.plugin.reachability) ANALYZE:   Reachable interfaces:          238
[D] (org.qbicc.plugin.reachability) ANALYZE:   Reachable classes:             2162
[D] (org.qbicc.plugin.reachability) ANALYZE:   Instantiated classes:          1809
[D] (org.qbicc.plugin.reachability) ANALYZE:   Reachable functions:           16016
[D] (org.qbicc.plugin.reachability) ANALYZE:   Dispatchable instance methods: 11162
[D] (org.qbicc.plugin.reachability) ANALYZE:   Accessed static fields:        188
[D] (org.qbicc.plugin.reachability) ANALYZE:   Deferred dispatchable methods: 141
[D] (org.qbicc.plugin.reachability) ANALYZE:   Deferred exact methods:        40

```
After
```
[D] (org.qbicc.plugin.reachability) ANALYZE: Reachability Statistics
[D] (org.qbicc.plugin.reachability) ANALYZE:   Reachable interfaces:          238
[D] (org.qbicc.plugin.reachability) ANALYZE:   Reachable classes:             2159
[D] (org.qbicc.plugin.reachability) ANALYZE:   Reachable functions:           15373
[D] (org.qbicc.plugin.reachability) ANALYZE:   Dispatchable instance methods: 11341
[D] (org.qbicc.plugin.reachability) ANALYZE:   Invokable instance methods:    10529
[D] (org.qbicc.plugin.reachability) ANALYZE:   Accessed static fields:        188
[D] (org.qbicc.plugin.reachability) ANALYZE:   Instantiated classes:          1769
[D] (org.qbicc.plugin.reachability) ANALYZE:   Deferred dispatchable methods: 876
[D] (org.qbicc.plugin.reachability) ANALYZE:   Deferred exact methods:        39
```

plugins/dispatch/src/main/java/org/qbicc/plugin/dispatch/DispatchTables.java
```diff
@@ -87,12 +87,17 @@ public VTableInfo getVTableInfo(LoadedTypeDefinition cls) {
...
     void buildFilteredVTable(LoadedTypeDefinition cls) {
         tlog.debugf("Building VTable for %s", cls.getDescriptor());
+        ReachabilityInfo reachabilityInfo = ReachabilityInfo.get(ctxt);
 
         ArrayList<MethodElement> vtableVector = new ArrayList<>();
         for (MethodElement m: cls.getInstanceMethods()) {
-            if (ctxt.wasEnqueued(m)) {
-                tlog.debugf("\tadding reachable method %s%s", m.getName(), m.getDescriptor().toString());
-                ctxt.registerAutoQueuedElement(m);
+            if (!m.isPrivate() && reachabilityInfo.isDispatchableMethod(m)) {
```
dgrove-oss, line 94: I think this change to not put private methods in the vtable is correct, but I am not a totally reliable Java language lawyer... 
If it isn't correct, I'll need to have the invokeExact handling also make instance methods dispatchable....

dmlloyd: I agree that this seems correct.

dmlloyd (approved):

plugins/dispatch/src/main/java/org/qbicc/plugin/dispatch/DispatchTables.java
DanHeidinga: Sorry, late to the party on this but as of JDK 11 when NestMates feature was added, `invokevirtual` can now be used to invoke private methods.

Quoting from the JEP bug:
> With the change to the access rules, and with suitable adjustments to byte code rules, we can allow simplified rules for generating invocation bytecodes:
invokespecial for private nestmate constructors,
invokevirtual for private non-interface, nestmate instance methods,
invokeinterface for private interface, nestmate instance methods; and
invokestatic for private nestmate, static methods

https://bugs.openjdk.java.net/browse/JDK-8046171

dgrove-oss: Thanks Dan.  I'll do a follow up PR to put the private methods back into vtables.  Skimming the JEP, I think we also need to put them into itables (which I think we also might not be doing currently).

DanHeidinga: It's easiest to put them in the itables but I think we can avoid it.  With J9, we used the low bits of the itable index to indicate if this was a direct call or not so that the private methods didn't end up in the itable and were able to do the equivalent of a invokespecial after confirming the receiver implemented the interface.

DanHeidinga: Given the closed world, we may be able to aggressively re-write `invokevirtual/invokeinterface` of private methods to `invokespecial` and keep them out of all vtable/itables.

dgrove-oss: ok, so maybe we leave this as is for now and think further.

dmlloyd: I verified (empirically) that private methods are not using virtual dispatch even when made accessible via nestmates despite using the `invokevirtual` bytecode.

```java
class Scratch {
        final Sub sub = new Sub();
        System.out.println(((Base) sub).doIt());
    }

    private static class Base {
        private int doIt() {
            return 123;
        }
    }

    private static class Sub extends Base {
        private int doIt() {
            return 456;
        }
    }
}
```

Output:
```text
123

Process finished with exit code 0
```

Bytecodes:
```javap
Classfile /Users/david/Library/Caches/JetBrains/IdeaIC2021.2/compile-server/qbicc-parent_dc4ec81f/scratches/out/Scratch.class
[...]
  public static void main(java.lang.String[]);
    descriptor: ([Ljava/lang/String;)V
    flags: (0x0009) ACC_PUBLIC, ACC_STATIC
    Code:
      stack=2, locals=2, args_size=1
         0: new           #7                  // class Scratch$Sub
         3: dup
         4: invokespecial #9                  // Method Scratch$Sub."<init>":()V
         7: astore_1
         8: getstatic     #10                 // Field java/lang/System.out:Ljava/io/PrintStream;
        11: aload_1
        12: invokevirtual #16                 // Method Scratch$Base.doIt:()I
        15: invokevirtual #22                 // Method java/io/PrintStream.println:(I)V
        18: return
      LineNumberTable:
        line 3: 0
        line 4: 8
        line 5: 18
      LocalVariableTable:
        Start  Length  Slot  Name   Signature
            0      19     0  args   [Ljava/lang/String;
            8      11     1   sub   LScratch$Sub;
}
SourceFile: "scratch_1.java"
NestMembers:
  Scratch$Sub
  Scratch$Base
InnerClasses:
  private static #45= #7 of #28;          // Sub=class Scratch$Sub of class Scratch
  private static #46= #17 of #28;         // Base=class Scratch$Base of class Scratch

Classfile /Users/david/Library/Caches/JetBrains/IdeaIC2021.2/compile-server/qbicc-parent_dc4ec81f/scratches/out/Scratch$Base.class
[...]
  private int doIt();
    descriptor: ()I
    flags: (0x0002) ACC_PRIVATE
    Code:
      stack=1, locals=1, args_size=1
         0: bipush        123
         2: ireturn
      LineNumberTable:
        line 9: 0
      LocalVariableTable:
        Start  Length  Slot  Name   Signature
            0       3     0  this   LScratch$Base;
}
SourceFile: "scratch_1.java"
NestHost: class Scratch
InnerClasses:
  private static #22= #7 of #19;          // Base=class Scratch$Base of class Scratch

Classfile /Users/david/Library/Caches/JetBrains/IdeaIC2021.2/compile-server/qbicc-parent_dc4ec81f/scratches/out/Scratch$Sub.class
[...]
  private int doIt();
    descriptor: ()I
    flags: (0x0002) ACC_PRIVATE
    Code:
      stack=1, locals=1, args_size=1
         0: sipush        456
         3: ireturn
      LineNumberTable:
        line 15: 0
      LocalVariableTable:
        Start  Length  Slot  Name   Signature
            0       4     0  this   LScratch$Sub;
}
SourceFile: "scratch_1.java"
NestHost: class Scratch
InnerClasses:
  private static #22= #2 of #19;          // Base=class Scratch$Base of class Scratch
  private static #23= #7 of #19;          // Sub=class Scratch$Sub of class Scratch
```

So our present policy of converting invocations on private methods to exact invocations - and consequently keeping them out of vtables - still seems correct to me despite the usage of the `invokevirtual` bytecode.

PR #1207, opened 2022-03-14
Run CI on Mac OS as well

dmlloyd:
I suspect this may fail, due to `libgcc_s`, but TIAS I guess.

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -54,4 +54,45 @@ jobs:
         if: failure()
         with:
           name: test-reports
-          path: 'test-reports.tgz'
+          path: 'linux/test-reports.tgz'
+  build-macos:
+    name: "Mac OS x86_64"
+    runs-on: macos-latest
+    env:
+      CC: gcc
+    steps:
+      - name: Install LLVM
+        uses: KyleMayes/install-llvm-action@v1.5.0
+        with:
+          version: "12.0.0"
+          directory: ${{ runner.temp }}/llvm-install
+
+      - name: Install OpenJDK 17
+        uses: joschi/setup-jdk@v2
+        with:
+          java-version: '17'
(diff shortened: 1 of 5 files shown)


dmlloyd:
Current failure:

```text
[E] (org.qbicc.tests.integration.SimpleAppTest) ADD: <stdin>:7: error: C Tool Chain: fatal error: 'pthread.h' file not found

[E] (org.qbicc.tests.integration.SimpleAppTest) ADD: : error: C Tool Chain: #include <pthread.h>

[E] (org.qbicc.tests.integration.SimpleAppTest) ADD: : error: C Tool Chain:          ^~~~~~~~~~~

[E] (org.qbicc.tests.integration.SimpleAppTest) ADD: : error: C Tool Chain: 1 error generated.
```

dmlloyd:
Current failure:

```text
[E] (org.qbicc.tests.integration.SnippetsTest) GENERATE: ld: error: /usr/bin/gcc: library not found for -lgcc_s
```

We need to figure out a way to locate `libgcc_s.dylib` on here. It's not trivial to just make symlinks so something clever is called for.

dmlloyd:
Testing out using `brew` to find the GCC library search paths, plus adding the ability to add library search paths as options to `Main`.

dmlloyd:
It looks like the `brew`-installed `libgcc_s.dylib` is actually named `libgcc_s.1.dylib`, and it's apparently not finding it under the shorter name, so we'll have to overcome that somehow.  Maybe use `gcc_s.1` as the name on mac...?

dmlloyd:
Another thing we could do is to stop using the unwind ABI and instead do the stack unwind ourselves using the method information that we already gather (enhanced, perhaps, with landing pad addresses).  Requires a little bit of digging to work out the details though.

dgrove-oss (approved):

Issue #1177, opened 2022-03-07
Eagerly initialize MethodHandles for VarHandles


dgrove-oss:
When we instantiate a VarHandle at build time, we need to also eagerly initialize the MethodHandles that are going to be used at runtime at build time.  

The test program below fails with a bus error because `java.util.concurrent.FutureTask` uses a VarHandle to access its `runner` instance field from its `run` method and the corresponding MethodHandle was not created.

```
import java.util.concurrent.ExecutorService;
import java.util.concurrent.Executors;

public class HelloWorldExecutor {
    public static void main(String[] args) {
        ExecutorService e = Executors.newSingleThreadExecutor();
        e.submit(HelloWorldExecutor::run);
        //        e.shutdown();
    }

    static void run() {
        System.out.println("Hello World");
        System.out.flush();
    }
}
```

dgrove-oss:
This has been fixed; example program runs successfully.

PR #1208, opened 2022-03-15
fix vtable layout for VarHandle subclasses

dgrove-oss:


dgrove-oss:
This makes some adjustments in our vtable layout code to account for #1201.  The HelloWorldExecutor program from #1177 now executes successfully for me.

dmlloyd (approved):

PR #1211, opened 2022-03-15
Global variable/method debug improvements

dmlloyd:
This changes the nice-name of global variables and of methods so that they match Java syntax, to an extent anyway.

With this change Mac debug works as far as it can... however, when entering a Java frame with `lldb` you are hit with this message:

```text
This version of LLDB has no plugin for the language "java". Inspection of frame variables will be limited.
```

But it really means that inspection of *any* variables is essentially nonexistent.  So going forward, we have to decide whether to rewrite the debug info to seem like C++ or some other language, or else commit to getting support for Java debug info upstream to `lldb`.

### DIFF
--- a/driver/src/main/java/org/qbicc/driver/CompilationContextImpl.java
+++ b/driver/src/main/java/org/qbicc/driver/CompilationContextImpl.java
@@ -54,7 +54,6 @@ import org.qbicc.type.definition.element.ExecutableElement;
 import org.qbicc.type.definition.element.FieldElement;
 import org.qbicc.type.definition.element.FunctionElement;
 import org.qbicc.type.definition.element.InitializerElement;
-import org.qbicc.type.definition.element.InvokableElement;
 import org.qbicc.type.definition.element.MemberElement;
 import org.qbicc.type.definition.element.MethodElement;
 import org.qbicc.type.descriptor.ClassTypeDescriptor;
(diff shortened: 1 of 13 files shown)


dmlloyd:
Hmmm this is building cleanly for me locally...

dmlloyd:
I think the issue here involves using the `clang` assembler versus `gcc`, which is more restrictive.

dmlloyd:
It is possible to work around this issue at an assembly level in a way that GCC accepts, e.g. changing:

```asm
        callq   "java.lang.Object.getClass()Ljava/lang/Class;"@PLT
```
to:
```asm
        .set    Object_getClass, "java.lang.Object.getClass()Ljava/lang/Class;"
        callq   Object_getClass@PLT
```

It is also possible to sidestep this problem (and the problem where LLVM 13+ generates `.file` directives that `gas` doesn't understand) by compiling directly from `llc` to `o`.  However, I want to keep the door open to bypassing LLVM at some point in the future and compiling directly to assembly, so I think we need a solution that would work going forward. And, while we can force `gas` to accept this using the `.set` trick, I'm not sure that we can count on platform link-loaders and debuggers and other tooling to consume it correctly.

Unfortunately I think this might mean going back to some form of name mangling. But it might be a good idea to consider using a "standard" mangling scheme such as [the JNI one](https://docs.oracle.com/en/java/javase/17/docs/specs/jni/design.html#resolving-native-method-names) which is a pretty reasonable compromise between usability and compactness (as opposed to, say, appropriating C++ name mangling).  This could be done with a view towards contributing specific debugger support for understanding this naming syntax.

PR #1216, opened 2022-03-16
Manually expand `unw_getcontext` on aarch64

### LINKED ISSUE #1215: `unw_getcontext` expands to inline assembly on `aarch64-linux`
Our simple trick of detecting functions-as-macros falls apart for `unw_getcontext` on `aarch64-linux`.  The probe uses C preprocessor macro quoting/pasting to acquire the "true" name of the function, expecting something along the lines of '_Uaarch64_unw_getcontext()`.  But what we get in this case is:

```c
(({ unw_tdep_context_t *unw_ctx = (); register uint64_t *unw_base __asm__ (\"x0\") = (uint64_t*) unw_ctx->uc_mcontext.regs; __asm__ __volatile__ ( \"stp x0, x1, [%[base], #0]\\n\" \"stp x2, x3, [%[base], #16]\\n\" \"stp x4, x5, [%[base], #32]\\n\" \"stp x6, x7, [%[base], #48]\\n\" \"stp x8, x9, [%[base], #64]\\n\" \"stp x10, x11, [%[base], #80]\\n\" \"stp x12, x13, [%[base], #96]\\n\" \"stp x14, x13, [%[base], #112]\\n\" \"stp x16, x17, [%[base], #128]\\n\" \"stp x18, x19, [%[base], #144]\\n\" \"stp x20, x21, [%[base], #160]\\n\" \"stp x22, x23, [%[base], #176]\\n\" \"stp x24, x25, [%[base], #192]\\n\" \"stp x26, x27, [%[base], #208]\\n\" \"stp x28, x29, [%[base], #224]\\n\" \"str x30, [%[base], #240]\\n\" \"mov x1, sp\\n\" \"stp x1, x30, [%[base], #248]\\n\" : [base] \"+r\" (unw_base) : : \"x1\", \"memory\"); }), 0)"
```

We might want to add a guard here, and actually implement this in assembly, with this exact expansion, for this case.

dmlloyd:
Fixes #1215.


dmlloyd:
See also libunwind/libunwind#341.

dmlloyd:
This finished its test run on my aarch64 system (finally, after over an hour - it's pretty slow). Unfortunately the try-catch snippet in the integration tests crashes (segfault) but compilation does succeed. I don't *think* the crash is directly related to this (the LLVM code generated by us for `unw_getcontext` is now identical* to that produced by a simple C test program via `clang`) though.

* Apart from superficial (I think!) LLVM IR syntactical differences...

dmlloyd:
Exception throwing is also failing on M1 already, without this change (that M1 Macs do not need), so I think exception throwing on aarch64 is just broken right now.

PR #1220, opened 2022-03-17
Track non-primitive args to calls returned #1172

### LINKED ISSUE #1172: Fix anonymous class wrong stack allocation with EA
>David Lloyd: in the anonymous class assigned to jdk.internal.misc.TerminatingThreadLocal#REGISTRY, the new IdentityHashMap<>(4)) is allocated on the stack

galderz:
Fixes #1172 

If new allocations get passed to a call that is returned, at the method level we can't guarantee that they don't escape. So, make sure those are marked for tracking as argument escape.

### DIFF
--- a/plugins/optimization/src/main/java/org/qbicc/plugin/opt/ea/EscapeAnalysisIntraMethodBuilder.java
+++ b/plugins/optimization/src/main/java/org/qbicc/plugin/opt/ea/EscapeAnalysisIntraMethodBuilder.java
@@ -20,6 +20,7 @@ import org.qbicc.graph.CheckCast;
 import org.qbicc.graph.DelegatingBasicBlockBuilder;
 import org.qbicc.graph.Executable;
 import org.qbicc.graph.InstanceFieldOf;
+import org.qbicc.graph.Invoke;
 import org.qbicc.graph.LocalVariable;
 import org.qbicc.graph.New;
 import org.qbicc.graph.Node;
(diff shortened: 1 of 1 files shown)


plugins/optimization/src/main/java/org/qbicc/plugin/opt/ea/EscapeAnalysisIntraMethodBuilder.java
```diff
@@ -154,14 +159,24 @@ public void startMethod(List<ParameterValue> arguments) {
     public BasicBlock return_(Value value) {
         final BasicBlock result = super.return_(value);
 
-        // Skip primitive values truncated, they are not objects
-        if (!(value instanceof Truncate)) {
+        if (value instanceof Call call && !isPrimitive(call.getType())) {
```
dmlloyd, line 162: Other types of calls that can return are `Invoke` and `CallNoSideEffects`; however `CallNoSideEffects` is not allowed to modify memory already so you can exclude that one. But you do need to consider `Invoke`.

galderz: Hmmm, is that really possible? `Invoke` is not of type `Value`.

galderz: Or you mean `invokeNoReturn`?

dmlloyd: It is; to get the return value of an `Invoke`, you have to call `getReturnValue()` on it.  The reason it is like this is that it is the only terminator which has a value.

galderz: I'll handle `Invoke` once https://github.com/qbicc/qbicc/pull/1233 has been fixed.

galderz: Sorry for delay, I'm adding `Invoke`...

galderz:
@dmlloyd added `Invoke`

dmlloyd (approved):
Thanks Galder!

PR #1226, opened 2022-03-22
Remove interpreter halted exception

dmlloyd:
This does not really help as much as it was intended to. Instead, continue compilation, blocking off failed methods with an in-interpreter error thrown


dmlloyd:
I'm just putting this in draft for now until I can test it a little more thoroughly.

dgrove-oss (approved):
lgtm

PR #1231, opened 2022-03-25
Use alt. names for streams on Mac OS

dmlloyd:
Replaces #1230.


dmlloyd:
This approach seems to work but I accidentally gave only one `_` on the symbol prefixes when it should have two `__`. 🤞🏻

PR #1239, opened 2022-04-04
Array classes are both abstract and final

dmlloyd:


dmlloyd:
This change (I think) causes array classes to have no emitted vtable. We need to code an exception for classes which are not actually abstract.

dgrove-oss:
I give in :). Words mean what the OpenJDK spec says they mean...

So we need to adjust the vtable/itable code accordingly and special cases arrays

dmlloyd:
We have `I_ACC_NOT_REALLY_FINAL`, maybe `I_ACC_NOT_REALLY_ABSTRACT` is its natural companion. 🥲

dmlloyd:
Since the combination of `ACC_ABSTRACT` and `ACC_FINAL` is unique to arrays and primitives, adding a check for `isFinal` with `isAbstract` should be enough. Trying it now.

dgrove-oss (approved):

PR #1243, opened 2022-04-04
Preserve invariant that supertype IDs come before subtype IDs

dmlloyd:


plugins/instanceof-checkcast/src/main/java/org/qbicc/plugin/instanceofcheckcast/SupersDisplayBuilder.java
```diff
@@ -41,6 +41,7 @@ public void accept(CompilationContext ctxt) {
         tables.assignTypeID(jlo);
         // arrays, including reference array
         CoreClasses coreClasses = CoreClasses.get(ctxt);
+        tables.assignTypeID(coreClasses.getArrayBaseTypeDefinition());
```
dgrove-oss, line 44: We have `19` sprinkled around in a few random places as a magic number for typeIds (DispatchTables is guilty).  I think assigning a typeId to the arrayBase class might shift that to 20?

dmlloyd: Ah yes, I see. I'll edit those accordingly, but we probably should revisit that eventually...

dgrove-oss: Agreed.  It's not a good situation.

dgrove-oss (approved):

PR #1250, opened 2022-04-08
If CC is set in the environment, don't look for alternative compilers

dgrove-oss:
If the user is telling us to use a specific C compiler, then either it works or qbicc should fail.  Don't silently allow alternative C Tool Chains to also be considered.


DanHeidinga (approved):
lgtm

PR #1251, opened 2022-04-12
Reconfigure sections class hierarchy to track object offsets

### LINKED ISSUE #1249: Modify `ProgramModule` to declare section order
There are two dimensions of order inherent in linking within qbicc.  On one axis, we have ordering based on the linkage order of classes (see #1247).  Objects within a class file which appears earlier in the linkage order will in turn appear earlier than other objects within class files which occur later, assuming the two objects are in the same section.

On the other axis, sections are linked in the order that is defined by the linker.  At present we rely on the "implicit" section for programs and data, where the LLVM compiler decides what actual segment to place objects into. Additionally, we rely on the default ordering for sections that is imposed by the linker.

However, in order to support "load-and-go" programs which expect objects to load directly into GC regions for immediate execution, we must ensure that at least some sections are loaded in a specific position relative to the overall image.  For example, the sections(s) containing the initial heap objects must be placed at the end of the data segment so that the memory into which the sections are loaded can be made contiguous with a memory chunk which is `mmap`ed for the use of GC.

The consequence of this requirement is that we need a mechanism by which program linkage can be controlled. This is complicated by the implementation differences between the LLVM linker, the Mac OS variant of the LLVM linker, and GNU `ld`.

The linker invocation API that we define must be enhanced with controls which allow for custom sections to be defined along with the establishment of the relative order of sections within each target segment and the final executable.

#### Notes

* With GNU `ld`, use the command `ld --verbose` to dump the default linker script.
* Mac OS `ld` has a fairly comprehensive `man` page which explains how to add segments to sections.

dmlloyd:
Rework the API for sections, program modules, and program objects:

* Removes section from declarations; they're always in the implicit section.
* Sort declarations in order by name.
* Create concept of "data-only" section, with known offset tracking for data objects.
* Introduce separate `Section` and `ModuleSection` concepts: `Section` defines the properties of the section for the entire image, vs `ModuleSection` which contains all of the object definitions in a particular section in a particular program module.
* Create an initial `Segment` concept, and utility methods to format section and segment names based on the target platform.
* Introduce `Declaration` intermediate abstract class for declaration types, which connects to a `ProgramModule`, as the complement for `SectionObject` which connects to a `ModuleSection`.

Resolves #1249.

### DIFF
--- a/compiler/src/main/java/org/qbicc/context/CompilationContext.java
+++ b/compiler/src/main/java/org/qbicc/context/CompilationContext.java
@@ -17,8 +17,9 @@ import org.qbicc.interpreter.VmClassLoader;
 import org.qbicc.machine.arch.Platform;
 import org.qbicc.object.Function;
 import org.qbicc.object.ProgramModule;
+import org.qbicc.object.ModuleSection;
 import org.qbicc.object.Section;
-import org.qbicc.type.ClassObjectType;
+import org.qbicc.object.Segment;
 import org.qbicc.type.FunctionType;
 import org.qbicc.type.InvokableType;
 import org.qbicc.type.TypeSystem;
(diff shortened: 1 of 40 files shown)


dmlloyd:
There's a regression with method handles; checking it out.

dmlloyd:
#1254 fixes the regression which I think might be caused by a race condition around initial heap setup.

dgrove-oss (approved):
lgtm

PR #1260, opened 2022-04-20
Introduce type definitions for primitives

### LINKED ISSUE #1235: Primitive types should have type definitions
We should create type definitions for primitive types. There is a decent amount of code dedicated to dealing with the possibility that a `VmClass` might or might not have a type definition, and to deal with reserving special space for primitives at the start of the type table. If every type - including every primitive type - has a type definition then a lot of special-case code goes away.

This would also make it easier to treat native primitive/value types (like `uint64_t` for example, or even structure types) in a manner which is more consistent with Java language primitive types, if we wanted to explore this in the future.

Considerations:

* `DefinedTypeDefinition#getDescriptor()` currently returns a `ClassTypeDescriptor`; this would have to be generalized to a `TypeDescriptor` (since primitive types would have a descriptor of type `BaseTypeDescriptor`) with some impact on existing code.
* Similar problem with `getSignature()` (though to a much lesser extent).
* Since primitive classes cannot be loaded by name, they would have to be excluded from registration in the class loader.
* There should be an `I_ACC_VALUE` or `I_ACC_PRIMITIVE` flag to indicate whether a class is primitive; this is simpler and more extensible than the range check.
* Primitive classes already pre-load their array classes so nothing needs to be done here.

dmlloyd:
Fixes #1235.

### DIFF
--- a/compiler/src/main/java/org/qbicc/type/Primitive.java
+++ b/compiler/src/main/java/org/qbicc/type/Primitive.java
@@ -6,6 +6,10 @@ import org.qbicc.type.descriptor.BaseTypeDescriptor;
 
 public enum Primitive {
     // Predefine the set of primitive types in Java
+
+    // Void goes first
+    VOID("void"),
+
     BOOLEAN("boolean"),
     BYTE("byte"),
     SHORT("short"),
@@ -14,7 +18,7 @@ public enum Primitive {
     FLOAT("float"),
     LONG("long"),
     DOUBLE("double"),
-    VOID("void");
+    ;
 
     private final String name;
     volatile private int typeId;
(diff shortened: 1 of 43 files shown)


compiler/src/main/java/org/qbicc/type/Primitive.java
```diff
@@ -14,7 +18,7 @@ public enum Primitive {
     FLOAT("float"),
     LONG("long"),
     DOUBLE("double"),
-    VOID("void");
+    ;
```
dgrove-oss, line 21: extra comma after DOUBLE ?

plugins/instanceof-checkcast/src/main/java/org/qbicc/plugin/instanceofcheckcast/SupersDisplayBuilder.java
```diff
@@ -31,16 +32,25 @@ public void accept(CompilationContext ctxt) {
...
         tables.buildSupersDisplay(jlo);
         info.visitReachableSubclassesPreOrder(jlo, tables::buildSupersDisplay);
+        CoreClasses coreClasses = CoreClasses.get(ctxt);
+
         // Assign typeIDs to classes
         // [0] Poisoned entry for easier debugging
         tables.reserveTypeIds(1);
-        // [1 - 9] for primitive types
-        tables.assignTypeIdToPrimitives();
+        // [1 - 9] for primitive types and void
```
dgrove-oss, line 40: Nit picking...
// [1 - 9] for void and primitive types

dgrove-oss (approved):

dmlloyd:
On the downside, there is a test failure. On the upside, the logs are actually fairly short now so it doesn't take 5 minutes to show where it happened. 😀

PR #1265, opened 2022-04-26
generate flat class array

dgrove-oss:
Doing a CI test to verify I'm headed down a plausible path; not really ready to merge yet.


dgrove-oss:
This PR changes things so we generate a flat class array in the build time heap.  The array is not the first object defined (because Class objects refer to other objects, and its a big cyclic mess).  To get it to be the first definition in the heap, we'll most likely need to move the definition to its own section (or similar).  But that can be done separately.

dgrove-oss:
Realized over the weekend it was straightforward to put the root class array in a separate .ll file, so adding that functionality to this PR.

dmlloyd (approved):

PR #1268, opened 2022-04-28
Add Homebrew clang as option tool discovery

galderz:


dgrove-oss (approved):

dgrove-oss (changes requested):
Actually, @dmlloyd is right.  We should use clang from Xcode, not clang from homebrew on mac.

PR #1285, opened 2022-05-06
put interned strings into their own .ll file

dgrove-oss:
Organize heap to get interned strings into their own .ll file.

This almost does what we want, except for a single byte[] that "leaked" from its containing String object into a java.nio.fs.UnixPath object due to the use of JavaLangAccess.getBytesNoRepl.  When serializing the heap, we found the byte[] from the UnixPath object "first" and as a result there is a single pointer from the internedString section of the heap "out" to the main heap.

It would be easy to identify when this happens and generate a root table for GC scanning (which would be tiny) when we actually get to worrying about scanning the heap roots.

### DIFF
--- a/compiler/src/main/java/org/qbicc/interpreter/Vm.java
+++ b/compiler/src/main/java/org/qbicc/interpreter/Vm.java
@@ -238,6 +238,11 @@ public interface Vm {
     VmString intern(String string);
 
     /**
+     * Is the given VmObject and interned VmString?
+     */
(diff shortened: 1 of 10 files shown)


dgrove-oss:
It would be pretty easy to use different prefixes for names in the interned string section and main heap section.  Might be useful for later debugging of heap structure invariants?  (`@qbicc_initial_heap_obj_14` vs `@qbicc_interned_string_obj_14`)

dmlloyd (approved):

PR #1293, opened 2022-05-12
Fix vararg function signatures to be exactly correct

dmlloyd:


evacchi:
empirical evidence[1] seems to show that we should change `utf8z` in `CNativeIntrinsics`:

```
-            Literal literal = lf.literalOf(ts.getArrayType(ts.getUnsignedInteger8Type(), bytes.length), bytes);
+           Literal literal = lf.literalOf(ts.getArrayType(ts.getSignedInteger8Type(), bytes.length), bytes);
```

[1]: tried the snapshot on my fork:

```
[E] (org.qbicc.tests.integration.SimpleAppTest) ADD: FileDescriptor$_native.java:74: error: Invalid pointer conversion from pointer to u8 to pointer to s8
    member: method close0@ bci 35
  location: type FileDescriptor
```

dmlloyd:
I think you're right. While (in my opinion) it's more correct to represent a UTF-8 character with an unsigned 8-bit integer, in practice most string-accepting functions will expect pointer-to-`char` which is almost universally signed. I've added a method to the type system to get the native `char` type and `utf8z` now uses that.

PR #1297, opened 2022-05-13
Fix Unwind pointer type

evacchi:
I have hard evidence that these should fixed:

```
_Unwind_GetLanguageSpecificData
_Unwind_GetRegionStart
_Unwind_GetIP
_Unwind_SetIP
```

because they are returning/expecting a `uint64_t` whereas it should be a `uintptr_t` -- this is obviously only important for architectures where the word size is not 64 bits.

I am also updating `_Unwind_SetGR` because, although I haven't encountered issues directly, it seems consistent to do so


dmlloyd:
I think the "most correct" type is `unsigned long`, which is what is in the `unwind.h` header file; `uintptr_t` might be equivalent to `unsigned long` for every platform we care about but it's probably still a good idea to be exact.

dmlloyd:
Note that `unsigned long` is probably not actually "correct" for the upstream project. On Windows x86_64 for example, `unsigned long` is only 32 bits and thus is not large enough to hold all valid pointer values.

evacchi:
I referred to https://clang.llvm.org/doxygen/unwind_8h.html where:

> typedef uintptr_t  _Unwind_Ptr

dmlloyd:
Interesting. My header file doesn't quite match that.

dmlloyd:
I opened https://github.com/libunwind/libunwind/issues/361 at any rate.

evacchi:
updated to `unsigned long`

dmlloyd (approved):

PR #1299, opened 2022-05-13
Remove deprecated method

dmlloyd:


dmlloyd:
Hmm 🤔

dmlloyd:
This is a real problem with the "bridge hack". Apparently `javac` will still target the bridge method *if* it is more specific than the non-bridge method.

As a workaround, I think I can set up a capturing intrinsic which grabs calls to the old method, and then after a release cycle, we can remove that again.

dmlloyd:
OK so this is also a consequence of the fact that we're resolving the method before we find the intrinsic for it, so if a method doesn't exist, it doesn't exist - the intrinsic BBB will never have a chance to replace the call.  A tricky puzzle.

dmlloyd:
This is probably solvable but it's also not worth investing a lot of time right now. Closing.

PR #1311, opened 2022-05-18
VIO and VFS initial implementation

dmlloyd:
This is the initial virtual I/O and filesystem API and implementation.  It provides facilities to open virtual file descriptors which correspond to various host resource kinds, so that files in the virtual filesystem can be opened and manipulated.

- [x] Base VIO machine API
- [x] `HostIO` interface
- [x] VIO plugin for non-FS I/O operations
- [x] VFS machine API
- [x] VFS plugin for FS I/O operations

The module relationship is like this:
```mermaid
graph BT;
  pvfs("plugins/vfs")-->pvio("plugins/vio")
  pvfs-->mvfs("machine/vfs")
  mvfs-->mvio("machine/vio")
  pvio-->mvio
```


dgrove-oss:
wow...you've been busy!

dgrove-oss (approved):

PR #1314, opened 2022-05-25
Switch EA code to a visitor

galderz:
* Created a new EA intra method analysis class to include visitor logic.
* Removed deferred edges because they don't add anything in the current flow insensitive implementation of EA.
* ConnectionGraph now uses a custom ParameterValue array to capture start method ParameterValue instances.
* Enhance EA dot node visitor to add points-to and field edges.
* EA dot not edges involving terminators (e.g. Invoke.ReturnValue) are tricky to deal with immediately. Instead delay their printing until they're visited, in order to maintain correct subgraph structure.
* Switch to Value in points-to map value side.
* Remove local new nodes and add field edges only for ParameterValue.
* Tighten fieldEdges typing.


galderz:
Can we re-run CI? Mac one failed for some weird 403 error while getting the jdk.

dmlloyd (approved):

PR #1320, opened 2022-05-31
[#1257] Link files in ascending order by type ID

### LINKED ISSUE #1257: Link object files in order by corresponding `type_id` (ascending)

dmlloyd:
Fixes #1257. If two types have the same type ID (for example, -1) then they will have indeterminate order.  I am thinking maybe `LTDI.getTypeId` should throw an exception if the type ID was not set.


dgrove-oss (approved):
LGTM

PR #1335, opened 2022-06-08
Invoke `clang` to generate *.o from *.ll, skip `llc`, `opt`

evacchi:


evacchi:
my bad this is a bit too involved when using a pipe

PR #1346, opened 2022-06-10
Eliminate `ProgramObjectLiteral` (by `PointerLiteral`)

dmlloyd:
There are two ways to specify a pointer to a program object; since the interpreter requires support for `Pointer`, we should go with `PointerLiteral` so there is only one way to do it.


dgrove-oss (approved):
Makes sense;  no reason to have two ways to say the same thing...

PR #1348, opened 2022-06-10
Stub out GraalVM Feature emulation for qbicc

dgrove-oss:


dgrove-oss:
Not sure if it really makes sense to merge this before I actually connect it up to Reachability Analysis, but I'm not sure if I'll finish that today and wanted to give people a comment on the basic framework.

dgrove-oss:
If you build the native image for knative-quarkus-benchmark sleep benchmark, then you can compile with qbicc with the simple command line arguments:
`--jar sleep-1.0.0-SNAPSHOT-native-image-source-jar/sleep-1.0.0-SNAPSHOT-runner.jar --build-feature io.quarkus.runner.AutoFeature`

plugins/graalvm-nativeimage-emulation/src/main/java/org/qbicc/plugin/nativeimage/FeatureProcessor.java
```diff
@@ -0,0 +1,67 @@
...
+            rc.setAccessible(true);
+            Resources resourcesInstance = rc.newInstance();
+            qiss.add(Resources.class, resourcesInstance);
+        } catch (ClassNotFoundException | NoSuchMethodException | InstantiationException | IllegalAccessException | InvocationTargetException e) {
+            ctxt.error("Unable to instantiate com.oracle.svm.core.jdk.Resources");
+            return;
+        }
+
+        // Set up the current Thread to load application classes into the hostVM via its CCL
+        ClassLoader savedCCL = Thread.currentThread().getContextClassLoader();
```
dmlloyd, line 38: I'm not super excited about this approach in the long term - I think it could collapse in certain cases (like when there are library or JDK conflicts or significant target platform disparity). In the short term though it's fine to get something off the ground. Hopefully in the future we can use VM invokables, perhaps combined with helper runtime classes, or a similar solution to retain the isolation between host and application.

dmlloyd (approved):

plugins/graalvm-nativeimage-emulation/src/main/java/org/qbicc/plugin/nativeimage/FeatureProcessor.java
dgrove-oss: Yeah, I view this as a tactical approach to consume what Quarkus is currently generating.  If Quarkus makes progress on https://github.com/quarkusio/quarkus/issues/25975 and starts generating json files instead we could just read those json files directly instead of having to have a fake graalvm native image environment.

PR #1352, opened 2022-06-14
Remove collectedPointers, convert references

evacchi:
Remove the notion of "collected pointer" and use references instead. Convert instead of casting interned strings and objects of the inital heap.


evacchi:
re-draft, I'll try to get rid also of `getAddrspace()` since that would always return zero anyway

dmlloyd (approved):

PR #1357, opened 2022-06-16
Allow operations on values of UnresolvedType

dgrove-oss:
It's quite reasonable to allow programs to be compiled where a runtime
null value is allowed to flow through arguments, fields, and variables
of UnresolvedType.  Change our treatment of UnresolvedType to support
that.

### DIFF
--- a/compiler/src/main/java/org/qbicc/graph/ReferenceHandle.java
+++ b/compiler/src/main/java/org/qbicc/graph/ReferenceHandle.java
@@ -4,6 +4,7 @@ import org.qbicc.graph.atomic.AccessMode;
 import org.qbicc.type.ObjectType;
 import org.qbicc.type.PointerType;
 import org.qbicc.type.ReferenceType;
+import org.qbicc.type.UnresolvedType;
 import org.qbicc.type.definition.element.ExecutableElement;
 
 import static org.qbicc.graph.atomic.AccessModes.SinglePlain;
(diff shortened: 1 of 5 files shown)


compiler/src/main/java/org/qbicc/graph/ReferenceHandle.java
```diff
@@ -18,10 +18,12 @@ public final class ReferenceHandle extends AbstractValueHandle {
     ReferenceHandle(Node callSite, ExecutableElement element, int line, int bci, Value referenceValue) {
         super(callSite, element, line, bci);
         this.referenceValue = referenceValue;
-        if (referenceValue.getType().isComplete()) {
-            pointerType = ((ReferenceType) referenceValue.getType()).getUpperBound().getPointer();
+        if (referenceValue.getType() instanceof ReferenceType rt) {
+            pointerType = rt.getUpperBound().getPointer();
+        } else if (referenceValue.getType().isUnresolved()) {
```
dmlloyd, line 23: Since there's only one `UnresolvedType`, I think it'd be better to use `instanceof` here rather than have an `isUnresolved()` which only returns `true` for one type.

dgrove-oss: Fair enough.  Changed & squashed.

dmlloyd (approved):

PR #1358, opened 2022-06-17
Run programs from the application class path

dmlloyd:
This does not yet cover the following cases:

* Loading classes via `Class.forName()` or `appCl.loadClass()`
* Loading resources from the application class path

These would be done with additional invokers/intrinsics and we should probably have a story for how we want to bundle resources into the image before we go to that point.


dgrove-oss:
Should we change the -jar support I added last week to extend the appClassPath instead of the bootClassPath? 

https://github.com/qbicc/qbicc/blob/main/main/src/main/java/org/qbicc/main/Main.java#L1066-#L1089

dmlloyd:
> Should we change the -jar support I added last week to extend the appClassPath instead of the bootClassPath?

Ah, yes definitely. I'll update the PR.

dmlloyd:
Updated.

Issue #1356, opened 2022-06-16
Widening/Narrowing conversion with lambda capture


dgrove-oss:
This test program (extracted from larger examples) results in generating invalid llvm IR.  The problem is related to lexical capture of a boolean.  The function that creates the lambda has an i32; the invocation of that function has an i1.

```
import java.lang.reflect.AccessibleObject;
import java.security.AccessController;
import java.security.PrivilegedAction;

public class LambdaCapture {
    public static void setAccessible(AccessibleObject o, boolean flag) {
        AccessController.doPrivileged((PrivilegedAction<Void>) () -> {
                try {
                    o.setAccessible(flag);
                } catch (SecurityException ex) {
                    ex.printStackTrace();
                }
                return null;
            });
    }

    public static void main(String[] args) {
        setAccessible(null, false);
    }
}
```

Error:
```
/tmp/output/LambdaCapture.ll: error: /Users/dgrove/bin/opt: /Users/dgrove/bin/opt: <stdin>:278:397: error: argument is not of expected type 'i32'
  %L2 = call i8 addrspace(1)* (i8 addrspace(1)*, i8 addrspace(1)*, i8 addrspace(1)*, i32) @"_Jjava_lang_invoke_LambdaForm$DMH$36_newInvokeSpecial__Ljava_lang_Object_2Ljava_lang_Object_2I_Ljava_lang_Object_2"(i8 addrspace(1)* %thr0, i8 addrspace(1)* addrspacecast (%"class.java.lang.invoke.DirectMethodHandle$Constructor"* @qbicc_initial_heap_obj_54494 to i8 addrspace(1)*), i8 addrspace(1)* %p0, i1 zeroext %L1) "statepoint-id"="94905", !dbg !788 ; LambdaCapture.java:7 bci@2
```

dgrove-oss:
I also tried a variant when the captured variable was a `short` instead of a `boolean`.  Same problem.  So it's most likely a general missing widen operation at the call site , not something specific to booleans.

dgrove-oss:
I think what's going on here is our ReflectionIntrinsics for linking the method handles need to be prepared to do the same demote/promote operations that we do in MethodParser for the invokebytecodes.

dgrove-oss:
[...]

PR #1359, opened 2022-06-17
don't demote integral arguments to invokedynamic's method handle

dgrove-oss:
A potential fix for #1356, but I'm not 100% positive it is correct for all possible uses of invokedynamic.

It definitely is the right right thing for the MethodHandle's created to implement lambdas which collapse their parameters  to their basicTypeSignature to minimize the number of unique lambda classes created.  I spent a fair amount of time trying to see if there was a way to inspect the MethodHandle instance returned from `linkCallSite` and determine what descriptor should be used to invoke it, but wasn't convinced I'd found anything general.

FWIW, I believe if this is wrong, then we'll get an error from `llc` because the types of the actual parameters won't match the types of the function argument.  So at some level, if we aren't 100% positive this is right, its probably ok to merge it and wait until we find code that proves it wrong.


dmlloyd (approved):
This makes sense to me.

PR #1360, opened 2022-06-20
ref ptr refactor

evacchi:
- Use AddressSpace 0 for References
- wip
- Add LLVMReferencePointerFactory


dmlloyd (approved):
I think this approach makes sense, and also makes it fairly easy to implement compressed refs.

PR #1367, opened 2022-06-22
Skip IT compilation when `skipTests` is set

dmlloyd:


dmlloyd:
I think it must have skipped testing anyway! Hang on...

dgrove-oss (approved):

PR #1370, opened 2022-06-22
Do not assume that invocation target is an interface method

dmlloyd:


evacchi (approved):
Fixes the build when `Heap#initHeap()` is invoked explicitly rather than as a constructor

PR #1377, opened 2022-06-23
Add Emscripten support to linker and llvm module. Add branches to Main

evacchi:
Attempting to add Emscripten support to the main flow.

if `isWasm`:

- use `SIMPLE` reference pointers #1352 #1355 
- use `LLVMEmscriptenCompiler` instead of `LLVMCompilerImpl` (add LLVMCompiler interface and `LLVMCompiler.Factory`
- use `LLVMEmscriptenLinkStage` (goes through `emcc` instead of `ld`)

A lot of parameters are being hardcoded both in the linker and the compiler for now. Notably:

- `-s USE_ZLIB` so that `#include<zlib.h>` works
- `-pthread` is being ignored
- `libgcc_s` and `libunwind` are being ignored in the linker because support to exceptions has to be further investigated
- compilation still needs `-fwasm-exceptions` to finish

it is possible to run the test cases via node using the following commit https://github.com/evacchi/qbicc/commit/b982156c8715242e8daaf23b3df3266b882db911

still needs to remove the constructor https://github.com/evacchi/qbicc/commit/8629b1a43027aa3811efa29869c4408a75ac0de9

and invoke initHeap() in the class library https://github.com/qbicc/qbicc-class-library/pull/230

other than this, to try it is enough to set the platform to wasm-wasi in the plugin:

                        <configuration>
                            <mainClass>org.qbicc.tests.TestRunner</mainClass>
                            <platform>wasm-wasi</platform>
                        </configuration>

### DIFF
--- a/machine/tool/api/src/main/java/org/qbicc/machine/tool/CCompilerInvoker.java
+++ b/machine/tool/api/src/main/java/org/qbicc/machine/tool/CCompilerInvoker.java
@@ -120,6 +120,7 @@ public interface CCompilerInvoker extends MessagingToolInvoker {
     enum SourceLanguage {
         C,
         ASM,
+        LLVM_IR,
         ;
     }
 }
--- a/machine/tool/emscripten/src/main/java/org/qbicc/machine/tool/emscripten/EmscriptenCCompilerInvokerImpl.java
+++ b/machine/tool/emscripten/src/main/java/org/qbicc/machine/tool/emscripten/EmscriptenCCompilerInvokerImpl.java
@@ -5,7 +5,6 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 
-import org.qbicc.machine.arch.Platform;
 import org.qbicc.machine.tool.process.InputSource;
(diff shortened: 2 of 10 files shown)


machine/tool/api/src/main/java/org/qbicc/machine/tool/CCompilerInvoker.java
```diff
@@ -120,6 +120,7 @@ default void addDefinedSymbol(String name) {
     enum SourceLanguage {
         C,
         ASM,
+        IR,
```
dmlloyd, line 123: Minor detail, but this should be `LLVM_IR` since LLVM is not assumed at the C compiler level.

dmlloyd (approved):
Only one minor comment. Otherwise :+1:

machine/tool/api/src/main/java/org/qbicc/machine/tool/CCompilerInvoker.java
evacchi: done!

PR #1379, opened 2022-06-23
store the VmClassLoaderImpl in the Memory of the VmClassImpl in postConstruct

dgrove-oss:


dmlloyd:
Looks like a bad merge...

dgrove-oss:
Looks like it needs #1380 (now that we have non-null class loaders at runtime, we also need the getProtectionDomain0 fix from yesterday).

dmlloyd (approved):

Issue #1383, opened 2022-06-24
selected runtime Reflection Support


dgrove-oss:
I think we have many of the pieces done; sketching out a potential approach for feedback to make sure I haven't forgotten anything.

The big picture is that at build time, we want to allow a subset of the methods/fields/etc. in the program to be registered as needing to allow reflective access at runtime.   This allows us to support the reflective operations the program will actually perform while avoiding generating a complete description of the program that would allow unfettered reflective operations.

I've been looking at the internals of java.lang.Class today, and it seem like a viable approach would be to selectively populate instances of Class.ReflectionData at build time.  On all the code paths I've seen, if there is a ReflectionData available, the "higher level" reflection code will use it and not "call into the VM" to acquire the information dynamically.  It looks like this gives us a good control point to setup data structures at build time that we can use at runtime.  (As an optimization we could also initialize the `cachedConstructor` field for classes whose default constructor has been registered for reflection).

We can patch Class to eliminate the use of SoftReferences to "cache" the ReferenceData instances...they aren't cached for us.

We can build a general mechanism to register program elements for Reflection (a very limited form of this is already found in  from the RuntimeReflectionRoots in the reachability plugin).  The mechanism can then be used by the graalvm-native-image-emulation plugin (for Quarkus), but can also be made available via a more "qbicc-native" annotation system as well.

We have MethodHandles working.  We have the ability to do "new" on a Class instance via a compiler intrinsic.  The main bit we're missing is the programatic creation at build time of the VmObject graphs for ReflectionData and sticking them in the VmClass instances before we serialize them.

We should decide how we want to report errors where the user failed to properly register a program element at build time.  One possibility would be to initialize all Class instances with ReflectionData (where the vast majority of classes will share a single "no data" instance initialized with empty arrays).  I think this would then result in "NoSuch<Element>Exception" exceptions being raised at runtime when a non-registered access was attempted without us changing anything.

dgrove-oss:
One todo to look at (writing it down so I don't forget...).

#1420 set the exactDispatcher in the same block of native code where OpenJDK sets vmtarget/offset.  However, there's some sketchy looking in `MemberName(Field fld, boolean makeSetter)` that changes the Member from a getter to a setter after `MethodHandleNatives.init` is called.  I _think_ this means we're going to set our `exactDispatcher` to a getter inside of our `init` and miss when we need to change it to a setter.  

Need to look into this more carefully and perhaps patch/adjust the order in which things happen or otherwise compensate.

[...]

PR #1387, opened 2022-06-28
Initial cut at runtime reflection support

dgrove-oss:
Progress towards https://github.com/qbicc/qbicc/issues/1383

Needs https://github.com/qbicc/qbicc-class-library/pull/238

### DIFF
--- a/interpreter/src/main/java/org/qbicc/interpreter/impl/VmImpl.java
+++ b/interpreter/src/main/java/org/qbicc/interpreter/impl/VmImpl.java
@@ -104,6 +104,7 @@ public final class VmImpl implements Vm {
     // core classes
     final VmClassImpl objectClass;
     final VmClassClassImpl classClass;
+    final VmClassImpl reflectionDataClass;
     final VmStringClassImpl stringClass;
     final int stringCoderOffset;
     final int stringValueOffset;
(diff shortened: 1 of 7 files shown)


main/src/main/java/org/qbicc/main/Main.java
```diff
@@ -495,6 +495,17 @@ void call0(BaseDiagnosticContext initialContext) {
                                 builder.addBuilderFactory(Phase.ADD, BuilderStage.OPTIMIZE, SimpleOptBasicBlockBuilder::new);
                                 builder.addBuilderFactory(Phase.ADD, BuilderStage.INTEGRITY, ReachabilityBlockBuilder::new);
                                 builder.addBuilderFactory(Phase.ADD, BuilderStage.INTEGRITY, StaticChecksBasicBlockBuilder::new);
+                                builder.addPostHook(Phase.ADD, ctxt -> {
```
dgrove-oss, line 498: Not thrilled with having a big block of code here, but working around a circular dependency problem.  This code can't go in either reflection or reachability without causing a cycle through intrinsics, so if we don't want to put it in main, we need to add a new plugin just to have this that can import both reflection and reachability.

interpreter/src/main/java/org/qbicc/interpreter/impl/VmImpl.java
```diff
@@ -185,6 +186,7 @@ public final class VmImpl implements Vm {
         ClassContext bcc = ctxt.getBootstrapClassContext();
         classClass = new VmClassClassImpl(this);
         objectClass = classClass.getSuperClass();
+        reflectionDataClass = new VmClassImpl(this, classClass, bcc.findDefinedType("java/lang/Class$ReflectionData").load());
```
dmlloyd, line 189: This should have a `postConstruct` call down below (I think). Also the two-arg ctor should be OK since the `classClass` field is initialized at this point.

dgrove-oss: Good catch on the postConstruct.  Fixed.

dgrove-oss:
Reflective field access now works at runtime, as long as one builds the class library with a SNAPSHOT version of qbicc-rt that includes the fix in #1391.  Merging this PR isn't blocked on a qbicc-release followed by a class lib release, as the fix in #1391 only matters if one actually executes a reflective operation (and this needs getCallerClass to work).

dmlloyd (approved):

PR #1392, opened 2022-06-30
Phase 2 initialization with module setup

dmlloyd:
This progresses initialization a little farther. This change had to deal with two issues.

- Scheduling issue: we're doing pointer arithmetic on nodes which cannot be understood during interpretation. Originally my plan was to rework the scheduler to be smarter about delaying nodes, but this is a can of worms so it must be done incrementally and will take long enough that it's not acceptable for it to be a blocker for this. Instead, arithmetic on invalid pointers in the interpreter always yields a null pointer.
- Now the bad news: some of these invalid pointers are being used during interpretation, which is *not* expected (because memory accesses are ordered nodes and thus should never be moved outside of their conditionals, at least with the current scheduler). Because of this I'm making this a draft PR until I can figure out what's happening.

### DIFF
--- a/compiler/src/main/java/org/qbicc/interpreter/Vm.java
+++ b/compiler/src/main/java/org/qbicc/interpreter/Vm.java
@@ -37,6 +37,11 @@ public interface Vm {
     void initialize();
 
     /**
+     * Perform second-stage VM initialization in the currently attached thread.
+     */
+    void initialize2();
+
+    /**
      * Create a new thread.
      *
      * @param threadName the thread name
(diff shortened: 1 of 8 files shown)


dmlloyd:
The problem was due to not interpreting `AddressOf`, which actually works in the interpreter for most cases (local vars, fields, etc.)

dmlloyd:
Because we are still *really* speculative despite improvements like aggressive branch pruning, we speculatively try to access structure members that don't really exist everywhere. So for this kind of arithmetic, the choice is (once again) to either translate it into something invalid and trap the error late, or to make anything with any chance of being platform-specific be tied to control flow. I went with the former option.

PR #1401, opened 2022-07-01
refactor relationship between reachability and reflection

dgrove-oss:
Reworks the interactions between reachability, reflection, and annotation/feature processing with the goal of getting things into shape to allow us to generate all the method/constructor/field accessors we need to support runtime reflection in a controlled way during the ADD phase.


dgrove-oss:
I think the overall "shape" of the logic is right now.   

Reflective field access via @ReflectivelyAccessed annotations looks solid.

The idea of simply eagerly inflating Method and Constructor Accessors has hit what initially looks to be a nasty snag: the generated classes are exempted from verification, so jdk.internal.reflect.ClassFileAssembler doesn't bother to generate a StackMapTable, so qbicc refuses to compile the generated bytecode for invoke/newInstance because our phi generation code assumes it has stack maps.  So, we probably have to do the more involved approach of backporting enough of the JDK18 MethodHandle based code to get something limping along.

PR #1403, opened 2022-07-03
WASM: Configure `emcc` compilation+linking flags

evacchi:
- Configure some sensible defaults for `emcc` to compiler IR and link the Wasm+JS executable
- ~~Add Wasm to the list of Unix-like platforms~~
- Disable explicit linking of libunwind, gcc_s using `unless`
- Avoid invoking `mprotect()` during heap setup, since it is unsupported (prints a warning to stderr, causes the tests to fail)

Exceptions are still not supported (test TryCatch would fail, see also https://github.com/qbicc/qbicc/pull/1402)

### DIFF
--- a/machine/tool/emscripten/src/main/java/org/qbicc/machine/tool/emscripten/EmscriptenCCompilerInvokerImpl.java
+++ b/machine/tool/emscripten/src/main/java/org/qbicc/machine/tool/emscripten/EmscriptenCCompilerInvokerImpl.java
@@ -95,14 +95,23 @@ final class EmscriptenCCompilerInvokerImpl extends AbstractEmscriptenInvoker imp
                 cmd.add("-D" + key + "=" + val);
             }
         }
+
+        appendEmscriptenPorts(cmd);
+        enableExceptions(cmd);
+
         Collections.addAll(cmd,
-            "-s", "USE_ZLIB",
             "-Wno-override-module",
             "-mbulk-memory",
-            "-s", "ALLOW_MEMORY_GROWTH=1",
-            "-s", "EXIT_RUNTIME=1",
+            "-g",
+            "-c", "-x", sourceLanguageArg(), "-o", getOutputPath().toString(), "-");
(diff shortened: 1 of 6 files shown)


runtime/api/src/main/java/org/qbicc/runtime/Build.java
```diff
@@ -130,7 +130,7 @@ public static boolean isVirtual() {
 
         @Fold
         public static boolean isUnix() {
-            return defined(__unix__) || isMacOs();
+            return defined(__unix__) || isMacOs() || isWasm();
```
dmlloyd, line 133: Is there an emscripten-specific constant we can test for instead? We might (for example) use WASM in a bare WASI environment which isn't very UNIX-ish; IOW "WASM" doesn't imply "UNIX".

evacchi: I think `__WASI__` would imply unix, but there should be `__emscripten__` or something like that. Will check

dmlloyd: The docs at https://github.com/WebAssembly/WASI/blob/main/phases/snapshot/docs.md definitely imply a UNIX-like environment but I don't know if it is sufficiently UNIX-like to call it "UNIX".

I guess the definition of "UNIX" that we're actually using in qbicc is just "is there a `<unistd.h>`?" so maybe it's not actually a useful constant at all. For implementation we really only care if it's POSIX, Windows, or something else (and WASI would be something else *unless* we are using a POSIX API over the top of it, just as Windows could be considered POSIX *if* we used something like cygwin and had those POSIX header files and macros defined).

In conclusion, I guess it doesn't matter too much. We probably should remove `isUnix` altogether.

evacchi: I'll purge this commit from the PR since it's an orthogonal concern anyway

PR #1408, opened 2022-07-06
initialize MethodAccessors and ConstructorAccessors at build time

dgrove-oss:


dgrove-oss:
Won't actually work without https://github.com/qbicc/qbicc-class-library/pull/248, but I think it might pass CI anyways (since none of our tests are going to have an @ReflectivelyAccessed method or constructor that would activate this code).

dgrove-oss:
Strike that;  won't work without the class lib changes because it eagerly acquires the MethodElements it needs.

dmlloyd (approved):

PR #1416, opened 2022-07-11
WASM: Lower `throw` to `abort()`

evacchi:
Current implementation of EH leverages `libunwind`; lowering to a supported EH mechanism requires a supported runtime (in general this currently means "only NodeJS" or "V8"), which makes it harder to support smaller, embeddable runtimes such as `wasmer` and `wasmtime`.

This PR lowers `throw` to `abort()` allowing to do away with `libunwind` and EH in general for the time being. While this is not ideal, it is good enough to build self-contained nontrivial programs (e.g. good for simple plugin extensions, such as Envoy's*).

- WASM: AbortingThrowLoweringBasicBlockBuilder, abort()s instead of throwing
- WASM: Do not try to fill the stack trace

this is a follow-up to https://github.com/qbicc/qbicc/pull/1403 but may be merged independently

(*) disclaimer: Envoy currently has further limitations on the supported API surface so it does not work yet

### DIFF
--- a/main/src/main/java/org/qbicc/main/Main.java
+++ b/main/src/main/java/org/qbicc/main/Main.java
@@ -96,6 +96,7 @@ import org.qbicc.plugin.llvm.LLVMGenerator;
 import org.qbicc.plugin.llvm.LLVMIntrinsics;
 import org.qbicc.plugin.llvm.LLVMReferencePointerFactory;
 import org.qbicc.plugin.llvm.LLVMStripStackMapStage;
+import org.qbicc.plugin.lowering.AbortingThrowLoweringBasicBlockBuilder;
 import org.qbicc.plugin.lowering.BooleanAccessCopier;
 import org.qbicc.plugin.lowering.FunctionLoweringElementHandler;
 import org.qbicc.plugin.lowering.InitCheckLoweringBasicBlockBuilder;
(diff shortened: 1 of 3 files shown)


main/src/main/java/org/qbicc/main/Main.java
```diff
@@ -442,7 +443,7 @@ void call0(BaseDiagnosticContext initialContext) {
                                 if (llvm) {
                                     builder.addPreHook(Phase.ADD, LLVMIntrinsics::register);
                                 }
-                                builder.addPreHook(Phase.ADD, CoreIntrinsics::register);
+                                builder.addPreHook(Phase.ADD, ctxt -> CoreIntrinsics.register(ctxt, isWasm));
```
dmlloyd, line 446: Changing this can be avoided by instead using `ctxt.getPlatform().getCpu()` where the intrinsic is generated.

dmlloyd (approved):

plugins/intrinsics/src/main/java/org/qbicc/plugin/intrinsics/core/CoreIntrinsics.java
```diff
@@ -87,7 +85,11 @@ public static void register(CompilationContext ctxt) {
         registerJavaLangSystemIntrinsics(ctxt);
         registerJavaLangStackTraceElementInstrinsics(ctxt);
         registerJavaLangThreadIntrinsics(ctxt);
-        registerJavaLangThrowableIntrinsics(ctxt);
+        if (ctxt.getPlatform().getCpu() != Cpu.WASM32) {
+            registerEmptyJavaLangThrowableIntrinsics(ctxt);
+        } else {
+            registerJavaLangThrowableIntrinsics(ctxt);
+        }
```
evacchi, line 92: whoopsie I think this slipped the both of us @dmlloyd this is supposed to read '=='. I have no idea how come this has worked so far... (but probably I have made some mistake while cherry-picking from some integration branch I have locally)

PR #1418, opened 2022-07-13
Typed memory implementation

dmlloyd:
This change implements fully typed `Memory` instances.

The `Memory` instances for a given input type are bytecode-generated using the ASM API and loaded as hidden classes into the host JVM. Each member of the source structure is reflected by a correctly-typed and clearly-named field on the typed memory class. The performance characteristics are similar to the untyped memory implementation, but debuggability is vastly improved since every member of an object or structure can be easily examined in a debugger.

The base `VarHandleMemory` class works by implementing the `Memory` API with methods that delegate to a target `VarHandle`. The primary responsibility of the generated class is to contain an implementation of the `getHandle()` method which maps the given offset to a `VarHandle` for the appropriate field. Each `VarHandle` is a dynamic constant, so fields which are never accessed will never cause a `VarHandle` to be instantiated for it. At most, one `VarHandle` per member per type is generated, and reused for all instances of that type.

A property of hidden classes is that they may be defined with a "class data" object which can be easily loaded using a dynamic constant. This mechanism is used to populate the class with delegate factories for the case where a `Memory` contains nested `Memory` instances.

The vast majority of typed `Memory` implementations can be cloned using the JDK's `Object#clone()` shallow copy mechanism. This is very efficient, but only works when a shallow clone is sufficient. For memories which contain nested structures, a deep-clone copy constructor is generated and used by the `clone()` method.

A mechanism is included to automatically upgrade `long`-typed fields to `Pointer` type. For now, this flag is turned on all of the time, but in the future we could conceivably restrict this behavior to classes which are known to handle pointers in `long`-typed fields. Because typed memory is now also used for local variable storage, the same thing applies: the flag is always on for now, but can be changed later to use the same per-class logic. Note that if I finish implementing the method JIT mechanism then that *may* use standard Java local variables for local variable storage instead.

### DIFF
--- a/compiler/src/main/java/org/qbicc/interpreter/InvalidMemoryAccessException.java
+++ b/compiler/src/main/java/org/qbicc/interpreter/InvalidMemoryAccessException.java
@@ -1,4 +1,4 @@
-package org.qbicc.interpreter.impl;
+package org.qbicc.interpreter;
 
 import java.io.Serial;
 
--- a/compiler/src/main/java/org/qbicc/interpreter/Memory.java
+++ b/compiler/src/main/java/org/qbicc/interpreter/Memory.java
@@ -1,9 +1,17 @@
 package org.qbicc.interpreter;
 
+import static org.qbicc.graph.atomic.AccessModes.SinglePlain;
+
 import org.qbicc.graph.atomic.ReadAccessMode;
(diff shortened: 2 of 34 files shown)


dgrove-oss:
Looks pretty cool!

dgrove-oss (approved):

PR #1421, opened 2022-07-15
Add --llvm-opt-option and --llvm-llc-option to pass LLVM's opt and ll…

michihirohorie:


main/src/main/java/org/qbicc/main/Main.java
```diff
@@ -950,6 +956,12 @@ private static class GraphGenMethodsPhases {
             List<String> phases;
         }
 
+        @CommandLine.Option(names = "--llvm-opt-option", split = ",", description = "Pass options of the LLVM opt command")
```
dgrove-oss, line 959:
```suggestion
        @CommandLine.Option(names = "--llvm-opt-option", split = ",", description = "Pass options to the LLVM opt command")
```

main/src/main/java/org/qbicc/main/Main.java
```diff
@@ -950,6 +956,12 @@ private static class GraphGenMethodsPhases {
...
+        private List<String> optOptions = new ArrayList<String>();
+
+        @CommandLine.Option(names = "--llvm-llc-option", split = ",", description = "Pass options of the LLVM llc command")
```
dgrove-oss, line 962:
```suggestion
        @CommandLine.Option(names = "--llvm-llc-option", split = ",", description = "Pass options to the LLVM llc command")
```

dmlloyd (approved):

michihirohorie:
Thank you for the fixes.

PR #1425, opened 2022-07-18
Remove `ctxt` instance field, replace with class data entry

dmlloyd:
This avoids having a `ctxt` field on every single memory object.

Also introduce some constants for things that are being constructed on every method call. In particular, `Class#descriptorString` appears to do no caching.


dmlloyd:
Tested pretty thoroughly I think.

PR #1427, opened 2022-07-19
Rename visitor generic parameter

dmlloyd:
If you call the visitor parameter by the same name (lowercased) as the type or type variable, then the implementation name that is suggested will be given a name by the IDE corresponding to the resolved type argument, which is much more convenient.

This was done with regex replacement: `(T )?param` → `$1t`.


dmlloyd:
This is low-impact and not a functional change so I'm going to go with it.

PR #1428, opened 2022-07-19
Delete unused class `InitializerHandle`

dmlloyd:


dmlloyd:
Since none of this code is actually used by anything, I'm going to go ahead with it.

PR #1429, opened 2022-07-19
Rename `ValueHandle#getType` to be consistent with `Value`

dmlloyd:
This might help pave the way towards unifying `ValueHandle` and `Value` once again.


dgrove-oss (approved):
LGTM - I only skimmed the first commit (since it looked like an IDE-driven refactor).  Checked the second commit more carefully to see that the intended `op` was being used in the new version.

PR #1433, opened 2022-07-20
Draft of full commit series: converting `ValueHandle` to `Value`

dmlloyd:
This is not yet intended to be merged. The purpose of this PR is to preview the full commit series - as it evolves - which gets us to the point where `ValueHandle` is eventually completely dropped in favor of `Value`. This gives context to the related PRs that I've opened or will open related to this change.

The current state of the PR includes a few other currently-open PRs; as these are merged, I'll rebase this one and create more individual PRs for additional incremental refactors.  The goal of each refactor is to get us a step closer to the end state while minimizing disruption; once we've reached the end state, this PR will be empty.

The *current* end state of the PR is that `ValueHandle` is renamed to `PointerValue`, which now extends `Value` with corresponding changes to visitors and `instanceof` chains, with passing tests.


dmlloyd:
Closed in favor of cleaner approach.

PR #1436, opened 2022-07-25
Add isPThreads() build-time property

evacchi:
Allows to opt-out of PThreads support if that is not available (currently only Wasm is affected)

(this can be rebased on #1435 if it makes more sense to use `isWasi()` instead)


dmlloyd (approved):
Over time we can expand the definition but this simple one is fine to start with.

PR #1438, opened 2022-07-25
WASM: JLT#start0 early return instead of pthread_create()

evacchi:
it may change to `Build.Target.isPThreads()` if #1436 is merged

### DIFF
--- a/runtime/main/src/main/java/org/qbicc/runtime/main/VMHelpers.java
+++ b/runtime/main/src/main/java/org/qbicc/runtime/main/VMHelpers.java
@@ -1,6 +1,7 @@
 package org.qbicc.runtime.main;
 
 import org.qbicc.runtime.AutoQueued;
+import org.qbicc.runtime.Build;
 import org.qbicc.runtime.Hidden;
 import org.qbicc.runtime.Inline;
 import org.qbicc.runtime.InlineCondition;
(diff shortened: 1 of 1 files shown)


dmlloyd:
> it may change to `Build.Target.isPThreads()` if #1436 is merged

Let's do it!

runtime/main/src/main/java/org/qbicc/runtime/main/VMHelpers.java
```diff
@@ -221,6 +222,10 @@ public static void JLT_start0(void_ptr_unaryoperator_function_ptr runFuncPtr, vo
         // TODO once invokedynamic is working for lambda expressions use addr_of_function to fetch the correct function pointer
         //function_ptr<UnaryOperator<void_ptr>> runFuncPtr = addr_of_function(VMHelpers::threadWrapperNative);
 
+        if (Build.Target.isPThreads()) {
```
dmlloyd, line 225: Did you mean:
```suggestion
        if (! Build.Target.isPThreads()) {
```

evacchi: I did. I also though I had pushed the fix :D

dmlloyd (approved):

PR #1439, opened 2022-07-25
WASM: Add __EMSCRIPTEN__ to isWasi

evacchi:
While testing the changes I discovered that the current build setup will not work -- apparently when building with `emcc` the `__wasi__` constant may not be defined. I plan to do some follow-up PRs to actually switch to pure clang+wasi and/or keep emcc around as an alternate toolchain.

In the meantime, I am building the generated `*.ll`s (that compile fine with `emcc`) manually using clang+wasi at the shell prompt.

for now, though, I am setting `isWasi() := true` when either `__wasi__` or `__EMSCRIPTEN__` are defined, even if it is not 100% correct. I will clean it up when the two are actually split.

### DIFF
--- a/runtime/api/src/main/java/org/qbicc/runtime/Build.java
+++ b/runtime/api/src/main/java/org/qbicc/runtime/Build.java
@@ -197,7 +197,7 @@ public final class Build {
 
         @Fold
         public static boolean isWasi() {
-            return defined(__wasi__);
+            return defined(__wasi__) || defined(__EMSCRIPTEN__);
         }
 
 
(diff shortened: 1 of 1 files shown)


runtime/api/src/main/java/org/qbicc/runtime/Build.java
```diff
@@ -197,7 +197,7 @@ public static boolean isWasm() {
 
         @Fold
         public static boolean isWasi() {
-            return defined(__wasi__);
+            return defined(__wasi__) || defined(__EMSCRIPTEN__);
```
dmlloyd, line 200: Relatedly, you might want to add a WASI condition to `org.qbicc.plugin.core.ConditionEvaluation` as well.

dmlloyd (approved):

PR #1443, opened 2022-07-26
convert LDC of an UnresolvedType into a throw of NCDFE

dgrove-oss:


dmlloyd (approved):

dgrove-oss:
take two....this time it is actually correct :0

PR #1449, opened 2022-07-28
Introduce type-based `switch`

dmlloyd:
Add a new kind of `switch` node which operates on type IDs and doesn't require the values to be in order. During lowering, we can sort the values by their assigned type IDs and transform it to a regular integer `switch` that the backend can handle.

This is an alternative to generalizing `Switch` such that it accepts (for example) cases defined as `Literal`.


dmlloyd:
There is no language or bytecode construct that could reasonably be coerced into emitting this node type, however it would be useful in intrinsics and generated code. One key example is `System.arraycopy` which currently utilizes an inefficient `if`-`instanceof` tree.

An alternative would be to introduce a virtual method on our synthetic array base class and rely on vtable dispatch to find the correct implementation by intrinsifying `arraycopy` into a method call on the array object. This might allow a more friendly pure-Java implementation for array copying, but would require a way to patch methods onto the various array classes - something which we should consider finding a way to support in any event.

An in-between alternative would be to intrinsify `arraycopy` using a type switch to choose a static delegate method to call, where the delegate method would be implemented in plain Java.

PR #1452, opened 2022-07-29
VFS implementation of checkAccess

dgrove-oss:


dgrove-oss:
It seems plausible that we want to allow any file that exists in the VFS to be readable, but maybe I'm overlooking a use case.  In any case, this (combined with https://github.com/qbicc/qbicc-class-library/pull/260) lets the javax.crypto subsystem get fully initialized at build-time.

dmlloyd (approved):

PR #1457, opened 2022-08-11
reachability fix for lambdas

dgrove-oss:
Lambda instances are instantiated via Unsafe.allocateInstance,
so reachability analysis can't depend on seeing a `New` for
every type that is instantiated.  Work around this by recognizing
invokes to `<init>` that aren't from a `super.<init>` and
marking the declaring type of the `<init>` as being instantiated.

### DIFF
--- a/plugins/reachability/src/main/java/org/qbicc/plugin/reachability/ReachabilityBlockBuilder.java
+++ b/plugins/reachability/src/main/java/org/qbicc/plugin/reachability/ReachabilityBlockBuilder.java
@@ -185,6 +185,18 @@ public class ReachabilityBlockBuilder extends DelegatingBasicBlockBuilder implem
             if (visitUnknown(param, (Node)node)) {
                 ConstructorElement target = node.getExecutable();
                 param.analysis.processReachableExactInvocation(target, param.currentElement);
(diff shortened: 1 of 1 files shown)


plugins/reachability/src/main/java/org/qbicc/plugin/reachability/ReachabilityBlockBuilder.java
```diff
@@ -185,6 +185,13 @@ public Void visit(ReachabilityContext param, ConstructorElementHandle node) {
             if (visitUnknown(param, (Node)node)) {
                 ConstructorElement target = node.getExecutable();
                 param.analysis.processReachableExactInvocation(target, param.currentElement);
+                // The lambda meta factory and other reflective machinery in the JDK use Unsafe.allocateInstance
+                // instead of a "real" New to instantiate a class.  To compensate, reachability analysis assumes
+                // that if a ConstructorElementHandle is invoked and the invocation isn't the super.<init> that starts
+                // all non-Object consturctors, then the declaring type of the constructor _must_ have been instantiated somehow.
```
DanHeidinga, line 191:
```suggestion
                // all non-Object constructors, then the declaring type of the constructor _must_ have been instantiated somehow.
```

DanHeidinga (approved):
Looks reasonable.

PR #1458, opened 2022-08-12
Annotations to register reflective access to other elements

dgrove-oss:
Add annotations to allow a class to indicate that it will reflectively
access fields, methods, and constructors defined in other
classes. This complements the existing @ReflectivelyAccessed
annotation which allows annotating an element "in place" as being
reflectively accessed.


dgrove-oss:
My motivating example was from java.logging.  If java.util.logging.LogManager is reachable, then we need to tell reachability analysis about the default classes that are going to be instantiated reflectively by reading the logging.properties file.  These annotation allows us to encode that by annotating the LogManager class as reflectively accessing the constructors of SimpleFormatter, ConsoleLogger, etc.

dmlloyd (approved):

PR #1463, opened 2022-08-19
Fix vtable entry for the native method that has its implementation

michihirohorie:
Since some native methods' implementations are given by the `$_native` class in the qbicc-class-library, it looks fine to fill proper entries in vtables. `Runtime$_native` methods can be invoked with this change.


dgrove-oss (approved):
This should work, and I agree is needed because right now the plugin that resolves $_native methods doesn't clear the ACC_NATIVE bit in the class file.

PR #1473, opened 2022-09-02
Revert "compatibility kludge to allow relocation of qbicc_bound_thread field"

dgrove-oss:
This reverts commit 0c6c6e018ddf69df9d7d8b97b7612ef0dc4eae29.


dgrove-oss:
Does not work the way it was supposed to... will try a different approach.

dgrove-oss:
Never mind.  Found a simpler way out of the current dead end with fewer steps to completion.

PR #1488, opened 2022-09-20
define isatty in Unistd

dgrove-oss:


dmlloyd (approved):
The test failure was an unrelated sporadic infrastructure error so I've restarted it.

PR #1496, opened 2022-09-23
ensure we create JarFiles with appropriate runtime version

dgrove-oss:


driver/src/main/java/org/qbicc/driver/ClassPathElement.java
```diff
@@ -69,7 +70,7 @@ public static ClassPathElement forDirectory(Path path) {
      */
     public static ClassPathElement forJarFile(Path path) throws IOException {
         Assert.checkNotNullParam("path", path);
-        return new JarFileClassPathElement(new JarFile(path.toFile()));
+        return new JarFileClassPathElement(new JarFile(path.toFile(), true, ZipFile.OPEN_READ, JarFile.runtimeVersion()));
```
dmlloyd, line 73: This only works if the host JVM happens to be the same version as the class libs version. We probably need to propagate a parameter so that `Main` can communicate the RT JAR version somehow? Maybe that should be a follow up though; it might be somewhat nontrivial. WDYT?

dgrove-oss: That makes sense.  We should be able to get our hands on either `</version.qbicc.classlib>` and parse it or get out hands on the main RT.jar and extract from there.  Let me take a look and see if doing the right thing is relatively easy.

dgrove-oss: Wasn't too hard to propagate down from the command line argument processing. Fixed.

dmlloyd (approved):

PR #1497, opened 2022-09-26
enforce that RuntimeAspect fields cannot be accessed at build-time

dgrove-oss:
This cannot be merged until a class library that includes https://github.com/qbicc/qbicc-class-library/pull/290 has released.

### DIFF
--- a/plugins/patcher/src/main/java/org/qbicc/plugin/patcher/ClassContextPatchInfo.java
+++ b/plugins/patcher/src/main/java/org/qbicc/plugin/patcher/ClassContextPatchInfo.java
@@ -385,7 +385,7 @@ final class ClassContextPatchInfo {
                     if (isStatic && runTimeAspect) {
                         classPatchInfo.runtimeInitField(new RuntimeInitializerPatchInfo(internalName, i, initResolver, initIndex, fieldDesc, fieldName, controllingAnnotation));
                     }
-                    classPatchInfo.addField(new FieldPatchInfo(internalName, i, addModifiers, classFile, fieldDesc, fieldName, controllingAnnotation, null));
+                    classPatchInfo.addField(new FieldPatchInfo(internalName, i, 0, classFile, fieldDesc, fieldName, controllingAnnotation, null));
                 } else if (kind == K_REMOVE) {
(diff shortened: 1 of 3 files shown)


plugins/patcher/src/main/java/org/qbicc/plugin/patcher/ClassContextPatchInfo.java
```diff
@@ -385,7 +385,7 @@ void processClass(final ClassContext classContext, final String className) {
                     if (isStatic && runTimeAspect) {
                         classPatchInfo.runtimeInitField(new RuntimeInitializerPatchInfo(internalName, i, initResolver, initIndex, fieldDesc, fieldName, controllingAnnotation));
                     }
-                    classPatchInfo.addField(new FieldPatchInfo(internalName, i, addModifiers, classFile, fieldDesc, fieldName, controllingAnnotation, null));
+                    classPatchInfo.addField(new FieldPatchInfo(internalName, i, 0, classFile, fieldDesc, fieldName, controllingAnnotation, null));
```
dmlloyd, line 388: Maybe this should use `fieldMods` instead of zero? For example we might lose the `static` modifier.

dgrove-oss: What was there before was `addModifiers`, which isn't necessary now that the RuntimeInitializerPatch has the responsibility of adding `I_ACC_RUN_TIME` to the field.   We'll get `fieldMods` from the classfile that we are passing in as the resolver.  I toyed with removing the addModifiers parameter and associated code from `FieldPatchInfo` since we didn't need it (today...), but decided to leave it in case we need it later.

Passing in `fieldMods` here would be harmless, but probably should also be done in the `REPLACE` branch as well for symmetry. 

I don't have a strong opinion about any of the three options (remove, pass fieldMods in both add/replace, pass 0 always).  WDYT?

dmlloyd: Oh, right yes that makes sense. I forgot that this was literally *added* modifiers, so I think your change is correct.

dmlloyd (approved):

PR #1499, opened 2022-09-27
wip/probe bundles

evacchi:
- Probe bundles: wip
- Probe bundles: wip
- setup
- Prototype loading probes
- fixes/workarounds for wasm probes
- setup for CI
- setup for CI
- setup for CI
- setup CI for downloading bundles (🤞)


dmlloyd:
> `Error: Artifact path is not valid: /bundles/x86_64-linux-gnu-elf/types/void *.yaml. Contains the following character: Asterisk *`
> 
> dang. I thought I could get away with that on *nix :P

I think we probably should have a separate file for the probes that are done in `Main` (excluding reference and type ID sizes), something with each size like:

```yaml
--
platform-type-info:
    byte-bits: 8
    endian: little
    signed-char: true
    pointers:
        size: 8
        align: 8
    boolean:
        size: 1
        align: 1
    int8:
        size: 1
        align: 1
    # ...etc...
```

In addition to avoiding the `void *` problem this also reduces the number of files in total.

dmlloyd:
I think this was more or less implemented in an alternative branch, at least the important aspects.

PR #1503, opened 2022-09-29
Disable destructor

dmlloyd:
The theory here is that the destructor runs once the main thread exits, even if other threads are running, resulting in SIGSEGV for memory that was valid heap only moments before. The proper fix will probably involve blocking the destructor until all other threads have exited, potentially forcing daemon threads to terminate abruptly first.

### DIFF
--- a/runtime/gc/heap/src/main/java/org/qbicc/runtime/gc/heap/Heap.java
+++ b/runtime/gc/heap/src/main/java/org/qbicc/runtime/gc/heap/Heap.java
@@ -420,7 +420,7 @@ public final class Heap {
         return num;
     }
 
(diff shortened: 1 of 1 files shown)


dgrove-oss (approved):
That makes sense.  Our shutdown sequence is more Java-like now where there's nothing special about the main thread.  The process starts shutting down when the last non-daemon thread exits.  

Should we add an explicit call to destroyHeap() as part of that "VM" tear down process?

dmlloyd:
Maybe eventually. But first, I want to learn more about how we are going to set up the heap for GC and how things might work in a shared library situation.

PR #1506, opened 2022-09-29
Introduce block arguments

dmlloyd:
Introduce the concept of "block arguments" alongside, and eventually replacing, phi nodes.

Block arguments provide a simpler model for control flow by representing inbound block values as parameters whose actual values are given at the terminator for each predecessor block. The parameters are named, not positional, to easily facilitate "categories" such as JVM stack, JVM local variable, function argument, thrown exception, etc.

This is part one of a series of changes that will (among other things) eliminate phi nodes, fix the inliner, and replace the exception handler kludge from SimpleBasicBlockBuilder with a simpler mechanism.

For more information about block arguments, see:

 * https://mlir.llvm.org/docs/Rationale/Rationale/#block-arguments-vs-phi-nodes
 * https://github.com/apple/swift/blob/main/docs/SIL.rst


dgrove-oss (approved):
This all looks reasonable to me.

I'd love to be able to use block arguments instead of Phi nodes.  One thing that I'm still a bit unclear on from the LLVM documentation though is whether block arguments are supported in LLVM IR or whether this means that we are committing to generate MLIR instead (or whether we do a "backend" conversion ourself to translate the block arguments into phi nodes before generating llvm ir).

dmlloyd:
On the LLVM backend, they are transformed into phi nodes (it's actually a pretty straightforward transformation and the original phi lowering code is nearly unchanged (so far) to support this).

In a way it's better because LLVM has the hard requirement of listing phi nodes first in the block; since we have all of the block parameters and arguments ahead of writing the block, we can easily do this without worrying about how `PhiNode` is scheduled anymore.

PR #1509, opened 2022-09-30
support serialization of "native" memory arising from Unsafe.allocateMemory0

dgrove-oss:


dgrove-oss:
Motivating example is dealing with io.netty.buffer.EmptyByteBuf.  Trying to delay initializing this class to runtime has rippling effects and it is actually quite easy for us to serialize the ByteArrayMemory that represents "native" memory in the interpreter.

dmlloyd (approved):

PR #1511, opened 2022-10-04
two small fixes for quarkus

dgrove-oss:


plugins/intrinsics/src/main/java/org/qbicc/plugin/intrinsics/core/UnsafeIntrinsics.java
```diff
@@ -436,6 +441,10 @@ private static void registerPutIntrinsics(final CompilationContext ctxt) {
                     Value obj = arguments.get(0);
                     Value offset = arguments.get(1);
                     Value value = arguments.get(2);
+                    if (typeNameAndDesc.getValue().equals(BaseTypeDescriptor.Z)) {
```
dgrove-oss, line 444: It really seems like this _should_ be doable with a cast or conversion instead of a select, but after several attempts that didn't work I went with this.

dgrove-oss:
Need some more fixes to the qbicc feature processor....will open a new PR once I figure that out.

PR #1515, opened 2022-10-11
ServiceLoader support

dgrove-oss:
Implements only the provider configuration aspect of ServiceLoaders.  Module-based ServiceLoaders left for future work.

Inspired by (and replaces) #566.
Fixes #544.

PR #1517, opened 2022-10-12
Implement `MethodHandle`-based hook registration API

dmlloyd:
This API allows hook methods to be defined using a natural syntax by matching annotated methods on a hook class with corresponding methods in the interpreter.

The given hook class is instantiated using its sole constructor. If the constructor accepts a `Vm` or `VmImpl`, then that is injected. Other injectable values may be added in the future.

Each annotated method on the class is examined. The method may be `static` or non-`static` irrespective of whether the method being hooked is `static`. Non-`static` hook methods will be invoked upon using a singleton instance of the hook class.

The method to intercept is determined based on the name of the hook method. If more than one method with the same name exists on the target class, then a method descriptor string *must* be given on the `@Hook` annotation. If the name of the hook method differs from the name of the target method, then the `name` attribute can be given to the `@Hook` class and that name will be used instead.

A best effort is made to adapt the argument and return types of the hook method. For example, a hook method may accept a `VmString`, and if possible, the argument will be cast to that type. This is also true for `long` vs `Pointer`. Casting errors will result in an `Error` being thrown within the interpreted program.


dmlloyd:
This can wait until after #1515; I can resolve the conflicts and update it at that time.

dgrove-oss (approved):
I looked at the first 6 commits and the last commit fairly carefully.  Seemed good to me. I assume the rest of the commits that are the class-by-class ports to the new system are fine.

dgrove-oss:
#1515 only added 1 invokable.  I think it probably makes sense to merge this and I can easily rebase and update #1515.  Probably better to get this in, since we have to stall 1515 on a classlib release.

PR #1518, opened 2022-10-12
Add a couple necessary nodes for GC support

dmlloyd:
The following node types are added:

- `Reachable` - this node provides the implementation of `Reference.reachabilityFence(Object)`, by causing the given value to stay alive until at least that point
- `SafePoint` - this node represents a safepoint poll; no GC implements this (including NoGC) but this will eventually lower to e.g. a `test` instruction on the safepoint page address


dgrove-oss (approved):
LGTM

PR #1519, opened 2022-10-12
Add live-out calculation for basic blocks

dmlloyd:
This calculation sets the stage to allow us to stop using LLVM's `RS4GC` pass in favor of inserting our own `gc.statepoint` calls. In addition to potentially improving the LLVM part of compilation time, this should also make it possible to eventually support GC in non-LLVM backends.


dgrove-oss (commented):
The algorithm looks right to me.

It's not clear why locallyUsedRefs needs to be updated via a VarHandle-based compare and set while liveOuts can use a simpler scheme based on locking the basic block.  Is a comment explaining this warranted?

dgrove-oss (approved):

dmlloyd:
I've added a comment which explains the choice of locking or not.

PR #1522, opened 2022-10-13
Drop redundant `void` `Return`

dmlloyd:
Most other nodes use an empty `void`-typed value to signify `void` return; we can do the same for returns by dropping `Return`. Also rename `ValueReturn` to just `Return`.


compiler/src/main/java/org/qbicc/graph/Return.java
```diff
@@ -6,46 +6,71 @@
...
 
     int calcHashCode() {
-        return Objects.hash(Return.class, dependency);
+        return Objects.hash(Return.class, dependency, returnValue);
     }
 
     @Override
     String getNodeName() {
-        return "Return";
+        return "ValueReturn";
```
dgrove-oss, line 55: name this just "Return"?

dgrove-oss (approved):

compiler/src/main/java/org/qbicc/graph/Return.java
dmlloyd: Ah, good catch.

PR #1523, opened 2022-10-13
Recursive block building API

dmlloyd:
Introduce a recursion-friendly block building API. The purpose behind this API is to be able to construct blocks in the midst of constructing another block. The API presents a cleaner way of handling block early termination (by completely preventing its escape). It also encourages a safer pattern of returning the first block created rather than the last (or some other) block (as may happen when a terminator call is intercepted and replaced with a complex subprogram).

Another problem that arises from the current architecture is that the concept of the "first" or "entry" block is vague at best. In some cases we're creating the "first" block second, causing some difficult wiring situations since the builder assumes the first block begun will be the entry block.

Using something like this recursive API exclusively has the advantage that the first block will always be the first block; we could for example cause all block building to `finish` when the first block is done, returning the entry block. We could ensure that further blocks are not built after the first block completes, and therefore that all successor blocks must be created within the lexical scope of the first block.

The following use cases would be expected to be solved by this API:

* Method parsing:
   * Creating exception handlers on demand at their first use site
   * Reduce the amount of saved state needed
* Inliner:
   * Creating the inlined control structure while the calling block is active allows the inliner to avoid creating extraneous control breaks when inlining is canceled due to cost

### DIFF
--- a/compiler/src/main/java/org/qbicc/graph/BasicBlockBuilder.java
+++ b/compiler/src/main/java/org/qbicc/graph/BasicBlockBuilder.java
@@ -5,6 +5,8 @@ import static org.qbicc.graph.atomic.AccessModes.*;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.function.BiConsumer;
+import java.util.function.Consumer;
 
 import org.qbicc.context.ClassContext;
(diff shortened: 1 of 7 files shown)


dmlloyd:
Replaced by #1563.

PR #1532, opened 2022-10-17
extend ResourceLoading snippet with getResource/getResources tests

dgrove-oss:
Depends on #1531 and https://github.com/qbicc/qbicc-class-library/pull/301


dgrove-oss:
I pulled the test extensions that logically belong in #1530 into a separate PR to reduce the size of the changes that had to wait for a class lib release.

dmlloyd (approved):

PR #1541, opened 2022-10-21
Introduce switchable safepoint strategy

dmlloyd:
This introduces an API and implementation for implementing actual safepoint probes. The simple implementation uses a single static boolean field to request a safepoint.

The actual body of the safepoint is not yet implemented.

The next step is to implement an API which allows plugins to register safepoint steps (for example, marking the thread stack for GC, or synchronizing all threads to perform a sweep/compact operation).


dgrove-oss (approved):
LGTM

PR #1558, opened 2022-11-07
add native method for mkdir to SysStat

dgrove-oss:


runtime/posix/src/main/java/org/qbicc/runtime/posix/SysStat.java
```diff
@@ -81,4 +81,6 @@ public static final class struct_stat64_ptr extends ptr<struct_stat64> {}
     public static native c_int lstat(const_char_ptr pathName, struct_stat_ptr statBuf);
     @name(value = "fstatat$INODE64", when = { Build.Target.IsMacOs.class, Build.Target.IsAmd64.class })
     public static native c_int fstatat(c_int dirFd, const_char_ptr pathName, struct_stat_ptr statBuf, c_int flags);
+
+    public static native c_int mkdir(const_char_ptr pathname, c_int mode);
```
dmlloyd, line 85: There is a `mode_t` in `SysTypes`; that probably should be used instead in case it's not the same as `int`.

dgrove-oss: yeah.  Sorry.  I'd originally defined mkdir in Unistd.java and clearly didn't look at the rest of SysStat when I moved it to the right file.  Fixed.

dmlloyd (approved):

PR #1563, opened 2022-11-08
Replace `PhiValue` with `BlockParameter` completely

dmlloyd:
* Utilize the recursive block building API introduced in #1523
* Rely more heavily on the scheduler to produce a canonical node order
* Use a specific context object for building BBBs
* More useful methods on BBB
* Fix bugs in stack map processing
* Drop phi nodes in favor of block parameters
* Materialize constraint nodes into control flow branches (pi node problem)
* Enable last `TryCatch` test which was disabled
* Merge #1555 which now passes
* Basic blocks are now named in LLVM by their scheduled ID numbers
* Some LLVM values are now named for their scheduled block plus param slot _or_ schedule position; this may be worth expanding universally as it helps readability

There is still a lot of room for improvement but such improvements should be simpler now.


dmlloyd:
Closes #1523.

plugins/optimization/src/main/java/org/qbicc/plugin/opt/InliningBasicBlockBuilder.java
```diff
@@ -84,14 +82,12 @@ public Value call(ValueHandle target, List<Value> arguments) {
...
-                returnVal.setValueForBlock(ctxt, toInline, basicBlock, val);
-                return basicBlock;
-            }, () -> begin(resumeLabel));
+            BasicBlock inlined = doInline(target, toInline, arguments, null, val -> goto_(resumeLabel, Slot.result(), val), () -> {
+                begin(resumeLabel);
+                addParam(resumeLabel, Slot.result(), toInline.getType().getReturnType());
+            }, Map.of());
             if (inlined != null) {
-                return returnVal;
+                return null; // todo: Slot.result()
```
dgrove-oss, line 90: is returning `null` a safe enough "this isn't done yet bail our" or should we be throwing an exception?

dgrove-oss (approved):
Wow -- lots of good stuff.  I just had one question about inlining.  The rest looks plausible and simpler.  It's great to see phi nodes replaced by block args!

plugins/optimization/src/main/java/org/qbicc/plugin/opt/InliningBasicBlockBuilder.java
dmlloyd: Inlining is basically completely broken at the moment so it doesn't make a difference right now. Block parameters should allow a cleaner implementation but I didn't want to clutter up this change with yet more craziness.

PR #1577, opened 2022-11-10
GC: `DecodeReference` node

dmlloyd:
In order to ensure that only references are live across safepoints (and not their derived pointers), introduce a new `DecodeReference` ordered node which does not cross safepoints. This node replaces `ReferenceHandle` which is removed.

In order for this to work, a few bug fixes were necessary, and I hit a couple others on the way along with some helpful optimizations.


compiler/src/main/java/org/qbicc/graph/Throw.java
```diff
@@ -41,6 +41,10 @@ public Node getDependency() {
         return dependency;
     }
 
+    public boolean maySafePoint() {
+        return false;
```
dgrove-oss, line 45: Does this imply that we won't be able to scan a Thread stack during exception delivery?  Probably ok, but maybe overly limiting?

dgrove-oss (approved):
LGTM

compiler/src/main/java/org/qbicc/graph/Throw.java
dmlloyd: My thinking is that `Throw` is lowered to a native function which in turn doesn't safepoint. We can still scan the stack but we just can't rely on the immediate caller's frame for variable information. This should be OK though because all we need to do at this point is walk the stack return addresses which are generally stable even outside of safepoints (after all, we didn't even have safepoints until very recently 🙂).

PR #1586, opened 2022-11-15
Change three more `ValueHandle` to literal or equivalent

dmlloyd:
* `StaticField` becomes a literal
* `GlobalVariable` becomes a literal
* `LocalVariable` goes away in favor of debug declarations, phis, and/or stack allocation


interpreter/src/main/java/org/qbicc/interpreter/impl/Frame.java
```diff
@@ -1006,7 +1006,7 @@ public Object visit(VmThreadImpl thread, IsGe node) {
         } else if (isFloat64(inputType)) {
             return Boolean.valueOf(unboxDouble(left) >= unboxDouble(right));
         } else if (isInt64(inputType)) {
-            return Boolean.valueOf(unboxLong(left) >= unboxLong(right));
+            return Boolean.valueOf(isSigned(inputType) ? compareLongSigned(left, right, -1) >= 0 : compareLongUnsigned(left, right, -1) >= 0);
```
dgrove-oss, line 1011: Should we be doing something similar (although much simpler since they can't be hiding pointers) for the rest of the Integer types in the next stanza of the if-then-else?
```
isSigned(inputType) ? Integer.compare(left, right) ... : Integer.compareUnsigned(left, right) ...
```
Or is there an invariant that the types must be signed because of the way the graph is constructed?  Or something else...

dgrove-oss (approved):
LGTM - only one question

interpreter/src/main/java/org/qbicc/interpreter/impl/Frame.java
dmlloyd: Yes I think so. I'll do a separate PR for it.

dmlloyd:
Added a commit to address the failures on Mac which arose from using `new c_int[2]` to allocate a pair of integers (much like pipe). The commit translates `new NativeType[k]` to an automatic stack allocation.

dmlloyd:
The change wasn't sufficient; more class lib changes will be needed to move to the proper stack allocation idioms on Mac OS.

PR #1594, opened 2022-11-16
adjust cnative API structure for dirent

dgrove-oss:
It appears that struct_dirent is expected to be in dirent.h on most posix systems, but is in sys/dirent.h on MacOS.


dgrove-oss:
https://github.com/qbicc/qbicc-class-library/pull/323 didn't compile on linux; so try a different way of doing this...

dmlloyd (approved):

PR #1597, opened 2022-11-21
Introduce separate `LiteralVisitor` interface

dmlloyd:
Since `Literal` is a subtype of `Value`, `LiteralVisitor` is a supertype of `ValueVisitor`, which is modified with appropriate adaptors. Existing value visitors thus remain compatible and do not require additional changes.


dmlloyd:
I didn't do `ValueVisitorLong` yet because I'm not sure we need it anymore. But I didn't want to delete it yet either. Once the whole series of `ValueHandle` changes are in, if it's still not used we can drop it.

dmlloyd:
Forgot to update the literals themselves (I extracted this from another commit series...)

dmlloyd:
Fixed up some imports.

dgrove-oss (approved):
LGTM

PR #1602, opened 2022-11-28
Convert method calls from `ValueHandle` to `Value`

dmlloyd:


compiler/src/main/java/org/qbicc/graph/InterfaceMethodLookup.java
```diff
@@ -0,0 +1,38 @@
...
+    }
+
+    @Override
+    String getLabel() {
+        return "virtual";
+    }
+
+    @Override
+    String getNodeName() {
+        return "VirtualMethodLookup";
```
dgrove-oss, line 22:
```suggestion
        return "InterfacelMethodLookup";
```

plugins/instanceof-checkcast/src/main/java/org/qbicc/plugin/instanceofcheckcast/InstanceOfCheckCastBasicBlockBuilder.java
```diff
@@ -46,6 +51,103 @@ public InstanceOfCheckCastBasicBlockBuilder(final FactoryContext ctxt, final Bas
         this.ctxt = getContext();
     }
 
+    @Override
```
dgrove-oss, line 54: These unreachable type optimizations make sense, but maybe we can need to come up with a more descriptive name for this BBB now.  (Then again, the same comment could be made about the entire plugin-in....).

plugins/lowering/src/main/java/org/qbicc/plugin/lowering/InvocationLoweringBasicBlockBuilder.java
```diff
@@ -212,25 +185,21 @@ public ValueHandle visit(ArrayList<Value> args, VirtualMethodElementHandle node)
...
-        args.addAll(0, List.of(fb.load(fb.currentThread(), SingleUnshared), node.getInstance()));
-        final MethodElement target = node.getExecutable();
+        LiteralFactory lf = getLiteralFactory();
         DispatchTables dt = DispatchTables.get(ctxt);
         DispatchTables.ITableInfo info = dt.getITableInfo(target.getEnclosingType().load());
         if (info == null) {
             // No realized invocation targets are possible for this method!
-            MethodElement method = RuntimeMethodFinder.get(ctxt).getMethod("raiseIncompatibleClassChangeError");
-            throw new BlockEarlyTermination(fb.callNoReturn(staticMethod(method), List.of()));
+            return lf.nullLiteralOfType(target.getType().getPointer());
```
dgrove-oss, line 202: I think this changes the exception from an IncompatibleClassChangeError to a NullPointerException.  Was the exception wrong before for this case (ie, we should have got the NPE before we got the ICCE)?

dgrove-oss (approved):
Looks good; just a few small comments

plugins/lowering/src/main/java/org/qbicc/plugin/lowering/InvocationLoweringBasicBlockBuilder.java
dmlloyd: If there are no realized targets then the invocation could not have been reachable with a non-`null` receiver, thus a previous `null` check on the receiver would always have fired. Complicating matters, since the method lookups are now schedulable nodes, they could be hoisted up to an earlier block before any such null check, thus we can't safely introduce flow control at this point because it might be too early, so producing a literal is the safest bet, and I considered `null` to be the best choice since that will cleanly segfault if such a lookup result is ever used.

plugins/instanceof-checkcast/src/main/java/org/qbicc/plugin/instanceofcheckcast/InstanceOfCheckCastBasicBlockBuilder.java
dmlloyd: True. But it's possible these checks are not actually needed; I'll double-check.

dmlloyd: Yeah, they're not needed. This was part of an attempt to fix a reachability problem that I eventually resolved differently, so I've removed these checks.

plugins/lowering/src/main/java/org/qbicc/plugin/lowering/InvocationLoweringBasicBlockBuilder.java
dgrove-oss: ok. makes sense.

PR #1609, opened 2022-11-30
Fixes to get Quarkus compiling again

dgrove-oss:
Collection of fixes to get Quarkus compiling again.  Probably easiest to review each commit in turn.  

The changes in 2a8360ede do what I need, but depend on visitors to Literals being "well-behaved" and not replacing a Literal with an arbitrary Value as a result of their visit.  Maybe this should be enforced by tightening the return types of the visit methods?


compiler/src/main/java/org/qbicc/graph/Node.java
```diff
@@ -283,10 +301,53 @@ public Value visitUnknown(final Copier param, final Value node) {
...
             public Value visitAny(Copier copier, Literal literal) {
-                // literals remain unchanged by default
+                // literals without value dependencies remain unchanged by default
+                Assert.assertTrue(literal.getValueDependencyCount() == 0);
                 return literal;
             }
 
+            public Value visit(Copier copier, ArrayLiteral literal) {
+                List<Literal> values = copier.copyLiterals(literal.getValues());
+                return new ArrayLiteral(literal.getType(), values);
```
dmlloyd, line 311:
```suggestion
                return copier.getBlockBuilder().getLiteralFactory().literalOf(literal.getType(), values);
```

This approach would avoid needing to make the ctors public, and would allow us to collapse duplicates if we ever added a cache for it. Similar comment for the other ones.

interpreter/src/main/java/org/qbicc/interpreter/impl/VmClassImpl.java
```diff
@@ -514,13 +515,28 @@ public Literal getValueForStaticField(FieldElement field) {
...
+                        if (value == null) {
+                            return lf.zeroInitializerLiteralOfType(sit);
+                        } else {
+                            return lf.valueConvertLiteral(lf.literalOf(value), sit);
+                        }
+                    }
+                }
+                default:
+                    throw Assert.impossibleSwitchCase(sit.getMinBits());
+            }
```
dmlloyd, line 539:
```suggestion
            return switch (sit.getMinBits()) {
                case 8 -> lf.literalOf((byte) staticMemory.load8(offset, SinglePlain));
                case 16 -> lf.literalOf((short) staticMemory.load16(offset, SinglePlain));
                case 32 -> lf.literalOf(staticMemory.load32(offset, SinglePlain));
                case 64 -> {
                    try {
                        yield lf.literalOf(staticMemory.load64(offset, SinglePlain));
                    } catch (InvalidMemoryAccessException e) {
                        Pointer value = staticMemory.loadPointer(offset, SinglePlain);
                        if (value == null) {
                            yield lf.zeroInitializerLiteralOfType(sit);
                        } else {
                            yield lf.valueConvertLiteral(lf.literalOf(value), sit);
                        }
                    }
                }
                default -> throw Assert.impossibleSwitchCase(sit.getMinBits());
            }
```
Note: I've never actually used the `yield` syntax but it *should* work...

dmlloyd (commented):
Looks broadly good, just a couple of minor points.

compiler/src/main/java/org/qbicc/graph/Node.java
dgrove-oss: Good suggestion.  Fixed.

interpreter/src/main/java/org/qbicc/interpreter/impl/VmClassImpl.java
dgrove-oss: fancy :)  Fixed.

dmlloyd (approved):

PR #1610, opened 2022-12-01
bugfix in stack allocation

dgrove-oss:
Prefer to get the type of the variable being initialized from the initializer instead of the local variable entry unless the initializer is of type void (from a `auto()`). Fixes a problem with stack allocation of native arrays.

### DIFF
--- a/compiler/src/main/java/org/qbicc/type/definition/classfile/MethodParser.java
+++ b/compiler/src/main/java/org/qbicc/type/definition/classfile/MethodParser.java
@@ -309,7 +309,7 @@ final class MethodParser {
                     varType = lve.getType();
                     init = lf.zeroInitializerLiteralOfType(varType);
                 }
-                Value allocated = gf.stackAllocate(varType, lf.literalOf(1), lf.literalOf(al.getPointeeType().getAlign()));
+                Value allocated = gf.stackAllocate(varType, lf.literalOf(1), lf.literalOf(varType.getAlign()));
                 gf.store(allocated, init, SingleUnshared);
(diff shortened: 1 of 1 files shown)


compiler/src/main/java/org/qbicc/type/definition/classfile/MethodParser.java
```diff
@@ -303,11 +303,13 @@ void setLocal2(int index, Value value, int bci) {
...
-                Value allocated = gf.stackAllocate(lve.getType(), lf.literalOf(1), lf.literalOf(al.getPointeeType().getAlign()));
                 Value init = al.getInitializer();
-                if (init instanceof Literal lit && lit.isZero()) {
-                    init = lf.zeroInitializerLiteralOfType(lve.getType());
+                ValueType varType = init.getType();
+                if (init.getType().equals(ts.getVoidType())) {
+                    varType = lve.getType();
+                    init = lf.zeroInitializerLiteralOfType(varType);
                 }
+                Value allocated = gf.stackAllocate(varType, lf.literalOf(1), lf.literalOf(al.getPointeeType().getAlign()));
```
dmlloyd, line 312: I think the alignment probably should be `varType.getAlign()`, since `al`'s pointee type might be `void`.

dmlloyd (approved):
One comment otherwise looks good.

compiler/src/main/java/org/qbicc/type/definition/classfile/MethodParser.java
dgrove-oss: Changed.

PR #1611, opened 2022-12-01
simplify CompoundType.hashCode to support linked structs

dgrove-oss:
Avoid requiring the hashCode of all members to be computed as part of computing the hashCode of a CompoundType.  This avoids an infinite recursion when the CompoundType is a C-struct that contains an member which is a pointer to the same type, the default idiom for defining linked data structures in C.

### DIFF
--- a/compiler/src/main/java/org/qbicc/type/CompoundType.java
+++ b/compiler/src/main/java/org/qbicc/type/CompoundType.java
@@ -34,11 +34,9 @@ public final class CompoundType extends ValueType {
     private volatile List<Member> members;
     private volatile List<Member> paddedMembers;
     private volatile Map<String, Member> membersByName;
-    // late-computed hash code
-    private int hashCode;
 
     CompoundType(final TypeSystem typeSystem, final Tag tag, final String name, final Supplier<List<Member>> membersResolver, final long size, final int overallAlign) {
-        super(typeSystem, 0);
+        super(typeSystem, CompoundType.class.hashCode());
         // name/tag do not contribute to hash or equality
         this.tag = tag;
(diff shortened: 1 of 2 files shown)


compiler/src/main/java/org/qbicc/type/CompoundType.java
```diff
@@ -34,11 +34,9 @@ public final class CompoundType extends ValueType {
     private volatile List<Member> members;
     private volatile List<Member> paddedMembers;
     private volatile Map<String, Member> membersByName;
-    // late-computed hash code
-    private int hashCode;
 
     CompoundType(final TypeSystem typeSystem, final Tag tag, final String name, final Supplier<List<Member>> membersResolver, final long size, final int overallAlign) {
-        super(typeSystem, 0);
+        super(typeSystem, CompoundType.class.hashCode());
```
dmlloyd, line 39: We do not want the hash code of every `CompoundType` to be the same for performance reasons (we key maps by type with some regularity). How about leaving this as 0 but overriding `hashCode` to return `System.identityHashCode(this)`?

dgrove-oss: I was worried about breaking the invariant that objects that are `equal` should have the same `hashCode`.

Hashing on `name` doesn't work, because sometimes that is allowed to be `null`.  Maybe name (when non-null), size, number of members, and align would be enough?

dgrove-oss: Changed to include name, size, align in hashcode.

dmlloyd (approved):

PR #1614, opened 2022-12-02
Compute statepoints on our side instead of in LLVM

dmlloyd:
This is not the final form of this patch. For one thing it isn't applying relocations, which might (but might not) cause actual optimization issues in LLVM once we start actually relocating things.

Part of looking at the problem of relocations has made me think relocations *and* statepoints might have been better approached from within `LLVMCompatibleBasicBlockBuilder`, and from within a new custom node copier. Much of the information we use for statepoints is lost by the time we get into the tail end of the process, and frankly our node structure now has the capability to represent statepoint, result, and relocation calls directly within the compiler system. This could be done during lowering instead of generation, particularly since we now have value liveness information as part of the scheduler.

But, rather than spend a lot of time reworking this now, I wanted to push it up solely for the reason that it improves build time fairly substantially (by 30% in a few trivial tests), and the improvement should persist regardless of the size of the input program.


dgrove-oss (approved):
LGTM

PR #1615, opened 2022-12-02
Limit errors from `llc` to 1536 characters, and cap errors and warnings by default

dmlloyd:
It's just fixed at 100 for now, we can make it an option later though.


dgrove-oss (approved):
Good enough for now.
I'm fortunately down to only about 75 warnings compiling quarkus...

PR #1616, opened 2022-12-02
LLVM backend improvements

dmlloyd:
* Improve the speed, usability, and disk space requirements of the LLVM backend by making LLVM IR emission and assembly emission independently optional (and defaulting to off)
* Use process pipelines to directly feed LLVM IR source into `llc`
* Remove some redundancy between wasm and non-wasm targets
* Clean up some unused code
* Make process input sources run on the same thread that initiates the transfer, so that the LLVM compiler can still take advantage of large compiler thread stacks (also reduces thread creation churn)
* Abstract LLVM configuration into its own class
* Introduce a "reference strategy" configuration for LLVM


dmlloyd:
This is a draft because it is based on #1614.

dmlloyd:
Undoubtedly much faster! 🎉

PR #1623, opened 2022-12-06
Also apply deref operation to block arguments

dmlloyd:


dgrove-oss (approved):

dmlloyd:
Part of me thinks we should somehow merge derefs within destination blocks by extracting the pointers and creating separate block parameters for the pointer values, which are then dereferenced in the target block.... but that's pretty complex and maybe overkill. I think this will get us past the usability problem anyway.

PR #1637, opened 2022-12-13
Start moving things to their proper sections

dmlloyd:
This is the start of organizing things into sections so that GC can find them. This PR organizes classes, strings, variables of reference type, and other objects into discrete sections.

Also contains some significant bugfixes with how we handle and emit literals.


dgrove-oss:
I think you need to add `visit` methods in `Node.Terminus` for the new `OffsetFromLiteral` and `MemberOfLiteral` classes. Otherwise looks fine to me.

dgrove-oss (approved):

PR #1638, opened 2022-12-14
Drop `PointerLiteral`

dmlloyd:
Also lowers direct memory allocations to global variables which are dropped if they are not reachable at the end of compilation.

~~Still needs a solution for lowering direct byte buffers. My current plan is to lower direct memory allocations to global variable elements.~~


plugins/dispatch/src/main/java/org/qbicc/plugin/dispatch/DispatchTables.java
```diff
@@ -261,13 +260,13 @@ void emitVTable(LoadedTypeDefinition cls) {
                 MethodElement stub = methodFinder.getMethod(vtable[i].isAbstract() ? "raiseAbstractMethodError" : "raiseUnsatisfiedLinkErrorDispatchStub");
                 Function stubImpl = ctxt.getExactFunction(stub);
                 FunctionDeclaration decl = programModule.declareFunction(stub, stubImpl.getName(), stubImpl.getValueType());
-                PointerLiteral literal = ctxt.getLiteralFactory().literalOf(decl.getPointer());
+                Literal literal = ctxt.getLiteralFactory().literalOf(decl.getPointer());
```
dgrove-oss, line 263: Do we still need `decl.getPointer()` here (and in similar cases)?  If I'm following the flow in LiteralFactory correctly `literalOf(decl)` and `literalOf(decl.getPointer())` will both end up returning equivalent `InstanceMethodLiteral` instances

dgrove-oss (approved):

PR #1642, opened 2022-12-16
accept URLs for qbicc-feature.yaml

dgrove-oss:
Enables qbicc-feature to be obtained from a jar for quarkus-qbicc.

### DIFF
--- a/plugins/initialization-control/src/main/java/org/qbicc/plugin/initializationcontrol/QbiccFeatureProcessor.java
+++ b/plugins/initialization-control/src/main/java/org/qbicc/plugin/initializationcontrol/QbiccFeatureProcessor.java
@@ -11,11 +11,12 @@ import java.io.FileInputStream;
 import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.InputStream;
+import java.net.URL;
 import java.nio.file.Path;
 import java.util.List;
 
 public class QbiccFeatureProcessor {
-    public static void process(CompilationContext ctxt, List<Path> features) {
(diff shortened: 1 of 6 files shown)


plugins/initialization-control/src/main/java/org/qbicc/plugin/initializationcontrol/QbiccFeatureProcessor.java
```diff
@@ -27,17 +28,13 @@ public static void process(CompilationContext ctxt, List<Path> features) {
         mapper.setVisibility(PropertyAccessor.FIELD, JsonAutoDetect.Visibility.ANY);
         mapper.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
 
-
-        for (Path feature: features) {
+        for (URL feature: features) {
             QbiccFeature qf;
             try {
-                InputStream fs = new FileInputStream(feature.toFile());
+                InputStream fs = feature.openStream();
```
dmlloyd, line 34: I guess I missed this last time but the stream needs to be closed; it should go in a try-with-resources block.

dmlloyd (commented):
Just one comment but otherwise looks good.

plugins/initialization-control/src/main/java/org/qbicc/plugin/initializationcontrol/QbiccFeatureProcessor.java
dgrove-oss: Thanks.  Fixed.

dmlloyd (approved):

PR #1645, opened 2022-12-20
Enhance QbiccFeature to better support quarkus-qbicc extension

dgrove-oss:


dgrove-oss:
The main gap I know about is that I need to move `reflectiveMethods` and `reflectiveConstructors` to using just the parameter arguments, not the full descriptor.

dmlloyd:
Makes sense; really these specifications are more of a pattern to match than a definite identifier for a single element (which also becomes relevant when, for example, things get loaded in a different class loader than the app CL).

plugins/initialization-control/src/main/java/org/qbicc/plugin/initializationcontrol/FeaturePatcher.java
```diff
@@ -42,44 +45,139 @@ public boolean isRuntimeInitializedClass(String internalName) {
...
+            case "boolean" -> "Z";
+            case "byte" -> "B";
+            case "short" -> "S";
+            case "char" -> "C";
+            case "int" -> "I";
+            case "float" -> "F";
+            case "long" -> "J";
+            case "double" -> "D";
+            // TODO: friendlier processing of array types -- for now we just force the user to provide a descriptor if they want an array
+            default -> t.startsWith("[") ? t : "J"+t.replace('.','/')+";";
```
dmlloyd, line 151: I guess this was meant to be: 
```suggestion
            default -> t.startsWith("[") ? t : "L"+t.replace('.','/')+";";
```
?

dgrove-oss: 🤦 yeah, it was meant to be a descriptor.

dmlloyd (approved):

PR #1647, opened 2022-12-21
Implement `align` for variable elements

dmlloyd:
One last PR before the break ⛄


dgrove-oss (approved):

dgrove-oss:
Looked reasonable to me, but I guess CI disagrees :(

dmlloyd:
It looks like zero-array field layout was working only coincidentally due to relying on max alignment and I broke that by "correcting" the alignment calculation. Now it explicitly puts the array at the end, using the element type's alignment as the basis for its alignment calculation.

PR #1655, opened 2023-01-04
use relative paths for input files to linker command

dgrove-oss:
Workaround problems with long linker command lines by using paths relative to the tool's working directory for the input object files.


dgrove-oss:
This works, but maybe there is a more elegant way to accomplish it.  Happy to adjust if there is...

dmlloyd (approved):

PR #1658, opened 2023-01-05
do not put private methods in instanceMethods -- they don't go in vta…

dgrove-oss:
…bles


dgrove-oss:
This fixes https://qbicc.zulipchat.com/#narrow/stream/284393-qbicc-development/topic/quarkus.20apps/near/319682779

dgrove-oss:
If we decide this fix is semantically ok, I'd probably do a refactor like `getInstanceMethods` ==> `getDispatchedMethods` to emphasize that this isn't the set of all instance methods...

dmlloyd (approved):
I think this is probably OK as is, for now anyway.

PR #1660, opened 2023-01-06
Make annotations available at runtime for reflective classes

dgrove-oss:
If a class has been registered for some reflective operations, then ensure its annotationData is generated at build time so that we can support runtime reflection over its annotations.

### DIFF
--- a/plugins/initialization-control/src/main/java/org/qbicc/plugin/initializationcontrol/FeaturePatcher.java
+++ b/plugins/initialization-control/src/main/java/org/qbicc/plugin/initializationcontrol/FeaturePatcher.java
@@ -54,6 +54,10 @@ public class FeaturePatcher {
         }
     }
 
+    public boolean isReflectiveClass(String internalName) {
+        return reflectiveClasses.containsKey(internalName);
+    }
+
     public void addReflectiveConstructor(String className, String[] parameterTypes) {
         reflectiveConstructors.computeIfAbsent(className, k -> ConcurrentHashMap.newKeySet()).add(encodeArguments(parameterTypes));
     }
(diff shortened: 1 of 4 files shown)


plugins/initialization-control/src/main/java/org/qbicc/plugin/initializationcontrol/FeaturePatcher.java
```diff
@@ -127,14 +131,14 @@ private String encodeArguments(String[] args) {
         if (args == null || args.length == 0) {
             return "";
         }
-        if (args.length == 1 && args.equals("*")) {
+        if (args.length == 1 && args[0].equals("*")) {
```
dgrove-oss, line 134: Changes to encodeArguments are cleanups/fixes flagged by IntelliJ...

plugins/initialization-control/src/main/java/org/qbicc/plugin/initializationcontrol/QbiccFeatureTypeBuilder.java
```diff
@@ -129,6 +133,15 @@ public FieldElement resolveField(int index, DefinedTypeDefinition enclosing, Fie
         }
     }
 
+    @Override
+    public DefinedTypeDefinition build() {
+        DefinedTypeDefinition result = delegate.build();
+        if (reflectiveClass) {
+            ReachabilityRoots.get(classContext.getCompilationContext()).registerReflectiveClass(result.load());
```
dgrove-oss, line 140: forcing loading here is a bit suspect, but a class is only going to be reflective if it was tagged in a qbicc-feature, so it seems to work out ok.

dmlloyd (approved):

Issue #991, opened 2022-01-11
Run time configuration mechanism


dmlloyd:
There needs to be a way to statically configure the run time, beyond system properties.  The mechanism should allow run time components to query configuration that is set into the main builder, generally by way of specific command line properties and in a type-safe manner.  Note that free-form text properties are not adequate; historically the free-form approach has caused significant forward-compatibility and stability issues in other broadly similar situations.

The configuration values can be made accessible by individual `native` methods corresponding to each option.  The methods should be grouped into classes by subsystem. For example, garbage-collection-related classes which would be configured by groups of related options like `--gc-option-name=val` could be gathered in a class called `GC` with method names like `public static native int optionName();`, implemented by one or more intrinsic(s) (preferably one which produces a value based on the invoked method) in the corresponding GC plugin, and configured via some setup from `Main` to the corresponding plugin.  Additionally, in some cases one or more class implementing `BooleanSupplier` will be warranted to allow configured values to apply to `when=` and `unless=` clauses for annotation application.  These classes should be nested within the same class providing the configuration properties.

A couple of example use cases:
- As a JDK porter, I want to know whether the user wants to use (for example) Linux-kernel-specific features or enhancements or fall back to a generic implementation.
- As a GC implementer, I want to know what the base memory size configuration should be.

PR #1667, opened 2023-01-13
plumbing to allow quarkus-qbicc to define system properties

dgrove-oss:


interpreter/src/main/java/org/qbicc/interpreter/impl/VmImpl.java
```diff
@@ -1214,6 +1215,15 @@ public static VmImpl create(CompilationContext ctxt, Consumer<VmObject> manualIn
...
 
+    public VmImpl setPropertyDefines(List<String> propDefines) {
+        this.propertyDefines = propDefines;
+        return this;
+    }
+
+    List<String> getPropertyDefines() {
+        return propertyDefines;
+    }
+
```
dmlloyd, line 1226: I don't really like this way of passing information from the front end/configuration to the back end... but until we have #991 worked out this will do. Once we have some kind of "proper" configuration relay that the run time can consume, we'll want to revisit this.

dmlloyd (approved):

PR #1677, opened 2023-01-24
Add support for resource loading from any class loader

dmlloyd:
* Change `CompilationContextImpl` to use a builder since the argument list is so long
* Add initial support for the platform class loader
* Handle all "special" class loaders identically


dgrove-oss (approved):
LGTM

PR #1680, opened 2023-01-25
implement a simple reflection analysis

dgrove-oss:
Identify when some or all of the arguments to reflective operations are compile time constants and auto-register program elements for runtime reflection.


plugins/reflection/src/main/java/org/qbicc/plugin/reflection/ReflectionAnalyzingBasicBlockBuilder.java
```diff
@@ -0,0 +1,196 @@
...
+
+            // If the target is a method of java.lang.Class, then see if it is one of the reflection entry points
+            // we want to handle specially.
+            if (el.getExecutable() instanceof MethodElement me && me.getEnclosingType().load().equals(km.jlc)) {
+                // replace Class.forName(<literal String>) with a class literal if the type can be resolved at build time
+                if (me.equals(km.forName) && arguments.get(0) instanceof StringLiteral sl) {
+                    String internalName = sl.getValue().replace('.', '/');
+                    DefinedTypeDefinition dtd = getCurrentClassContext().findDefinedType(internalName);
+                    if (dtd != null) {
+                        // TODO: It would be really nice to simply replace the call to forName with a classOf(...).
```
dgrove-oss, line 74: I feel like I'm missing something hopefully simple here.  Without doing the "constant fold" here, we don't actually get much benefit in terms of either optimization or analysis.  Posting the PR to get some ideas...

dgrove-oss: Worked out a plan on Zulip: https://qbicc.zulipchat.com/#narrow/stream/284393-qbicc-development/topic/Class.20init/near/323749027

dgrove-oss:
The analysis piece is working, but I've found a timing issue in how we generate the reflective method accessors.  The current  scheme of using an ElementReachableAdaptor in an Add only works if all reflective methods are known in advance (ie listed in qbicc features or annotations).   Need to instead fire off a background task on the queue to generate the accessors, to allow us to discover that a method is reflectively accessed after we compile it (because it was also reachable non-reflectively).

dgrove-oss:
Needs https://github.com/qbicc/qbicc-class-library/pull/368 to be fully effective for letting us prune the qbicc-feature in quarkus-qbicc, but should be mergable independently.

dgrove-oss:
I was wrong.  Can't be merged until we have a class lib release with the change to the accessor strategy.

dmlloyd (approved):

dgrove-oss:
I added a significant refactor in 3d8145a5.  Should probably review again.

plugins/reflection/src/main/java/org/qbicc/plugin/reflection/ReflectionAnalyzingBasicBlockBuilder.java
```diff
@@ -0,0 +1,200 @@
...
+        return replacement != null ? replacement : super.call(targetPtr, receiver, arguments);
+    }
+
+    @Override
+    public Value callNoSideEffects(Value targetPtr, Value receiver, List<Value> arguments) {
+        Value replacement = process(targetPtr, receiver, arguments);
+        return replacement != null ? replacement : super.callNoSideEffects(targetPtr, receiver, arguments);
+    }
+
+
```
dmlloyd, line 54: An overload for `tailCall` is missing.

plugins/reflection/src/main/java/org/qbicc/plugin/reflection/ReflectionAnalyzingBasicBlockBuilder.java
```diff
@@ -0,0 +1,200 @@
...
+    private Value process(Value targetPtr, Value receiver, List<Value> arguments) {
+        if (targetPtr instanceof ExecutableLiteral el) {
+            KnownMethods km = KnownMethods.get(ctxt);
+
+            // If the target is a method of java.lang.Class, then see if it is one of the reflection entry points
+            // we want to handle specially.
+            if (el.getExecutable() instanceof MethodElement me && me.getEnclosingType().load().equals(km.jlc)) {
+                // replace Class.forName(<literal String>) with a class literal if the type can be resolved at build time
+                if (me.equals(km.forName) && arguments.get(0) instanceof StringLiteral sl) {
```
dmlloyd, line 69: It does still seem like this would be easier (and more efficient) using the intrinsics API to register these hooks, having them return `null` if the criteria are not met. Was there a limitation or problem with the intrinsic API that pushed you in this direction?

plugins/reflection/src/main/java/org/qbicc/plugin/reflection/ReflectiveElementRegistry.java
```diff
@@ -127,6 +143,65 @@ public boolean isReflectiveMethod(String className, String methodName, MethodDes
...
+        return reflectiveLoadedTypes.contains(ltd);
+    }
+
+    public boolean registerReflectiveMethod(MethodElement e) {
+        boolean added = reflectiveMethodElements.add(e);
+        if (added) {
+            ReachabilityRoots.get(ctxt).registerReflectiveEntrypoint(e);
+            ctxt.submitTask(e, methodElement -> {
+                Vm vm = ctxt.getVm();
+                VmThread thr = vm.newThread("genMethodAccessor", vm.getMainThreadGroup(), false,  Thread.currentThread().getPriority());
```
dmlloyd, line 164: I guess this is somehow happening during the earliest phases where we might not have a thread attached to the main thread?

dgrove-oss: I'm taking a look. Might not actually be needed now.

I think it might actually have been the converse... some of this logic used to be run in an ADD postHook, so the ADD TaskWrapper didn't automatically attach the task to a thread.

dgrove-oss: Yep...cargo cult programming.  Didn't need the attach here (or in several other places) got rid of them all in 5c3f3942

dmlloyd (approved):

plugins/reflection/src/main/java/org/qbicc/plugin/reflection/ReflectionAnalyzingBasicBlockBuilder.java
dgrove-oss: I was mentally conflating intrinsics and interpreter hooks when I was working through how this could work.

I also had the thought that a BBB could be run multiple times (for example in the ANALYZE phase).  However, we aren't really setup to construct any of the Method/Field/Constructor data structures we would need after the ADD phase completes so even if we did run an optimizer pass in ANALYZE, we'd still need something in ADD.

So, changed it to be via intrinsics.  The actual code isn't that different, just the setup and boilerplate around it.

PR #1685, opened 2023-01-30
Correctly handle LLVM pointer-backed references

dmlloyd:
The difference is mostly whitespace so [ignore whitespace](https://github.com/qbicc/qbicc/pull/1685/files?diff=unified&w=1) for an easier review.


dgrove-oss (approved):

dmlloyd:
All set, though there is still one difficulty: linkage fails in WASM due to pointers being stored in `long`-typed fields.

dgrove-oss (approved):

PR #1700, opened 2023-02-14
Add a digest to hidden class identifiers

### LINKED ISSUE #1687: Hidden class names & reproducibility
There are many things in a Java program that makes reproducibility difficult, but one of the more annoying ones which relates to qbicc is the way we name/number hidden classes.

When a hidden class is defined, the name of the class is not derived from the class file bytes; rather, an index is incremented and a name is generated based on the name of the template class plus the index. However, the ordering of hidden class definition is indeterminate as it is based on program execution during compile. This lack of consistency between builds makes it very hard to compare a class or LLVM IR or assembly file between compile runs, especially for finicky things like method handles and lambdas.

We should use a more consistent scheme, like an SHA-1 or SHA-256 hash (or hash prefix) as a name suffix for hidden classes, where the hash is built over the bytes of the template class. An index should only be added when more than one class hashes to the same hash (which should only happen when the same class is deliberately defined multiple times).

dmlloyd:
Fixes #1687.

### DIFF
--- a/compiler/src/main/java/org/qbicc/type/ObjectType.java
+++ b/compiler/src/main/java/org/qbicc/type/ObjectType.java
@@ -3,6 +3,7 @@ package org.qbicc.type;
 import java.lang.invoke.ConstantBootstraps;
 import java.lang.invoke.MethodHandles;
 import java.lang.invoke.VarHandle;
+import java.util.Base64;
 
 import org.qbicc.type.definition.DefinedTypeDefinition;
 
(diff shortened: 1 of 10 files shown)


dmlloyd:
This attempt calculates a hash over every class file. I don't know if I like that; hash calculation isn't exactly cheap so I'd rather only do it where it's needed. But, we'll see if the impact is actually observable before I go optimizing things.

dmlloyd:
I've changed it to only add a hash for hidden classes.

dgrove-oss (approved):

PR #1707, opened 2023-02-20
need to use valueConvert in ptrDiff to get valid llvm-ir

dgrove-oss:


dgrove-oss:
Test snippet:
```
    public static int doit2(char_ptr p1, char_ptr p2) {
        return p1.minus(p2).intValue();
    }
```
When we were using bit cast, we got invalid ir:
```
  %bb1.5 = bitcast i8* %p0 to i64, !dbg !825 ; Minus.java:13 bci@2
```

dmlloyd (approved):

PR #1713, opened 2023-02-22
remove ioctl; now defined in posix.SysIoctl

dgrove-oss:
Probably needs a class lib release; submitting PR now so I don't forget about it...


dmlloyd (approved):
`ioctl` isn't really POSIX, but it is similar across OSes. We can always revisit if it's a problem later.

dgrove-oss:
yeah, agreed.  its one of those defacto standards.  We need a place for "things that are the same on MacOS and Linux but that aren't formally specified" :)

PR #1718, opened 2023-02-23
Deprecate pointer-form classes and methods; introduce non-generic pthread start function interface

dmlloyd:
Generic type inference of functional interfaces for function pointers is still a bit screwy. We can reduce complexity by just removing support for generic function invocation interfaces.

Also, deprecate all of the special pointer-form methods and classes from `Pthread`, replacing them with `ptr`-based methods.


dgrove-oss (approved):

dmlloyd:
This exception was the exact reason for doing this :)

PR #1734, opened 2023-03-02
Bring back `deref()`

dmlloyd:
Also, `sel()` is given a method body. The method body is not directly used (it's still an intrinsic), however it makes it easy to use an IDE's "inline method" refactoring to easily transform `x.sel()...` forms to `deref(x)...`.


dgrove-oss (approved):
LGTM

PR #1742, opened 2023-03-08
Revert "Statepoint relocations"

dmlloyd:
Reverts qbicc/qbicc#1740. Unfortunately this breaks (unsupported) on aarch64, so the patch has to be reworked so that it is only enabled for supported platforms, while we look for another solution.


dmlloyd:
Went with #1744 instead.

PR #1743, opened 2023-03-08
Disable statepoint relocations

dmlloyd:
An alternative to #1742 which just disables statepoint relocations. Later we can remove them completely if needed by removing this method.


dgrove-oss (approved):

dmlloyd:
Went with #1744 instead.

PR #1750, opened 2023-03-13
Basic support for unions

dmlloyd:
At build time, a value of union type must contain only zero, otherwise an error will be produced. Type punning via unions is not presently allowed.


dmlloyd:
Includes #1749 so I'll rebase once that's in.

dgrove-oss (approved):
LGTM

PR #1783, opened 2023-04-04
Fix AArch64 (non-Mac) to store registers properly

dmlloyd:
The non-LLVM version of `libunwind` stores registers in a field which is offset from the base. The LLVM version (used by mac) does not have the bug that the inline assembly works around. Fix the assembly to store the PC and flags as well.

Also add predicates to test for AArch64.


dmlloyd:
We might have to just drop the workaround and suggest that only LLVM's libunwind be used.

Or perhaps better yet, drop libunwind and do the stack walking ourselves - it's generally a pointer-chase along the stack. Maybe this can be a long-term goal.

PR #1789, opened 2023-04-10
Support and implementation of updated mark, safepoint, and iteration strategy

dmlloyd:


plugins/gc/common/src/main/java/org/qbicc/plugin/gc/common/safepoint/SafePoints.java
```diff
@@ -33,31 +25,15 @@ public static void enqueueMethods(CompilationContext ctxt) {
...
+        final ClassContext bcc = delegate.getContext().getBootstrapClassContext();
+        final DefinedTypeDefinition dt = bcc.findDefinedType(THREAD_INT_NAME);
         return new DelegatingBasicBlockBuilder(delegate) {
+
             @Override
             public Node safePoint() {
-                strategy.safePoint(getFirstBuilder());
-                return nop();
+                final MethodElement pollSafePoint = dt.load().requireSingleMethod("pollSafePoint");
+                return call(getLiteralFactory().literalOf(pollSafePoint), List.of());
```
dmlloyd, line 35: This requires that `pollSafePoint` exists on `Thread`, so it'll likely require another classlibs release. Need to think about ordering a bit.

dmlloyd:
Merged separately

PR #1819, opened 2023-04-20
Add Cirrus build configuration

dmlloyd:


dmlloyd:
I believe that the build is now currently failing due to llvm/llvm-project#61265. Once the fix gets into some stable branches we can update and try again.

Still to do:
- [ ] Await fix for llvm/llvm-project#61265
- [ ] Introduce equivalent of matrix build for more than one OS and LLVM configuration (we might want to keep github actions as well, so in this case select combinations (aarch64) that do not overlap with what github already tests)

PR #1832, opened 2023-04-28
Set context class loader when initializing classes

dmlloyd:


dmlloyd:
Change didn't account for recursive init. Fix incoming.

dgrove-oss (approved):

PR #1838, opened 2023-05-02
Test with opaque pointers inline with job

dmlloyd:


dmlloyd:
The LLVM 15 runs take around 50% longer (from ~12 to ~18 wallclock minutes for linux and from ~21 to ~34 for macos), which is reasonably in line with expectations, in exchange for a net reduction in terms of overall runner-minutes (adding ~19 runner-minutes and subtracting ~30-35 runner-minutes). Worth it?

dgrove-oss:
Seems like a reasonable tradeoff to me, given the throttling we see for macOS runners.

PR #1950, opened 2023-11-28
Try Mac OS M1 runners

dmlloyd:


dmlloyd:
Doesn't seem to work; the run just terminates immediately as a "success". Revisit later...

PR #2024, opened 2024-07-23
Removed collection which is updated, but never queried

dk2k:


dmlloyd:
This was left intentionally for a future enhancement that may or may not ever arrive, but either way I'm inclined to leave it as-is. Thanks.

